PR #26, opened 2026-03-08
Add pivot table feature (Phase 1)

hackerwins:
## Summary
- Add pivot table calculation pipeline (parse, group, aggregate, calculate, materialize) to the sheet package
- Add pivot editor side panel with field configuration (row/column/value/filter fields) and aggregation support (SUM, COUNT, COUNTA, AVERAGE, MIN, MAX)
- Wire pivot creation into context menu and document tabs with pivot sheet protection (block direct cell edits)
- Extend Store interface with `setPivotDefinition`/`getPivotDefinition` and `invalidate()` for external write support

## Test plan
- [x] All 976 existing tests pass (`pnpm verify:fast`)
- [x] 7 new pivot test files covering parse, group, aggregate, calculate, materialize, e2e, and protection
- [x] No regressions to existing features (mobile handles, absolute references, backend tests preserved)
- [x] Manual: select a data range → right-click → "Insert pivot table" → configure fields in editor panel
- [x] Manual: verify pivot sheet blocks direct cell edits, row/column insert/delete, and paste

## Summary by CodeRabbit

* **New Features**
  * Pivot table functionality with a right-side editor for rows, columns, values, filters, and a Refresh action
  * Support for SUM, COUNT, COUNTA, AVERAGE, MIN, MAX aggregations; multi-value pivots and totals display
  * Insert pivot table option in the sheet context menu and create-from-selection flow
  * Pivot results materialize as a protected pivot sheet (read-only for structural edits)

* **UX**
  * Deletion confirmation that lists and can cascade-delete dependent pivot tabs when removing source tabs

* **Tests**
  * Comprehensive unit and end-to-end tests for pivot parsing, grouping, aggregation, materialization, and protection

### DIFF
--- a/packages/frontend/src/app/documents/document-detail.tsx
+++ b/packages/frontend/src/app/documents/document-detail.tsx
@@ -22,6 +22,15 @@ import { usePresenceUpdater } from "@/hooks/use-presence-updater";
 import { IconFolder, IconSettings, IconDatabase } from "@tabler/icons-react";
 import { fetchWorkspaces, type Workspace } from "@/api/workspaces";
 import { toast } from "sonner";
+import { Button } from "@/components/ui/button";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@/components/ui/dialog";
 import { TabBar } from "@/components/tab-bar";
 import {
   SpreadsheetDocument,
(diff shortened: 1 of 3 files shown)


hackerwins:
@coderabbitai Thanks for the thorough review on the latest commit. Here's our assessment of the 3 new Major comments:

**#1 Stale delete confirmation** — The race condition requires another client to add/remove tabs during the few seconds the dialog is open. We already re-collect dependents at confirm time, which is a reasonable safeguard for this edge case. We'll revisit if collaboration-heavy usage reveals real issues.

**#2 Reset editor on pivot load** — `loadPivotDefinition()` is called during `initialize()`, before any editing session starts. The scenario of an already-open editor transitioning to pivot state doesn't occur in practice.

**#3 Mobile/formula-bar edit bypass** — `showCellInput()` already has the `isPivotSheet()` guard, so mobile double-tap paths that flow through it are blocked. `finishEditing()` requires an open editor to reach, but the editor can't be opened on pivot sheets in the first place.

All three are defensive improvements for edge cases that are unlikely in Phase 1. We'll harden these paths as the pivot feature matures.

PR #118, opened 2026-04-11
Harden DOCX table merge import for row-shape edge cases

hackerwins:
## Summary

Follow-up to PR #117 focused on the `cells.length === numCols` contract
that the rest of the docs table stack (layout, renderer, click handler,
exporter) relies on. Five independent edge cases could each leave a
row mis-shaped relative to the `tblGrid`, making clicks land on the
wrong cell or truncating merged regions at render time.

- **Honor `w:gridBefore` / `w:gridAfter`** — rows that leave leading or
  trailing grid columns empty now pad with `colSpan=0` placeholders so
  they match `numCols`. Common in Korean government forms.
- **Widen vMerge continue to the owner's `colSpan`** — when a
  vertical-merge owner declares `gridSpan=N` but the continuation tc
  declares a smaller span (Word can emit this after manual edits), the
  continuation now covers the full merged width.
- **Promote orphan `vMerge=continue` to a standalone owner** — a
  continuation tc with no prior restart (some writers leave these
  behind when the anchor row is deleted) was silently turning the
  column's grid positions into unreachable placeholders. It now
  becomes a real cell with its own blocks.
- **Clamp out-of-range `gridSpan`** — a tc can declare `w:gridSpan`
  larger than `numCols - colIdx`. The clamp stops `colIdx` from
  walking past the grid, which previously left the row longer than
  every other row in the table.
- **Final row-shape normalize** — last-line safety net that pads or
  truncates each row to `numCols`, so downstream code can trust the
  rectangular shape even if any of the above misses a case we have
  not seen yet.

All behavior is gated on `tblGrid` being present (`numCols > 0`). When
the fixture has no `tblGrid` we leave the row walk unchanged.

Each fix landed with a dedicated vitest fixture (TDD). The new test
file adds 7 cases; `pnpm --filter @wafflebase/docs test` goes from 464
to 471 passing tests.

## Test plan

- [x] \`pnpm --filter @wafflebase/docs test\` — 471 passing
- [x] \`pnpm verify:fast\` (pre-commit hook on every commit)
- [x] Pre-push harness lanes (verify:fast, frontend/backend/cli builds, chunks, entropy)
- [x] Manual smoke test: reload real-world form.docx and click each cell of a gridBefore/gridAfter row to confirm the cursor lands correctly
- [x] Manual smoke test: reload a file with a vMerge restart that has gridSpan > 1 and confirm the merged region stays rectangular

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DOCX table import to enforce grid-aligned rows, clamp overrun spans, and correctly handle vertical-merge continuations and orphan continues so downstream layout is stable.

* **Tests**
  * Added thorough tests covering grid padding/truncation, span clamping, vMerge continuation/promote behavior, and gridless-table compatibility.

* **Documentation**
  * Added task and lessons notes tracking remaining mapping gaps, verification rules, and importer hardening decisions.

### DIFF
--- a/docs/tasks/active/20260411-docx-table-merge-gaps-lessons.md
+++ b/docs/tasks/active/20260411-docx-table-merge-gaps-lessons.md
@@ -1,0 +1,85 @@
+# DOCX table merge import hardening — lessons
+
+Paired with `20260411-docx-table-merge-gaps-todo.md`. Records what
+was surprising or non-obvious while landing the five
+row-shape fixes on PR #118.
+
+## The `cells.length === numCols` contract is load-bearing
+
+Layout, the Canvas renderer, the click handler, and the exporter all
+index `rows[r].cells[c]` by grid column and silently assume every
+row is the same length as `columnWidths`. There is no central
+assertion enforcing it — the whole pipeline trusts whatever the
+importer built. A single off-by-one row drags the entire table out
+of alignment. When touching the DOCX importer, treat the "numCols"
+contract as a public API: if a code path can push a non-rectangular
+row, gate it or normalize it.
+
+## `colSpan === 0` means "covered", not "vertical merge"
+
+`Doc.mergeCells` (`packages/docs/src/model/document.ts`) sets
+`colSpan = 0` for every merged covered cell — horizontal, vertical,
+or both. The importer's `makeCoveredCell` helper follows the same
+contract. The exporter, however, maps any `colSpan === 0` cell to
+`<w:vMerge/>` unconditionally, which is wrong for horizontal-only
+merges and for the new `gridBefore`/`gridAfter` and safety-net
+placeholders. This mismatch is pre-existing (predates PR #118) but
+worth remembering: if you introduce a new source of `colSpan: 0`
+cells on the import side, you are also widening the scope of the
+exporter bug. The exporter-side disambiguation lives as follow-up
+item **E1** in the todo.
+
+## `numCols === 0` is the legacy gridless path
+
+When `<w:tblGrid>` is missing, `columnWidths.length === 0` and the
+hardening logic (clamp, normalize) is explicitly disabled so the
+row walk keeps the historical "one entry per tc" shape. Any
+*new* transform that depends on `numCols` must gate on
+`numCols > 0` and have a matching gridless regression test.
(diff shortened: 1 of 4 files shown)


hackerwins:
Thanks for the review — addressed in `2d71e560`.

**1. Missing paired lessons file** ✅ — added `docs/tasks/active/20260411-docx-table-merge-gaps-lessons.md` alongside the todo, summarizing the `cells.length === numCols` contract, the `colSpan === 0` overloading, and the squash-merge/cherry-pick workflow that produced this PR.

**2. `gridBefore`/`gridAfter` applied when `tblGrid` is absent** ✅ — confirmed the inconsistency: the clamp and final normalize were gated on `numCols > 0` but the skip-marker padding was not. Fixed by gating both `readGridSkip` calls on `numCols > 0`.

**3. Synthetic placeholders round-trip as bogus `<w:vMerge/>`** ⏭️ — valid but **pre-existing**, not introduced by this PR. The exporter at `packages/docs/src/export/docx-exporter.ts:211-214` serializes every `colSpan === 0` cell as `<w:vMerge/>`, and `Doc.mergeCells` (`packages/docs/src/model/document.ts:713-714`) has always set `colSpan = 0` for *every* merged covered cell — including the horizontal-merge placeholders added in #117. There is no existing export test covering merged cells (`grep colSpan|gridSpan|vMerge|merge` in `test/export/docx-exporter.test.ts` returns nothing), so any horizontal merge already round-trips as bogus vMerge today. This PR just widens the exposure with gridBefore/gridAfter and the safety-net pad.

The correct fix is exporter-side disambiguation: a covered position absorbed by a prior `gridSpan` in the same row should not emit a tc at all; a position whose owner lives in an earlier row should emit `<w:vMerge/>`; and `gridBefore`/`gridAfter` should be emitted via `trPr` skip markers. That's a distinct change and belongs in a follow-up. Filed as item **E1** in `20260411-docx-table-merge-gaps-todo.md`.

**4. Missing gridless-table counterpart tests** ✅ — added `should not inject gridBefore placeholders when the table has no tblGrid` and the matching `gridAfter` case, both driven TDD-first against the fix in #2. They assert the row keeps exactly the tc count the source shipped.

Test count: 471 → 473 passing. Pre-push harness lanes still green.

PR #145, opened 2026-04-20
Fix table row splitting: click, cursor, selection, and rendering

hackerwins:
## Summary

- Split tall table rows across page boundaries so content renders continuously
- Fix click resolution, cursor placement, and cell selection for split rows with nested tables
- Fix nested table background rendering order so selection highlights remain visible
- Fix Cmd+A delete leaving empty table shells instead of empty paragraphs

## Key changes

**Row splitting** (`pagination.ts`, `table-layout.ts`, `table-renderer.ts`, `doc-canvas.ts`):
- Extend `PageLine` with `rowSplitOffset`/`rowSplitHeight` metadata
- Detect rows exceeding page space and compute safe split heights
- Render split fragments with clipping and re-drawn borders per page

**Click resolution** (`text-editor.ts`):
- Use labeled break (`break findInnerTableY`) so nested table Y computation stops at the correct fragment
- Account for split offset in nested table inner row resolution
- Clamp target line to visible fragment bounds
- Return correct block for deeper nesting instead of always `blocks[0]`
- Guard against empty tables (rows=0) in click handling

**Cursor rendering** (`peer-cursor.ts`):
- Extract `findSplitFragment` helper for fragment selection
- Add straddling match for nested tables crossing split boundaries
- Clamp cursor Y to fragment top when straddling produces negative offset

**Selection highlighting** (`selection.ts`, `table-renderer.ts`):
- Recurse into nested tables during background pass so backgrounds don't cover selection highlights
- Support split row fragments in cell-range rects for both top-level and nested tables

**Other** (`text-editor.ts`):
- Cmd+A delete replaces everything with empty paragraph (not empty table shell)
- Line-by-line arrow navigation in table cells

## Test plan

- [ ] Open a document with tall table rows that span page boundaries
- [ ] Verify rows split across pages without blank space or clipping
- [ ] Click on cells in split row fragments — cursor appears on correct page
- [ ] Click on nested table cells within split rows — cursor appears at clicked position
- [ ] Select cells in nested tables on both pages — selection highlight visible on both
- [ ] Arrow Up/Down across split row boundaries moves cursor correctly
- [ ] Cmd+A then Delete produces empty paragraph, not empty table

## Summary by CodeRabbit

* **New Features**
  * Tables can now split across page boundaries, allowing tall rows to span multiple pages instead of forcing oversized pages.
  * Enhanced cursor navigation and text selection for split table rows across page fragments.

* **Tests**
  * Added comprehensive test coverage for table row splitting during pagination.

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -47,6 +47,7 @@ Word processor engine — rich text, tables, pagination, collaboration.
 | [docs-wordprocessor-roadmap.md](docs/docs-wordprocessor-roadmap.md)              | Docs word processor roadmap — 6-phase plan for Google Docs parity                                 |
 | [docs-ruler.md](docs/docs-ruler.md)                                              | Docs ruler design                                                                                 |
 | [docs-table-copy-paste.md](docs/docs-table-copy-paste.md)                        | Docs table copy-paste — cell-range clipboard, whole-table block, external HTML table paste        |
+| [docs-table-row-splitting.md](docs/docs-table-row-splitting.md)                  | Table row splitting — split tall table rows across pages, recursive nested table support          |
 
 ## Common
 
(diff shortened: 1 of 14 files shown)


hackerwins:
@coderabbitai review

PR #231, opened 2026-05-13
fix: support OFFSET range references

### LINKED ISSUE #213: `OFFSET()` ignores height and width parameters
### What happened:
`OFFSET(ref, rows, cols, height, width)` should return a range of size `height × width`. Wafflebase parses `height` and `width` but always returns a single cell, so any caller (typically wrapped in `SUM`, `INDEX`, or used as a range argument) sees an empty value or just one cell.

`OFFSET(ref, rows, cols, height, width)` should return a range of size `height × width`.

**Setup:** A1=`10`, A2=`20`, A3=`30`.

| Cell | Formula                          | Expected     | Actual       | Status |
| ---- | -------------------------------- | ------------ | ------------ | ------ |
| B1 | `=SUM(OFFSET(A1, 0, 0, 3, 1))` | `"60"` | `"10"` | ✗ |
| B2 | `=OFFSET(A1, 100, 0)` | `"#REF!"` | `""` | ✗ |

`B1` should sum A1:A3 (height=3 produces a 3-row range), but only A1 is passed to `SUM`. `B2` should error on out-of-bounds offset; instead returns empty silently.

### Environment:

- Operating system:
- Browser and version:

lisiqi1983:
## Summary
- make `OFFSET` return references/ranges so callers like `SUM` can consume the full `height x width` result
- validate offset dimensions and return `#REF!` for references outside the available formula grid
- include static `OFFSET` output ranges in dependency extraction so recalculation fetches the cells the offset range will read
- add regression coverage for scalar offsets, multi-cell offset ranges, omitted range dimensions, dynamic dimensions, cross-sheet references, invalid grid keys, and static dependency extraction edge cases
- document the new OFFSET helper functions to reduce review noise around docstring coverage

Fixes #213

## Verification
- `pnpm --filter @wafflebase/sheets test -- test/formula/formula.test.ts` (489 tests passed)
- `git diff --check -- packages/sheets/src/formula/formula.ts packages/sheets/src/formula/functions-lookup.ts packages/sheets/test/formula/formula.test.ts`
- commit hook `pnpm verify:fast` (passed)

## Enhancements

* Improved formula engine to better identify cell and range references within formulas
* Extended OFFSET function support to discover implied references based on computed arguments
* Enhanced reference tracking to recognize dynamic cell references in complex formula expressions

[Review Change Stack](https://app.coderabbit.ai/change-stack/wafflebase/wafflebase/pull/231)

### DIFF
--- a/packages/sheets/src/formula/formula.ts
+++ b/packages/sheets/src/formula/formula.ts
@@ -345,6 +345,7 @@ function buildStaticOffsetReference(
         ? toSref(newFrom)
         : toSrng([newFrom, newTo]);
     return `${prefix}${offsetReference}`;
+    /* c8 ignore next 3 -- Formula grammar only produces valid REF/REFRANGE nodes here. */
   } catch {
     return undefined;
   }
--- a/packages/sheets/src/formula/formula.ts
+++ b/packages/sheets/src/formula/formula.ts
@@ -30,6 +30,8 @@ import {
   parseRef,
   parseRange,
   isCrossSheetRef,
+  toSref,
+  toSrng,
 } from '../model/core/coordinates';
 
 /**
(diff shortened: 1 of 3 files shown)


ggyuchive:
Thanks for your contribution, but there's another PR that fixes same issue.
So let me close it.

PR #234, opened 2026-05-13
Fix OFFSET height and width handling

### LINKED ISSUE #213: `OFFSET()` ignores height and width parameters

Thanhdn1984:
## Summary
- make `OFFSET` return a range when height/width are provided
- return `#REF!` for non-positive height/width
- add regression coverage for `SUM(OFFSET(...))` ranges

Fixes #213

## Verification
- `pnpm sheets test -- formula.test.ts --runInBand`
- `pnpm sheets typecheck`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the OFFSET formula to properly support height and width parameters, enabling it to return range references for multi-cell selections.
  * Fixed OFFSET to correctly handle invalid dimensions and out-of-bounds references by returning appropriate error values.

* **Tests**
  * Added test coverage for OFFSET function behavior with various parameter combinations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wafflebase/wafflebase/pull/234)

### DIFF
--- a/packages/sheets/src/formula/functions-lookup.ts
+++ b/packages/sheets/src/formula/functions-lookup.ts
@@ -10,6 +10,7 @@ import {
   parseRange,
   parseRef,
   toColumnLabel,
+  toSrng,
 } from '../model/core/coordinates';
 import {
   toStr,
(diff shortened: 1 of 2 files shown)


ggyuchive:
LGTM. Thanks for your contribution.

PR #257, opened 2026-05-17
Add design note: local caret anchoring (issue #237)

### LINKED ISSUE #237: Question: Should local caret shift when remote user inserts text upstream in the same paragraph?
### What happened:

In the collaborative document editor, when User A inserts text before User B's caret position within the same paragraph, User B's caret does not shift along with the growing text. It stays at the same absolute offset, so it visually appears at the same screen position while the surrounding text moves to the right. As a result, User B's caret no longer points to the character it was originally anchored to.

I'm not sure whether this is a bug or an intentional design choice, so I'd like to confirm.

I expected User B's caret to remain anchored to the same logical position (e.g., still right after the character it was placed after), shifting its visual position to the right as User A's text is inserted upstream. This matches how Google Docs and Notion behave.

Example:

- Initial paragraph: `Hello world`
- User B's caret is placed right after `wo` (between `wo` and `rld`)
- User A types `Hi ` at the beginning of the same paragraph
- Expected: paragraph becomes `Hi Hello world`, User B's caret is still right after `wo`
- Actual: paragraph becomes `Hi Hello world`, but User B's caret remains at the same absolute offset, now pointing somewhere inside `Hello` instead of after `wo`

1. Open the same document in two browsers (User A and User B)
2. As User B, click into a paragraph and place the caret in the middle of the text (right after a specific character)
3. As User A, place the caret near the beginning of the same paragraph and type some characters
4. Observe User B's caret position — it stays at the same visual location instead of shifting right along with the inserted text

There is a valid design argument for keeping an idle caret visually stable to avoid disorienting the user during remote edits, so this might be intentional. A couple of questions:

- Is the current behavior intentional?
- Or is anchoring the caret to a relative position (e.g., Yorkie's path-based position / IndexTree path) planned but not yet implemented?

Happy to file a proper bug report or contribute a fix if this turns out to be unintended.

### Environment:

- Operating system: <e.g. macOS 14.5>
- Browser and version: <e.g. Chrome 138.0.7204.97>

semimikoh:
The local caret in the docs editor is stored as an absolute
`{ blockId, offset }`. When a remote peer edits the same block, the
render cache is invalidated but the local offset is not transformed, so
the caret ends up pointing at the wrong character (issue #237).

This PR lands both the design note and its implementation: local
cursor, non-collapsed selection endpoints, and the IME composition
start position are anchored to Yorkie Tree positions in the frontend
collaboration layer and resolved back to `DocPosition` only at render
time.

## Summary

- Design note `docs/design/docs/docs-local-caret-anchoring.md` pins the
  Yorkie API (`indexRangeToPosRange` / `posRangeToIndexRange`), makes
  the same-boundary affinity rule normative, documents the deterministic
  fallback ladder, and records `splitBlock` / `mergeBlock` as a known
  limitation (current `delete + insert` implementation cannot preserve
  `TreePos` through deleted content; native `splitLevel` is a
  follow-up).
- `YorkieDocStore` captures `AnchoredDocPosition` / `AnchoredDocRange`
  on cursor change, resolves both on every remote document change, and
  walks a five-step fallback (same block → previous region block end →
  next region block start → first editable region block → first
  editable body block) when the anchor's block was removed.
- `TextEditor` emits `onCompositionStart` / `onCompositionEnd`; the
  collaboration layer captures a composition anchor and pushes the
  drift-corrected start position back through
  `EditorAPI.updateCompositionStartPosition` so IME commits never land
  at a stale absolute offset.
- Yorkie-specific anchor state is kept inside `packages/frontend`;
  `packages/docs` still receives only resolved `DocPosition` values.

## Why

The maintainer confirmed in #237 that the absolute-offset drift is a
bug and pointed at anchoring through Yorkie Tree positions. The review
on this PR also asked for two doc updates that are now reflected in the
note: pin the concrete API surface and make the affinity rule
normative. The implementation lands here rather than in a follow-up
because the maintainer asked to keep design and implementation in the
same PR.

## Linked Issues

Fixes #237.

## Author checklist

## Verification

Local `pnpm verify:fast` runs green on each of the three commits in
this branch.

## Risk Assessment

- User-facing risk: Behavior change in the docs collaborative caret —
  before this PR, a remote insert before the local caret left the caret
  pointing at the wrong character; after, the caret follows the
  intended logical boundary. The change is scoped to docs and does not
  touch sheets/slides.
- Data/security risk: None. No schema or storage changes.
- Rollback plan: Revert the three commits in this branch. The store
  API additions (`resolveAnchoredLocalCursor`, `setCompositionStart`)
  and `EditorAPI` additions (`restoreLocalCursor`, composition hooks)
  are net-new and not depended on by callers outside this PR.

## Notes for Reviewers

- UI changes (screenshots/gifs if applicable): N/A — the visible
  behavior change is "the caret no longer drifts under concurrent
  remote inserts" and is exercised by the two-client integration tests
  rather than by a screenshot.
- Follow-up work: `splitBlock` / `mergeBlock` use `delete + insert`
  against the Yorkie Tree rather than native `splitLevel=2`. Anchors
  whose target text is removed by a remote split or merge therefore go
  through the deterministic fallback ladder instead of preserving the
  logical offset through CRDT semantics. Switching to native
  split/merge is tracked in *Known Follow-Ups* and would also need to
  restore the undo/redo regression previously documented under
  `splitLevel=2`.
- AI assistance: Drafted with help from Claude Code across both the
  design note and the implementation (anchor data flow, fallback
  ladder, IME wiring, tests, doc updates). I reviewed every line and
  own the technical choices — happy to defend or revise any of them.

## Release Notes

* **New Features**
  * Implemented local caret and selection anchoring for collaborative documents, automatically preserving cursor position during remote changes.
  * Added IME composition lifecycle support for improved text input handling.

* **Documentation**
  * Added design documentation outlining the caret anchoring strategy.
  * Added task tracking for implementation steps.

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -46,6 +46,7 @@ Word processor engine — rich text, tables, pagination, collaboration.
 | [docs-comments.md](docs/docs-comments.md)                                        | Docs comments — text-range threads, CRDT-stable posRange anchors, orphan preservation, shared frontend module |
 | [docs-font-controls.md](docs/docs-font-controls.md)                              | Docs font controls — curated family picker (14 fonts), Google-Docs-style size input, line spacing, clear formatting, shared text-formatting components |
 | [docs-pending-inline-style.md](docs/docs-pending-inline-style.md)                | Pending inline style at a collapsed caret — stored marks for toolbar toggles, IME-aware, view-local            |
+| [docs-local-caret-anchoring.md](docs/docs-local-caret-anchoring.md)              | Local caret anchoring — Yorkie Tree-anchored caret/selection, resolves to DocPosition at render time (issue #237) |
 
 ## Slides
 
--- a/docs/design/docs/docs-local-caret-anchoring.md
+++ b/docs/design/docs/docs-local-caret-anchoring.md
@@ -79,7 +79,8 @@ The Yorkie-backed frontend integration should own the anchored representation:
 ```ts
 type AnchoredDocPosition = {
   region: 'body' | 'header' | 'footer';
-  yorkiePosition: unknown;
+  yorkiePosition: TreePos;
+  lineAffinity?: 'forward' | 'backward';
 };
 
 type AnchoredDocRange = {
@@ -89,10 +90,19 @@ type AnchoredDocRange = {
 };
 ```
 
-The concrete `yorkiePosition` type should be selected from the Yorkie SDK Tree
-position APIs used by the collaboration layer. The important boundary is that
-Yorkie-specific state does not leak into the core docs model unless
-implementation proves that the boundary is too expensive or ambiguous.
+The concrete anchor is Yorkie's Tree position type. Convert a collapsed caret
+or range endpoint with `tree.indexRangeToPosRange([index, index])`, and resolve
+it with `tree.posRangeToIndexRange(posRange)`. A caret is a collapsed range, so
+`AnchoredDocRange` maps 1:1 to two endpoint anchors. Yorkie-specific state
+does not leak into the core docs model; `packages/docs` still receives only
+resolved `DocPosition` values.
+
+Headers, body, and footers all live in the same `root.content` Tree as sibling
+top-level containers. Header and footer blocks are wrapped by top-level
+`type: 'header' | 'footer'` nodes, while body blocks are direct top-level
+block nodes. `indexRangeToPosRange` therefore handles all three regions
+uniformly. The `region` tag is retained only so fallback stays region-local
+when the anchor target is deleted.
 
 ### Data Flow
 
@@ -108,6 +118,10 @@ Remote document change arrives
   -> editor cursor and selection are updated
   -> render cache is invalidated and repaint happens
 ```
+
+The anchor is canonical. A resolved `DocPosition` is a cache derived from the
+current Tree snapshot and is invalidated on remote change. `pendingCursorPos`
+continues to store absolute offsets for undo/redo presence history for now.
 
 ### Converting DocPosition To An Anchor
 
@@ -147,9 +161,21 @@ After remote insert at start: Hi Hello wo|rld
 
 ### Remote Insert At The Same Boundary
 
-Use Yorkie Tree position semantics for tie-breaking. The implementation should
-document whether the local caret resolves before or after the concurrently
-inserted remote text at the same boundary.
+Use Yorkie Tree position semantics for tie-breaking, then apply the local
+affinity rule before fallback. The local caret uses right affinity for the
+user's own insert so typing advances the caret after inserted text. Remote
+inserts at the same boundary use left affinity so the local caret stays before
+the remotely inserted text, matching Google Docs behavior.
+
+```text
+User B anchor: He|llo
+User A inserts "y" at the same boundary
+Resolved for B: He|yllo
+```
+
+Tests must assert this same-boundary affinity. If the Yorkie SDK's
+`posRangeToIndexRange` semantics differ, the frontend wrapper normalizes the
+resolved `DocPosition` to this rule.
 
 ### Remote Delete Before Caret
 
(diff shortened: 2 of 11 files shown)


hackerwins:
Thanks for the careful note, @semimikoh. Direction is right and the
edge-case enumeration (regions, tables, IME, fallback ladder) is exactly
what I was hoping for. Please pin two things in the doc, then continue
with the implementation in this same PR — no need to split into a
separate one.

## Doc updates

**1. Pin the Yorkie API.** Open Question #1 is answered by the official
guide: https://yorkie.dev/docs/sdks/js-sdk#custom-crdt-types — anchor
via `tree.indexRangeToPosRange`, resolve via `tree.posRangeToIndexRange`.
A caret is a collapsed range; `AnchoredDocRange` maps 1:1. Replace
`yorkiePosition: unknown` with the concrete `TreePos` type, and add
`lineAffinity?: 'forward' | 'backward'` to the anchor — line wrap is a
layout concern, not CRDT.

Note that headers, body, and footers all live in the same `root.content`
Tree as sibling top-level containers (`type: 'header' | 'footer'`
wrapping body blocks, see `yorkie-doc-store.ts`). `indexRangeToPosRange`
therefore handles all three regions uniformly — the `region` tag in
`AnchoredDocPosition` is only needed so the fallback ladder can stay
region-local when the anchor target is deleted. Worth saying this
explicitly in the doc so a reader doesn't assume separate Trees.

State as "verify-then-pin during implementation": same-boundary
tie-breaking, and behavior of a `TreePos` whose node was deleted.

**2. Make affinity normative** (L149–L153). Replace "implementation
should document..." with: local caret uses right affinity for its own
inserts and left affinity for remote inserts at the same boundary
(matches Google Docs). Verify against `posRangeToIndexRange` with a
test; wrap if SDK semantics differ.

## Nice-to-have

- State the invariant in Data Flow: anchor is canonical, resolved
  `DocPosition` is a cache invalidated on remote-change.
- One line that `pendingCursorPos` stays on absolute offsets for now.
- Add CodeRabbit's two-client selection test.

## Implementation in this PR

Please carry the implementation here rather than opening a new PR — the
context is already in one place. A few notes:

- Follow [CONTRIBUTING.md] for workflow (task plan, verify lanes,
  self-review).
- You mentioned Claude Code helped draft this; please skim
  [§ AI agent–assisted contributions][ai] — sign off on every line,
  match existing conventions in
  `packages/frontend/src/app/docs/yorkie-doc-store.ts` (~15
  `consumePendingCursor()` call sites stay as-is), and disclose AI
  assistance in the PR body.
- I'll review the doc updates and the implementation together once it's
  ready.

[CONTRIBUTING.md]: https://github.com/wafflebase/wafflebase/blob/main/CONTRIBUTING.md
[ai]: https://github.com/wafflebase/wafflebase/blob/main/CONTRIBUTING.md#ai-agentassisted-contributions

semimikoh:
Thanks @hackerwins — implementation is now in this PR alongside the design updates you asked for.

**Design doc** (`docs-local-caret-anchoring.md`):
- Pinned `TreePos` and the `indexRangeToPosRange` / `posRangeToIndexRange` API per the SDK guide; added `lineAffinity?: 'forward' | 'backward'`.
- Noted that header / body / footer share the same `root.content` Tree so `indexRangeToPosRange` handles all three regions uniformly, with the `region` tag retained only to keep fallback region-local.
- Made the affinity rule normative (right for the user's own insert, left for a remote insert at the same boundary) and added the matching unit test.
- Stated the canonical invariant in *Data Flow* (anchor is canonical, resolved `DocPosition` is a cache invalidated on remote change) and noted that `pendingCursorPos` stays on absolute offsets for now.
- Added a *Known Follow-Ups* entry for `splitBlock` / `mergeBlock`: today's `delete + insert` paths cannot preserve `TreePos` through removed content, so an anchor that lands in deleted text takes the deterministic fallback ladder rather than CRDT-preserved positions. Switching to native `splitLevel=2` is the natural follow-up and would also need to address the undo/redo regression previously documented.

**Implementation**:
- `YorkieDocStore` captures `AnchoredDocPosition` / `AnchoredDocRange` on cursor change and resolves them on every remote change; the fallback ladder walks same block → previous region block end → next region block start → first editable region/body block.
- IME `composition.startPosition` is anchored end-to-end: `TextEditor` exposes `onCompositionStart` / `onCompositionEnd`, the collab layer captures the anchor, and the resolved position is pushed back through `EditorAPI.updateCompositionStartPosition` before the next composing-text replacement.
- Yorkie-specific anchor state stays inside `packages/frontend`; `packages/docs` still receives only resolved `DocPosition` values, and the ~22 `consumePendingCursor()` call sites are unchanged.
- Added CodeRabbit's two-client non-collapsed selection test, plus body / header / footer / table-cell / split / merge / fallback unit coverage.

**Logistics**:
- Rebased onto `main` to clear the conflict (`docs/design/README.md` had a parallel entry, no other functional conflicts).
- Disclosed AI assistance in the PR body per CONTRIBUTING.md — Claude Code helped with structure, edge-case enumeration, the anchor data flow, IME wiring, and tests; every line was reviewed by me.

One small question on logistics: do you want the PR title updated to reflect that this is no longer design-only (e.g. `Anchor local caret to Yorkie Tree positions (#237)`)? Happy to either way.

hackerwins (approved):
Thanks for your contribution.

PR #281, opened 2026-05-22
Fix SUMPRODUCT boolean coercion

### LINKED ISSUE #196: `SUMPRODUCT()` does not coerce booleans to numbers
### What happened:
`=SUMPRODUCT((A1:A3="North")*B1:B3)` returns `#VALUE!` instead of the filtered sum. The boolean array from the comparison `(D2:D31="North")` is not coerced to `0`/`1` before multiplication, so the product step crashes.

`(A1:A3="North")` should coerce `TRUE→1`/`FALSE→0` before multiplying.

**Setup:** A1=`North`, A2=`South`, A3=`North`; B1=`10`, B2=`20`, B3=`30`.

| Cell | Formula                                | Expected | Actual        | Status |
| ---- | -------------------------------------- | -------- | ------------- | ------ |
| C1 | `=SUMPRODUCT((A1:A3="North")*B1:B3)` | `"40"` | `"#VALUE!"` | ✗ |

### Environment:

- Operating system:
- Browser and version:

emanueledenaro:
Closes #196.

## Summary
- evaluate range comparisons as arrays so expressions like `(A1:A3="North")` can be used inside `SUMPRODUCT`
- support array arithmetic for the comparison result multiplied by a range
- flatten array results in `SUMPRODUCT` while coercing booleans to `1`/`0`
- add regression coverage for `SUMPRODUCT((A1:A3="North")*B1:B3)`

## Tests
- `pnpm --filter @wafflebase/sheets exec vitest --run test/formula/formula.test.ts -t "SUMPRODUCT"`
- `pnpm --filter @wafflebase/sheets test`
- `git diff --check`

Note: the repo pre-push `verify:self` hook was also attempted. `sheets:build`, `docs:build`, and `slides:build` passed, but `verify:fast` failed in unrelated frontend docs tests with `localStorage.getItem is not a function` in this local Node environment.

### DIFF
--- a/packages/sheets/src/formula/formula.ts
+++ b/packages/sheets/src/formula/formula.ts
@@ -32,6 +32,8 @@ import {
   parseRef,
   parseRange,
   isCrossSheetRef,
+  parseCrossSheetRef,
+  toSref,
 } from '../model/core/coordinates';
 
 /**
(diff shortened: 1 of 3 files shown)


ggyuchive:
Hi, because of this issue is fixed , let me close this pr.

PR #339, opened 2026-06-04
Exit the list on Backspace in an empty list item in Docs

### LINKED ISSUE #338: Backspace on an empty list item does not remove the list (Enter does)
### What happened:

In the document editor, applying a numbered (or bulleted) list to an empty paragraph and then pressing Backspace to remove the list does nothing. The list marker / numbering stays and the block is not converted back to a normal paragraph, so there is no way to exit the list with Backspace.

Notably, **Enter on an empty list item already works correctly** — it exits the list and converts the item back to a normal paragraph. Only the Backspace path is missing this behavior, so the two are inconsistent.

Pressing Backspace at the start (offset 0) of an empty list item should remove the list formatting and convert the item back to a normal paragraph — matching Google Docs / Notion and matching what Enter already does in this editor.

2. In an empty paragraph, apply a numbered list (Numbering) from the toolbar
3. The caret is now in an empty list item (offset 0)
4. Press Backspace
5. Expected: the list formatting is removed and the block becomes a normal paragraph
6. Actual: nothing happens — the block stays a list item with its marker/number

Control: press Enter instead of Backspace on the empty list item → it correctly exits the list (works as expected).

- The asymmetry is a strong hint: Enter on an empty list item exits the list, but Backspace does not. The "exit empty list item → paragraph" handling appears to exist on the split (Enter) path but is missing on the delete-backward (Backspace) path.
- Likely localized to the docs delete-backward handling; a non-list empty paragraph handles Backspace normally, so this is specific to list items.
- AI assistance: drafted with help from Claude Code (repro phrasing and code-path hint). I reproduced the issue myself and confirmed the behavior.

### Environment:

- Operating system: macOS Sonoma 14.6.1
- Browser and version: Chrome 148.0.7778.215 (arm64)

semimikoh:
## Summary

Design note for issue #338: pressing Backspace at the start of an **empty list
item** should exit the list (convert it back to a paragraph), matching what Enter
already does. Today it does not — it is a no-op when the list item is the first
block (e.g. an empty document with numbering), or it merges into the previous
block otherwise.

This is the **design-note-first step** (no code yet) — opening for direction
before implementing. Implementation will follow on this same branch.

## Why

The two paths live in `packages/docs/src/model/document.ts`. Enter →
`splitBlock()` has an early branch that converts an empty `list-item` to a
`paragraph` to exit the list. Backspace → `deleteBackward()` has **no** such
branch: at `offset 0` it goes straight to "merge with previous", and the
`if (blockIndex <= 0) return pos` guard returns first when the item is the only
block — which is exactly why the empty-document case does nothing.

The note proposes mirroring `splitBlock`'s branch in `deleteBackward`, placed
**after** the `offset > 0` return and **before** the `blockIndex <= 0` guard so it
also fires for the first/only block. It flags nested-list outdent (`listLevel > 0`)
as a non-goal / open question (matching Enter's simpler behavior for now).

## Test plan

No code changes yet — design note only. The note's Test Plan covers the
implementation that will follow (mirrors the existing `splitBlock` empty-list-item
test in `packages/docs/test/model/document.test.ts`): empty list item as first
block, as non-first block, non-empty list item unchanged, mid-text delete
unchanged.

## Notes for Reviewers

AI tools assisted with this PR: the code investigation and the note draft were
produced with Claude Code (Claude Opus 4.8). I reviewed the root-cause analysis
and every line of the note.

Fixes #338.

## Summary by CodeRabbit

* **Bug Fixes**
  * Backspace on an empty list item now exits the list by converting it to a paragraph, preserving the caret position and avoiding unintended merges.

* **Documentation**
  * Added a design doc describing the Backspace-empty-list-item behavior, caret expectations, edge cases, risks, and a test plan; updated the Docs index.

* **Tests**
  * Added unit tests validating the new Backspace behavior for empty list items.

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -48,6 +48,7 @@ Word processor engine — rich text, tables, pagination, collaboration.
 | [docs-pending-inline-style.md](docs/docs-pending-inline-style.md)                | Pending inline style at a collapsed caret — stored marks for toolbar toggles, IME-aware, view-local            |
 | [docs-local-caret-anchoring.md](docs/docs-local-caret-anchoring.md)              | Local caret anchoring — Yorkie Tree-anchored caret/selection, resolves to DocPosition at render time (issue #237) |
 | [docs-ime-undo-history.md](docs/docs-ime-undo-history.md)                        | IME undo history — coalesce one composed character into a single Yorkie undo unit, fix Hangul undo toggle (issue #318) |
+| [docs-list-item-backspace-exit.md](docs/docs-list-item-backspace-exit.md)        | Backspace on an empty list item exits the list (→ paragraph), mirroring Enter/splitBlock (issue #338) |
 
 ## Slides
 
--- a/docs/design/docs/docs-list-item-backspace-exit.md
+++ b/docs/design/docs/docs-list-item-backspace-exit.md
@@ -1,0 +1,142 @@
+---
+title: docs-list-item-backspace-exit
+target-version: 0.4.4
+---
+
+<!-- Small single-file fix; note kept lightweight (issue #338). -->
+
+# Docs List-Item Backspace Exit
+
+## Summary
+
+Pressing Backspace at the start of an **empty list item** should exit the list —
+convert the item back to a normal paragraph — matching what Enter already does
+and matching Google Docs / Notion. Today Backspace does not do this (issue #338):
+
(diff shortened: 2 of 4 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #415, opened 2026-06-25
Render JSON and date query results as readable sheet cells

ggyuchive:
## Summary
ReadOnlyStore rendered every datasource query value with String(val), which turned Dates into locale-dependent strings and jsonb/array/object values into useless [object Object] cells.
Add a toCell() helper that maps Date to an ISO string and objects to JSON, falling back to String() for scalars, and use it when filling the grid.

Also stop pg from parsing DATE/TIMESTAMP/TIMESTAMPTZ into JS Date :
those parsers shift values by the Node runtime timezone, so a date-only DATE rendered as the previous day in UTC+9 and the test only passed under the local timezone. Return them as raw Postgres text via a per-client type parser and pin the session to UTC with ISO DateStyle, so cells show the stored value and rendering is independent of both runtime and server configuration.

## Why
ReadOnlyStore rendered every datasource query value with String(val), which turned Dates into locale-dependent strings and jsonb/array/object values into useless [object Object] cells.

## Linked Issues

Fixes #

## Author checklist

- [X] I searched existing issues and PRs and confirmed this is not a duplicate.
- [X] I read every line of this diff and can explain why each change is necessary.
- [X] Changes follow the conventions in the touched packages; any deviations are explained in *Why* above.
- [X] If AI tools assisted with this PR, I noted where in *Notes for Reviewers* below.

## Verification

Skip reason (if applicable):

## Risk Assessment

- User-facing risk:
- Data/security risk:
- Rollback plan:

## Notes for Reviewers

- Follow-up work (if any):

## Summary by CodeRabbit

* **New Features**
  * Improved database query results so date and time values display consistently in sheets.
  * Expanded conversion support for PostgreSQL built-in types when loading query results.

* **Bug Fixes**
  * Fixed timezone-related shifts in returned date/time values.
  * Updated cell rendering so complex values are shown in a more readable format instead of raw strings.

### DIFF
--- a/packages/backend/src/datasource/datasource.service.spec.ts
+++ b/packages/backend/src/datasource/datasource.service.spec.ts
@@ -126,14 +126,11 @@ describe('DataSourceService', () => {
     });
 
     expect(client.connect).toHaveBeenCalledTimes(1);
-    expect(client.query).toHaveBeenNthCalledWith(
-      1,
-      "SET statement_timeout = '30000'",
-    );
-    expect(client.query).toHaveBeenNthCalledWith(
-      2,
-      'SELECT * FROM (SELECT id FROM users) AS _q LIMIT 10001',
-    );
+    expect(client.query).toHaveBeenNthCalledWith(1, "SET statement_timeout = '30000'");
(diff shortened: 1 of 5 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #507, opened 2026-07-21
Align mixed-size text runs on a common baseline

dlgpdmsly2:
## Summary

On a line containing runs of different font sizes, smaller glyphs floated
upward instead of sharing a baseline with the larger runs (Google Docs
aligns mixed-size text along the bottom).

**Root cause:** `renderRun` (`view/paint-layout.ts`) derived the alphabetic
baseline from **each run's own** font size, so a smaller run computed a
smaller `baselineY` and sat higher than its neighbours.

**Fix:** derive the base baseline from the line's **tallest** run, stored as
`LayoutLine.maxFontSizePx` — a value `assignLineHeights` already computes
for line height and previously discarded. Super/subscript shifts stay keyed
to the run's own size (they move a run relative to itself). A
`?? originalFontSizePx` fallback keeps uniform-size lines byte-identical to
before.

**Scope:** the change is in the shared canvas painter, so it fixes both the
docs body (via `doc-canvas.ts`) and slides text boxes (via `paintLayout`).
Table cells (`table-renderer.ts`) and PDF export (`pdf-painter.ts`) carry
the same duplicated formula and are an intentional follow-up, tracked in
`docs/tasks/active/20260721-docs-mixed-fontsize-baseline-todo.md`.

## Test plan

- [x] New `packages/docs/test/view/paint-layout.test.ts`: a mock canvas
      context captures the real `baselineY` passed to `fillText` — asserts
      mixed-size runs share one baseline, the tallest run is unchanged when
      a smaller run joins, the smaller run drops to the shared baseline, and
      a uniform-size line is unchanged (regression guard).
- [x] `packages/docs/test/view/layout.test.ts`: `maxFontSizePx` equals the
      tallest run (order-independent) and equals the run size on a uniform
      line.
- [x] `@wafflebase/docs` typecheck + full test suite green (80 files, 1099
      tests).
- [x] `@wafflebase/slides` typecheck + tests green (shares the painter).
- [x] Manual: standalone docs playground — 36pt then 11pt on one line now
      rest on a common baseline.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vertical alignment for text runs with mixed font sizes on the same line by sharing a common baseline.
  - Prevented smaller runs from rendering too high while keeping superscript/subscript positioning correct.
  - Applied the same baseline behavior to body content and header/footer rendering.
  - Preserved the prior baseline calculation when line max sizing isn’t available.

- **Tests**
  - Added layout and rendering tests to validate shared-baseline behavior across mixed and uniform font sizes.

- **Documentation**
  - Added a new note describing the baseline issue and the rendering lessons learned.

### DIFF
--- a/docs/tasks/active/20260721-docs-mixed-fontsize-baseline-lessons.md
+++ b/docs/tasks/active/20260721-docs-mixed-fontsize-baseline-lessons.md
@@ -1,0 +1,50 @@
+# Lessons — Mixed font-size runs share a common baseline
+
+## What broke
+
+On a line with runs of different font sizes, smaller glyphs floated upward
+instead of sharing a baseline (Google Docs aligns them along the bottom).
+`renderRun`'s baseline formula `lineY + (lineHeight + fontSizePx * 0.8) / 2`
+mixed a **line-wide** term (`lineHeight`, sized to the tallest run) with a
+**per-run** ascent term (`originalFontSizePx * 0.8`). The two references
+disagreed, so each run centered on its own size and landed on a different
+baseline.
+
+## Lessons
+
+- **When a formula mixes a line-wide term with a per-run term, suspect the
+  mismatch.** The bug wasn't a wrong constant — `lineHeight` was already
+  line-wide and correct. The defect was that the *ascent* half of the same
+  expression used the run's own size. Fix = make both halves share one
+  reference (the line's max font size), not tweak the multiplier.
+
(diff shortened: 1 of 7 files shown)


harrykim8672:
@coderabbitai review

hackerwins:
@coderabbitai review this PR please.

hackerwins (approved):
Thanks for your contribution.

PR #515, opened 2026-07-22
Share one line-baseline helper across all docs renderers

dlgpdmsly2:
## Summary

Follow-up to #507 (mixed-size baseline fix), which was scoped to the screen
text painter. The same alphabetic-baseline formula
`lineY + (lineHeight + fontSizePx * 0.8) / 2` was copy-pasted across **six
sites in three files**, each keyed to the run's / marker's own font size — so
table cells, PDF export, and list markers still floated smaller runs above a
taller line on the same row.

This extracts one shared helper — `lineBaselineY(lineY, lineHeight,
maxFontSizePx)` in `view/theme.ts` — as the single source of truth, and routes
every renderer through it using the line's max font size:

- `paint-layout.ts` — `renderRun` (dedupe) + `renderListMarker` (new
  `lineMaxFontSizePx` param; the marker is still drawn at its own size, but its
  baseline now comes from the line max).
- `table-renderer.ts` — cell text run + list marker.
- `pdf-painter.ts` — `paintRun` (new param, threaded from the caller) + both
  marker sites. Kept **unrounded** (PDF uses continuous coords; rounding drifts
  the baseline between pages).
- `doc-canvas.ts` — passes `pl.line.maxFontSizePx` to the marker call.

`LayoutLine.maxFontSizePx` (added in #507) already reached these paths because
they render the same `LayoutLine` objects, so this touches painters only — no
layout changes. Uniform-size content is unchanged (`maxFontSizePx ≈ own size`
for a single-size line, plus a `?? own size` fallback).

## Test plan

- [x] `theme.test.ts` — `lineBaselineY` formula, the unrounded contract, and
      taller-max-lowers-baseline.
- [x] `render-list-marker.test.ts` — marker baseline now derives from the line
      max, not the marker's own size.
- [x] Existing `table-renderer` / `pdf-painter` suites pass unchanged
      (regression guard: uniform content renders identically).
- [x] `@wafflebase/docs` typecheck + full suite (81 files, 1105 tests).
- [x] `@wafflebase/slides` typecheck + tests (333 files, 2609) — consumes the
      painter via `paintLayout`.
- [x] Manual: docs standalone playground renders normally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vertical alignment of text and list markers across canvas, table, and PDF rendering.
  * Ensured lines containing mixed font sizes use the largest font size for consistent baseline positioning.
  * Preserved precise, unrounded positioning for PDF output.

* **Tests**
  * Added coverage for baseline calculations and list-marker alignment, including mixed-size text scenarios.

### DIFF
--- a/docs/tasks/active/20260722-docs-baseline-dedup-lessons.md
+++ b/docs/tasks/active/20260722-docs-baseline-dedup-lessons.md
@@ -1,0 +1,38 @@
+# Lessons — De-duplicate the line baseline formula
+
+## What broke
+
+Nothing new broke — this is the structural fix for *why* the #507 bug was
+possible. The baseline formula lived in six copies across `paint-layout`,
+`table-renderer`, and `pdf-painter`, each with a comment like "matching
+doc-canvas's body path." Hand-synced duplication drifts; #507 was one copy
+being fixed while the others kept the old per-run behaviour.
+
+## Lessons
+
+- **Fix the duplication, not just the instances.** The reviewer could have
+  asked for the same three-line change pasted into two more files. Extracting
+  `lineBaselineY` means the next baseline change happens once, and the
+  "matching X" comments (which were load-bearing documentation of an invariant
+  the compiler didn't enforce) become an actual shared function.
+
+- **A shared helper must respect the callers' real differences.** The canvas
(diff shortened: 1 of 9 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #508, opened 2026-07-21
Review panel: one orchestrator, four reviewer subagents

harrykim8672:
## Summary

Follow-up to the autonomous pipeline (#501, merged): replaces the single
independent reviewer with a **data-driven review panel** of specialized lenses,
run by **one orchestrator process that spawns a subagent per lens** (Claude Agent
SDK) — not four separate Actions jobs.

- `scripts/agent/review-panel.mjs` — orchestrator: parallel per-lens subagents
  (read-only), a **per-finding verifier refute pass** to cut false positives,
  dedup/synthesis, and **trusted** per-lens verdict computation (subagents only
  classify; the script decides the gate).
- `scripts/agent/lenses/` — `lenses.json` + rubric per lens: **correctness,
  security, design-fit, test-adequacy**. Adding a lens = a config entry + a `.md`.
- `scripts/agent/severity.mjs` — the block-iff-critical/major rule, now shared by
  the orchestrator and `read-review-verdict.mjs` (one source of truth).
- `agent-independent-review.yml` → `agent-review-panel.yml`: single `review-panel`
  job posting one unforgeable `agent-review-<lens>` **check run** per lens;
  `promote`/`fix`/`stalled` generalized to the panel.
- `mark-ready.mjs` gains `--require-checks` (all lens checks must pass).
- `agent-sdk-smoke-test.yml` + `scripts/agent/auth-smoke.mjs` — a one-press
  `workflow_dispatch` **live credential check** (see *Pre-arm*).

## Why

One generalist reviewer is shallow per-concern and shares one set of blind spots.
Focused lenses catch more (a security-only lens scrutinizes authz harder), and the
`design-fit` + `test-adequacy` lenses cover failure classes line-level review
misses (spec-drift, duplication, vacuous tests). Structure suggested by the
maintainer: dynamic-workflow + subagents in one process (efficient — the diff is
ingested once; enables synthesis + the verifier pass).

## Design decisions

- **Block-on-concrete:** lenses block only on cited, defensible critical/major;
  taste → advisory minor/nit.
- **Trust preserved from #501:** reviewers never execute branch code (SDK tools =
  Read/Grep/Glob); orchestrator + rubrics run from a **trusted `main` checkout**;
  check runs are author-unforgeable; fork-origin rejected; **fail-closed**.
- **design-fit** reads the originating issue (`Fixes #N`) for spec-conformance,
  only when a **human** filed it through the agent-task flow (labelled
[...]

PR #519, opened 2026-07-22
Fix review-panel SDK binary launch (preserve exec bits across the artifact)

harrykim8672:
## Summary

Fixes the review panel's first real run. On PR #513 the `review-panel` job ran
end-to-end but **every lens failed closed** with:

> `Claude Code native binary at …/@anthropic-ai/claude-agent-sdk-linux-x64/claude exists but failed to launch`

so no lens produced a verdict and the panel paged instead of reviewing.

## Root cause

`actions/upload-artifact` **does not preserve Unix file permissions**. The `deps`
job uploads `scripts/agent/node_modules` raw, which strips the `+x` bit from the
SDK's bundled native binaries (the `claude` CLI, ripgrep). The `review-panel` job
downloads a binary that is *present but non-executable* → spawn fails → the SDK
emits its generic "libc mismatch?" guess, but it's actually the lost exec bit.

Evidence it's **not** a wrong-variant problem: the lockfile pins per-libc variants
(`linux-x64` = glibc, `linux-x64-musl` = musl), and the failing binary is at the
**glibc** `linux-x64` path — i.e. the correct binary for `ubuntu-latest` was
installed. Only its permission bit was dropped in transit.

## Fix

Tar `node_modules` in the `deps` job (tar preserves perms **and** symlinks),
upload the tarball, and untar it in the `review-panel` job. No change to the
no-install-with-secrets rule — install still happens only in the unprivileged
`deps` job.

## Verification

- `node -c` / YAML parse clean.
- The definitive proof is the next armed run: the `agent-review-*` checks should
  post real verdicts (approved / changes-requested with findings) instead of
  "Reviewer did not produce a valid verdict."

## Context

Found while validating the merged pipeline (#508) on agent PR #513: environment
gate cleared, panel triggered, checks posted, fail-closed worked — the only
defect was this binary-launch issue. This unblocks the panel producing real
verdicts (and, in turn, the auto-fix loop).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Agent SDK setup in review workflows by preserving executable permissions and symbolic links.
  * Increased reliability when preparing the review panel environment.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -51,12 +51,17 @@ jobs:
         with:
           node-version: 22.x
       - run: cd scripts/agent && npm ci --no-audit --no-fund # reproducible: exact tree from the committed lockfile
+      # Tar before uploading. actions/upload-artifact does NOT preserve Unix file
+      # permissions (or symlinks), so a raw node_modules upload strips the +x bit
+      # from the SDK's bundled native binaries (the `claude` CLI, ripgrep). The
+      # review job then finds the binary present but non-executable → "binary
+      # exists but failed to launch". A tarball round-trips perms + symlinks intact.
+      - run: cd scripts/agent && tar -czf /tmp/agent-sdk-node-modules.tgz node_modules
       - uses: actions/upload-artifact@v4
         with:
           name: agent-sdk-node-modules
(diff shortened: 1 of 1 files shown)


hackerwins (approved):
Thanks.

PR #523, opened 2026-07-22
Restore selection range on docs undo/redo, not just the caret

### LINKED ISSUE #340: Undo restores the caret but not the selection range
### What happened:

In the document editor, undo restores the **caret position** but not the **selection range** that was active before the edit. After undoing, only the caret is placed; any previously selected range is lost.

The history snapshot records `activeCursorPos` (the caret) but does not record `activeSelection` (the range), so on undo there is nothing to restore the selection from.

Undo should restore both the caret and the selection range that were active at that history step, matching Google Docs / typical editors (undo brings back the prior selection, not just a collapsed caret).

2. Type some text, then select a range of it
3. Perform an edit that replaces/transforms the selection (e.g. type over it, or apply formatting)
4. Press undo (Cmd/Ctrl+Z)
5. Expected: the previous selection range is restored
6. Actual: only the caret is restored; the selection range is gone

- Localized to the docs history path in `yorkie-doc-store.ts`: the undo entry stores `activeCursorPos` only, never `activeSelection`.
- AI assistance: drafted with help from Claude Code (repro phrasing and code-path hints). I reproduced the issue myself and confirmed the behavior.

### Environment:

- Operating system: macOS Sonoma 14.6
- Browser and version: Chrome (arm64)

dlgpdmsly2:
Fixes #340.

## Summary

Undo restored the **caret** but collapsed the **selection range** — after
undo you got a bare caret, losing the highlight (Google Docs restores the
selection). Two causes:

1. The docs Yorkie store recorded only `activeCursorPos` in each mutation's
   `addToHistory` presence set — never `activeSelection`. Yorkie only reverses
   the keys present in the set, so the selection was never part of an undo
   unit.
2. The editor's `undoFn`/`redoFn` only moved the caret (`cursor.moveTo`),
   never re-applied a range.

This PR:

- **Store** (`yorkie-doc-store.ts`): a shared private `recordHistoryPresence`
  helper now records **both** `activeCursorPos` and `activeSelection` with
  `{ addToHistory: true }` — replacing all 21 recording sites. `setCursorForHistory`
  is widened to take the pre-edit selection and flush it to presence
  **synchronously** (Yorkie captures the *current* presence as the reverse, so
  the pre-edit selection must be in presence at mutation time; the live
  publish is throttled). New `getPresenceSelection()` reader mirrors
  `getPresenceCursorPos()`. `activeSelection` defaults to a **concrete
  collapsed range** at the caret (never `undefined`) so the presence key is
  always tracked and reads as "no selection".
- **Editor** (`editor.ts`): both `setCursorForHistory` call sites pass the
  current selection range; `undoFn`/`redoFn` re-apply the restored range via a
  shared `restoreSelectionFromPresence()` (guards block existence, clamps
  offsets — mirroring the caret restore; collapsed/missing/stale → clears).

## Test plan

- [x] New store unit tests (pure in-memory Yorkie, **no docker/backend**) in
      `yorkie-doc-store.test.ts`: undo restores the range; redo collapses it;
      `applyStyle` undo restores the styled range; two-unit type-over
      (`deleteText` + `insertText`) restores the original selection on the
      second undo. Existing caret-restore tests unchanged.
- [x] `@wafflebase/docs` typecheck + full suite (1105 tests).
- [x] `@wafflebase/frontend` typecheck + full suite (862 tests).

## Notes

- Typing over a selection is **two undo units** (`deleteText` then
  `insertText`) by existing design, so the original selection returns on the
  *second* undo — matches Google Docs.
- Deferred (R4): redo of a style op shows a caret rather than re-highlighting
  the range; optional follow-up passes the styled range to the helper.
- Manual in-app repro needs the full stack; covered here by store unit tests.

## Summary by CodeRabbit

* **New Features**
  * Enhanced editor undo/redo to preserve and restore highlighted text selections (not just caret position).
  * Type-over edits now restore the original selection when fully undone.
  * Redo selection behavior follows the updated semantics, including appropriate collapsing.
* **Bug Fixes**
  * Fixed selection highlights disappearing or collapsing incorrectly during undo/redo after formatting and other edits.
* **Tests**
  * Added regression tests for selection restoration via toolbar styling and type-over flows.
* **Documentation**
  * Updated undo/redo selection documentation with the failure mode and key guidance.

### DIFF
--- a/docs/tasks/active/20260722-docs-undo-selection-lessons.md
+++ b/docs/tasks/active/20260722-docs-undo-selection-lessons.md
@@ -1,0 +1,43 @@
+# Lessons — Undo/redo selection restore (#340)
+
+## What broke
+
+Undo history recorded the caret (`activeCursorPos`) but not the selection
+(`activeSelection`), and the editor's undo/redo only moved the caret. So undo
+placed a bare caret and dropped the highlight.
+
+## Lessons
+
+- **Yorkie presence-with-history reverses only the keys you set.** `p.set(x,
+  { addToHistory: true })` records a reverse for exactly the keys in the
+  partial. `activeSelection` was never in the partial, so undo never restored
+  it. The fix is simply to include the key — the "restored value" is Yorkie's
+  reverse (the pre-edit presence), not something we stash.
+
+- **Set a concrete collapsed range, never `undefined`.** A key set to
+  `undefined` can be dropped during (esp. networked) presence serialization,
+  so its reverse may not be recorded. A collapsed range (`anchor === focus`)
+  guarantees the key is tracked and reads as "no selection" (`hasSelection()`
+  is false for it), which is the correct post-edit/redo state.
+
+- **The reverse is captured from *current* presence, so timing matters.** The
+  live cursor publish is throttled, so "select then immediately type" could
+  record a stale selection. `setCursorForHistory` now flushes the pre-edit
+  selection synchronously before the mutation. This was the non-obvious race
+  (R2) — the fix only works because the pre-edit selection is guaranteed to be
+  in presence at mutation time.
+
+- **Restoring presence ≠ restoring the view.** Undo puts the range back into
+  Yorkie presence, but the editor's `Selection` object (the visible
+  highlight) doesn't update itself — `undoFn`/`redoFn` must read it back and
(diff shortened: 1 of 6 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #526, opened 2026-07-23
Fix loops: page a human on a no-commit round; bump fixer turns 40->80

harrykim8672:
## Summary

Stop the autonomous fix loops from **silently dead-ending** when a fix round
produces no commit, and give the fixer more room to converge.

Found on #521: the review-fix loop ran, the fixer was fed the correctness
finding, but it burned its 40-turn budget (with 3 permission denials) and pushed
**no commit**. No commit → no new CI → no new panel → no next round — and the
fix job still reported `success`, so nothing paged. The PR sat blocked, unfixed,
and un-escalated.

## Changes (both fix loops: `agent-review-panel.yml` + `agent-iterate-ci.yml`)

1. **Page a human on a no-commit round.** Record the branch head before the
   fixer; after it, if the head didn't advance, post the loop's paged latch +
   `agent:needs-human-review` label. Deliberately **not** a blind retry — the
   fixer just failed to converge (turn budget / denials), so re-running the same
   config would burn budget and likely fail again; a human should take over.
2. **Bump `--max-turns` 40 → 80.** A fix that must keep `verify:fast` (build +
   tests) green needs more room than 40 — #521's fixer hit exactly 40.
3. **Upload the fixer's execution log** as an artifact so turn usage and
   permission denials are debuggable (they weren't visible in the plain job
   log on #521).

## One deliberate difference between the two loops
- **Review-fix** page step omits `always()`: a *hard* fixer error falls through
  to that workflow's existing `stalled` net (avoids a double-page).
- **CI-fix** page step uses `always()`: `agent-iterate-ci.yml` has **no** stalled
  net, so a hard fixer error must page there directly.

## Validation
Both workflow YAMLs parse; `verify:fast` pre-commit hook green; the SDK/agent
paths run only in real CI. The end-to-end proof is a blocked PR whose fixer
commits nothing → it should now page instead of stalling, with the transcript
artifact attached.

CODEOWNERS-protected (`.github/workflows/agent-*.yml`) → requests @hackerwins'
review.

## Note
Touches `agent-review-panel.yml`, which the `agent/fix-promote-token` branch also
edits — whichever merges second will need a small rebase.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automated fix workflows to detect stalled runs by checking whether the branch head advances after the fixer.
  - Added automatic “needs human review” paging when no new commit is produced, preventing silent dead ends.
  - Enhanced review-panel convergence by re-checking and merging prior blocking findings across rounds, reducing false negatives.

- **Chores**
  - Improved CI and review-panel observability with always-uploaded fixer transcripts and more machine-readable per-lens outputs.
  - Increased fixer capacity and expanded lens sampling to reduce single-run non-determinism.
  - Updated documentation to reflect the review verifier’s false-negative hardening behavior.

### DIFF
--- a/.github/workflows/agent-iterate-ci.yml
+++ b/.github/workflows/agent-iterate-ci.yml
@@ -156,6 +156,14 @@ jobs:
             echo "HARNESS_EOF"
           } >> "$GITHUB_OUTPUT"
 
+      # Record the branch tip before the fixer runs, so the step after it can
+      # detect a fix round that produced NO commit (see "Page if the fix
+      # produced no commit").
+      - name: Record branch head before fix
+        id: before-fix
+        if: steps.guard.outputs.proceed == 'true'
+        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
       - name: Run Claude Code
         if: steps.guard.outputs.proceed == 'true'
         # Pinned to the v1 release commit (immutable) for supply-chain safety.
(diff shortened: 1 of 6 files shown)


harrykim8672:
Addressed the review (pushed):

- **Major — local vs remote HEAD:** valid, real gap. The detector compared local `git rev-parse HEAD`, but the loop only re-triggers on a *push*, so a commit-but-no-push round (turn budget exhausted after commit, or a push denial) was mis-read as progress and didn't page. Both loops now capture the **remote** tip via `git ls-remote origin refs/heads/$BRANCH` before the fixer and compare after — "committed locally but never pushed" now correctly pages.
- **Minor — iterate-ci message:** valid. That page step runs under `always()` (it's the CI-fix loop's only safety net), so it can fire on a hard error too. Reworded to a cause-agnostic message that also names "an earlier step errored" and points at both the run log and the transcript artifact.
- **Minor — test adequacy:** the detection is now trivial SHA-equality shell; the real risk is the git/CI integration, which a mocked unit test wouldn't exercise. Taking the reviewer's offered alternative: an **end-to-end dry-run on a throwaway `agent/` branch is a merge prerequisite** (below)

harrykim8672:
@coderabbitai review

hackerwins (dismissed):
Thanks for your contribution.

hackerwins (dismissed):
Thanks for your contribution.

PR #533, opened 2026-07-23
Pin note code-block copy button and refine code styling

### LINKED ISSUE #518: Add background to inline/block code and hide raw backticks in Note
### Description:
I would like to suggest a visual improvement for the rendering of `inline code` and
```
code blocks.
```

Additional Issue:
There is also a minor rendering issue with inline code. When wrapping text in backticks (e.g., `code`), the backticks themselves remain visible in the rendered output. They should ideally be hidden once the markdown is parsed.

### Why:
Currently, when formatting text as code, only the text style changes. It lacks a background color, making it somewhat difficult to distinguish the exact boundaries of the code elements from regular text. It would be great to add a background color to these elements so their areas are easily recognizable. For dark mode, the background rectangle should be slightly lighter to provide proper contrast.

hotsunchip:
## Summary

Fixes styling of markdown code blocks in the Notes preview.

- **Copy button no longer drifts.** The "Copy" button was a child of the
  horizontally-scrolling `<pre>`, so a long code line scrolled it out of view.
  The fenced block is now wrapped in a non-scrolling `.note-code-wrapper`, with
  the button a *sibling* of the `<pre>` (wrapper owns positioning, `<pre>` owns
  overflow), so it stays pinned to the block's top-right.
- **Inline code** gets a subtle neutral pill (`:not(pre) > code`, light + dark),
  and the decorative backtick pseudo-elements `@tailwindcss/typography`'s
  `prose` injects around `<code>` are cancelled (markdown-it already consumed
  the real backticks, so they were duplicate visible cruft).
- **Fenced blocks** get their own background + a thin scrollbar (light + dark),
  and the forced `background: transparent` is removed from the shared `.hljs`
  rule so it only carries the syntax palette.

Copy-button JS wiring is unaffected: the handler resolves `<code>` via
`button.parentElement?.querySelector('code')`, which still finds the nested
`<code>` after the move.

resolves #518 

## Test plan

- `packages/notes/src/view/preview.test.ts` — new assertion that the copy
  button anchors to `.note-code-wrapper` and sits outside the scrolling
  `<pre>`; existing clipboard test transitively covers the handler through the
  new DOM.
- `pnpm --filter @wafflebase/notes typecheck` — green.
- Notes preview suite — green (`preview.test.ts` passes).
- Manual smoke recommended in `pnpm dev`: confirm the button sits pinned on the
  block (not floating in the margin above) and scrolls stay put with long lines.

> Note: the full local `verify:fast`/`verify:self` gate is unreliable on this
> WSL/jsdom host — vitest fork workers time out on random *unrelated* files
> (`memory`/`editor`/`note-sync`/`text-edit-section`), not on this change.
> Relying on CI as the authoritative gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed note code-block copy buttons drifting when code is horizontally scrolled.
  * Improved styling for inline and fenced code, including backgrounds, spacing, syntax colors, and dark mode.
  * Removed unwanted decorative backticks from inline code.
  * Added coverage to verify copy buttons remain correctly positioned.

### DIFF
--- a/docs/tasks/README.md
+++ b/docs/tasks/README.md
@@ -19,8 +19,10 @@ Track task-specific plan/review and lessons files using the active/archive layou
 | Task | Todo | Lessons |
 |---|---|---|
 | maintainer merge skill (2026-07-23) | [20260723-maintainer-merge-skill-todo.md](./active/20260723-maintainer-merge-skill-todo.md) | [20260723-maintainer-merge-skill-lessons.md](./active/20260723-maintainer-merge-skill-lessons.md) |
+| note code block style (2026-07-23) | [20260723-note-code-block-style-todo.md](./active/20260723-note-code-block-style-todo.md) | [20260723-note-code-block-style-lessons.md](./active/20260723-note-code-block-style-lessons.md) |
 | docs baseline dedup (2026-07-22) | [20260722-docs-baseline-dedup-todo.md](./active/20260722-docs-baseline-dedup-todo.md) | [20260722-docs-baseline-dedup-lessons.md](./active/20260722-docs-baseline-dedup-lessons.md) |
 | docs hyperlink formatting exit (2026-07-22) | [20260722-docs-hyperlink-formatting-exit-todo.md](./active/20260722-docs-hyperlink-formatting-exit-todo.md) | - |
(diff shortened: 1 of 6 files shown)


hackerwins (approved):
Thanks for your contribution.

NewMassMedia:
이욜ㅋ

PR #551, opened 2026-07-24
metrics: repost effort summary at bottom; sweep per-session records on promote

harrykim8672:
## Summary

Follow-up to the review-panel/metrics work, addressing two things surfaced while
watching #547:

1. **The effort summary looked like it never posted.** `metrics.mjs summarize`
   *upserted* the summary comment (edit-in-place), so it stayed pinned at its
   original creation point — often an early *paged* hand-off near the top — and
   the up-to-date summary ended up buried mid-thread.
2. **Blank comment clutter.** Every agent session appends a hidden
   `<!-- agent-metric … -->` record; GitHub renders each as an **empty comment
   box**, so a many-round PR (#547 had 8) fills with blanks.

### Change
`summarize` now:
- **Posts the summary fresh at the bottom** (delete old + POST new) instead of
  editing in place, so the current summary lands where a human looks. Deletes are
  best-effort — metrics never fail the pipeline.
- **On the terminal promote (`--final`) sweeps the hidden per-session records** —
  their totals are already captured in the summary. Wired **only** on the promote
  job's summary step; paged/stalled summaries omit `--final` so a PR that may
  re-run keeps its records to re-aggregate the full history.

`parseArgs` now handles boolean flags (`--final`).

## Test plan
- [x] `cd scripts/agent && node --test *.test.mjs` — 49/49, incl. a guard test
  that the sweep filter (`METRIC_PREFIX`) never matches the summary marker
  (`agent-metric ` vs `agent-metrics-summary`) — so promotion can't delete its
  own summary
- [x] `node --check`; workflow YAML parses; `pnpm verify:entropy` (dead-code 0 —
  removed the now-unused `upsertComment`/`findComment`)
- [ ] End-to-end (armed): on promote, confirm exactly one summary at the bottom
  and no leftover blank `agent-metric` comments

## Summary by CodeRabbit

* **Improvements**
  * Agent metrics summaries are now posted as fresh comments, keeping the latest summary at the bottom of the discussion.
  * Finalized reviews automatically remove hidden per-session metric details while preserving the consolidated summary.
  * Ongoing reviews retain session metrics for future aggregation.

* **Bug Fixes**
  * Prevented cleanup from accidentally removing the newly posted summary.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -442,7 +442,10 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR: ${{ needs.review-panel.outputs.pr }}
-        run: node ./scripts/agent/metrics.mjs summarize --pr "$PR"
+        # --final: terminal promotion → repost the summary at the bottom AND sweep
+        # the hidden per-session agent-metric records. The paged/stalled summaries
+        # omit it (a paged PR may re-run, so its records must survive to re-aggregate).
+        run: node ./scripts/agent/metrics.mjs summarize --pr "$PR" --final
 
   fix:
     needs: review-panel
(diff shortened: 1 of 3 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #557, opened 2026-07-25
Fork-PR comment fix + running→edited comments for review/summarize/loop

harrykim8672:
## Summary

Two related changes to the on-demand `@claude` verbs (`review` / `summarize` /
`loop`):

1. **Fork-PR fix.** These verbs post their result as a PR comment. On a **fork PR**
   the run is triggered by `issue_comment`, and GitHub forces the automatic
   `GITHUB_TOKEN` to **read-only** — so `issues.createComment` returns
   `403 Resource not accessible by integration`. Posting now uses a **GitHub App
   installation token** (the one the promote/fix jobs already use), which isn't
   subject to the fork restriction.
2. **"Running → result" UX.** Each verb now posts a `🤖 working…` placeholder
   comment immediately, then **edits that same comment** into the review / summary /
   loop-result when done.

## Why

Observed on **#550** (a fork PR): [the `@claude review` run](https://github.com/wafflebase/wafflebase/actions/runs/30165338383)
ran the panel fine and failed only on the final comment —
`POST /repos/wafflebase/wafflebase/issues/550/comments → 403`. Every *read* in the
job worked; only the *write* failed, because the fork-triggered `GITHUB_TOKEN` is
read-only regardless of the `permissions:` block.

## What changed

- **`agent-review-on-demand.yml`**: `authorize` posts the placeholder (with the
  throttle marker) via an App token and outputs its id; the `review` job edits it
  into the findings using a second App token minted *after* the read-only panel.
  The review job is `always() && proceed` so a `deps` failure edits the placeholder
  to a fail-closed note instead of stranding it.
- **`agent-summarize.yml`**: split into an `announce` job (gate + placeholder,
  holds the App token, runs no untrusted code) and the `summarize` job (fetch data
  → read-only Claude → mint App token **after** Claude → edit). The token never
  coexists with the model.
- **`agent-loop.yml`**: posts/edits its comment with the App token (so the
  fork-fallback note works on a fork PR); label + CI re-run stay on `GITHUB_TOKEN`
  (same-repo path). Posts a placeholder → does the work → edits with the outcome.
- All edits fall back to a fresh comment if the placeholder id is missing, and run
  under `always()` so a failure still replaces the placeholder.

## Author checklist

- [x] I read every line of this diff.
- [ ] Tests — none (workflow-only). Needs a live re-run of each verb (incl. on a
      fork PR) after merge.
- [x] AI assistance noted below.

## Verification

- All workflows parse as valid YAML; the App-token action pin matches the existing
  promote/fix usage.
- **Effective only after merge to `main`** — `issue_comment` workflows always run
  the default-branch version, so re-running `@claude review` on #550 will keep
  using the old (broken) workflow until this lands.

## Risk Assessment

- App token (scoped `issues: write`) is introduced for comment writes. In `review`
  and `summarize` it is never present in the same job as the untrusted-influenced
  model/SDK step (placeholder posts from a no-untrusted-code job; the result-edit
  token is minted after the model runs). `loop` runs no untrusted code at all.
- Fork-PR review still reviews untrusted code read-only (Read/Grep/Glob, trusted
  `main` orchestrator, pinned SHA) — unchanged.

## Notes for Reviewers

**AI assistance:** authored with Claude Code, reviewed line-by-line. Diagnosed from
the failed run logs of #550's `@claude review`.

## Summary by CodeRabbit

* **Improvements**
  * Workflow status comments are now posted immediately and updated in place with final results.
  * Review, summarization, and loop workflows reliably replace progress messages even when processing fails.
  * Fork pull requests receive clearer outcome messages, including whether CI was rerun.
  * Workflow permissions are more tightly scoped to required actions.

### DIFF
--- a/.github/workflows/agent-loop.yml
+++ b/.github/workflows/agent-loop.yml
@@ -67,7 +67,36 @@ jobs:
               core.setFailed(`@${actor} lacks write access (permission: ${data.permission}); refusing to run.`);
             }
 
+      # Comments are posted/edited with a GitHub App token, NOT GITHUB_TOKEN: on a
+      # fork-PR-triggered run GITHUB_TOKEN is read-only, so the fork-fallback note
+      # would 403. The App token isn't subject to that restriction. Loop runs no
+      # untrusted code (API calls only), so minting it here is safe. Scoped to the
+      # comment API (issues:write); the label + CI re-run stay on GITHUB_TOKEN,
+      # which is writable on the same-repo path where they actually execute.
+      - name: Generate GitHub App token (comments)
+        id: app-token
+        uses: actions/create-github-app-token@d72941d797fd3113feb6b93fd0dec494b13a2547 # v1.12.0
+        with:
+          app-id: ${{ secrets.AGENT_APP_ID }}
+          private-key: ${{ secrets.AGENT_APP_PRIVATE_KEY }}
+          permission-issues: write
+
+      # Post the "working on it" comment immediately, then edit it with the result.
+      - name: Post running placeholder
+        id: placeholder
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ steps.app-token.outputs.token }}
+          script: |
+            const owner = context.repo.owner, repo = context.repo.repo;
+            const { data } = await github.rest.issues.createComment({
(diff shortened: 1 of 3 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #566, opened 2026-07-27
Add issue-category classifier to the agent pipeline

dlgpdmsly2:
## Summary

Adds issue-category tracking to the autonomous issue→PR pipeline so agent performance can be sliced by the *kind* of issue worked on (bug fix vs. feature vs. refactor …, and for bug fixes the root-cause class).

On each `agent-implement` run, `scripts/agent/classify.mjs`:

- Classifies the issue by **primary type** (bug-fix / feature / refactor / performance / security / testing / build-ci / docs) from the issue text, and — for bug fixes that produced a PR — the **root-cause bug class** from the diff (with a cheap two-pass verification) plus localization scope.
- Applies the results as **namespaced GitHub labels** — `type:…` always, `bugclass:…` for verified bug fixes — to the PR, or to the issue when no PR was produced, creating any missing label idempotently. This is the queryable/browsable surface.
- Records the **full JSON** (confidence, rationale, localization, verify flag) behind those labels as a hidden `<!-- agent-class {json} -->` comment — the same PR ledger `metrics.mjs` already uses for agent run info.

Wired in as one `if: always()` step in `agent-implement.yml`, right after the metrics step, so it runs even on failed runs (labels the issue when no PR exists) and keeps "attempted, no PR" rows for the resolve-rate denominator. Fail-safe: `exit 0` on any error, `continue-on-error` on the step.

Read back by listing PRs/issues by label, or by scraping the `agent-class` comments and joining to `agent-metric` by PR number.

## Notes

- `classify.mjs` **shares** `metrics.mjs`'s `gh` / PR-resolution / comment-listing helpers (now exported) rather than duplicating them; the upsert + label logic is classification-specific (the metrics ledger is append-only). Its CLI is guarded so the pure helpers are importable.
- Model tiering: Haiku for the coarse type, Opus for the bug class + verification pass.

## Test plan

- `scripts/agent/classify.test.mjs` covers the pure logic (serialize/parse round-trip, `-->` terminator neutralization, truncation, localization-scope branches): `node --test scripts/agent/*.test.mjs` → green.
- `agent-implement.yml` change is purely additive (one step); no existing step is modified. `metrics.mjs` change is only `export` on four existing helpers.

### DIFF
--- a/.github/workflows/agent-implement.yml
+++ b/.github/workflows/agent-implement.yml
@@ -292,6 +292,18 @@ jobs:
           ISSUE: ${{ github.event.issue.number || github.event.inputs.issue_number }}
         run: node ./scripts/agent/metrics.mjs record --execution "$RUNNER_TEMP/claude-execution-output.json" --kind implement --issue "$ISSUE"
 
+      # Classify the issue by category into a hidden `<!-- agent-class -->`
+      # comment, mirroring the metrics ledger (Dimension A from the issue always;
+      # root-cause bug class from the diff when a PR exists). Fail-safe: exit 0.
+      - name: Classify issue category
+        if: always()
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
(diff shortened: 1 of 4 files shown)


harrykim8672:
@claude review

harrykim8672 (approved):
Thanks for contributing

PR #572, opened 2026-07-28
Add offline replay evaluation harness for the review agent

dlgpdmsly2:
## Offline replay evaluation harness (`scripts/agent/eval`)

Foundation for **reproducible evaluation of the review agent** — replays a frozen
corpus of historical product diffs through a review configuration and measures
**cross-run reliability** (does the same diff, re-reviewed under the same judge,
reach the same gate verdict?). This is the label-free "before" baseline that a
future judge/composition change can be measured against.

### Design
- **Target Adapter seam** — the runner is role-agnostic; the review panel is one
  adapter (`adapters/reviewer.mjs`). A code-fixer or end-to-end target later is a
  new adapter, not a harness rewrite.
- **Run/score separation** — model output is written once (expensive); metrics are
  computed from cached artifacts (cheap, re-runnable). Change a metric → re-score,
  never re-run.
- **Immutable, content-addressed identity** — `config_hash` (the judge) ×
  `corpus_version` (the inputs) is the comparability key; replicate runs share it.
- **Data lives in a separate results repo**, reached through an `ArtifactStore`
  (`GitFsStore`), so this package holds *code only* and the store is relocatable.

### Modules (pure logic unit-tested — 27 tests, `node --test scripts/agent/eval/*.test.mjs`)
| File | Role |
|---|---|
| `config-hash` / `config-build` | judge identity + config manifest & reproduction snapshot |
| `store` (`GitFsStore`) | the immutable `ArtifactStore` |
| `extract-corpus` | freeze historical PR diffs into a versioned corpus (records provenance) |
| `adapters/reviewer` + `run` | role-agnostic runner over `review-panel.mjs` |
| `reliability` | cross-run verdict stability + Fleiss κ |

### Metric note
κ is computed **only on the binary gate verdict** (block/approve) — a fixed item
set × K replicate runs, where chance-corrected agreement is well-defined. It is
deliberately **not** computed over raw finding sets (open-ended detection → the
negative class is unbounded → κ mis-specified).

### Scope / safety
- Standalone package (`@wafflebase/agent-scripts`), **not part of the workspace
  build**; touches no existing pipeline code.
- The runner is the only model-calling path; it needs the SDK installed and
  `CLAUDE_CODE_OAUTH_TOKEN`. Everything else is offline and tested.

### Status
Pure logic is fully tested; the model-calling **runner has not yet had its first
live end-to-end run** (opened as draft for that reason). Fidelity caveat (v1):
replay is diff-only — the panel runs against an empty `--repo`, so lenses reason
from the diff, not surrounding code they could `Read` in production; reliability
under identical conditions is unaffected. Runbook in `scripts/agent/eval/README.md`.


dlgpdmsly2:
Closing while the eval harness is still iterating — moving the churn off upstream. Work continues on the fork branch (dlgpdmsly2:feat/agent-eval-harness); will reopen a clean, squashed PR once there's a working version.

PR #573, opened 2026-07-28
Make the review verifier independent and require grounded refutations

hwisoo-kim:
## Summary

Make the per-finding verifier in `review-panel.mjs` an **independent** check
rather than a second read of the same diff, and require a refutation to be
*grounded* before it may drop a finding.

Script-only, no workflow edits. Prerequisite for the incremental-review PR later
in this series.

## Two defects, same function

**1. The verifier was not independent.** `verifyFinding` received the same diff
as the lens that raised the finding. A lens that misreads a hunk hands that
misreading to a reader of the identical text, which confirms it — the
correlated-error failure mode adversarial-review work identifies as the core
weakness of naive multi-agent panels.

A different *model* would not fix this. The shared input is what makes the errors
correlate, so independence needs a different **evidence source**. And that turned
out to be nearly free: the SDK call already ran with `cwd: repo` and
`Read`/`Grep`/`Glob` allow-listed. The diff was being passed *in addition to*
tools that could reach the real code. Deleting one parameter was most of the fix.

**2. Refuting cost nothing.** The old rule dropped a blocking finding on
`{verdict: "refuted", confidence: "high"}` — two enum fields. `reason` was in the
schema but nothing read it, so the prompt's *"only refute with a concrete
reason"* was prose the trusted script never enforced. On the audited PRs this
pass killed 50–60% of surviving findings on exactly that shape.

## What changed

| change | why |
|---|---|
| `verifyFinding` drops `diff`, takes `changedFiles` | the independence fix above; prompt tells it to distrust the finding's quoted evidence ("checking it IS the job") |
| `VERIFIER_SCHEMA` gains required `refutationGround` + `groundedIn` | turns "refute only with a concrete reason" from prose into a shape trusted code can check |
| new pure export `isDroppingVerdict(v)` | the complete drop rule in one checked place; `applyVerifications` calls it |
| new pure export `changedFileContext(files, max = 200)` | bounds a list re-sent per finding per lens per round, and decides whether it is authoritative |
| `askStructured` gains `maxTurns`; verifier capped at 8 | the verifier now needs tool calls, and that budget multiplies across every blocking finding |
| `verifierTally` gains `dropped`, threaded into metrics | see *Measurement* below |

The refutation grounds are `not-present`, `already-guarded`, `out-of-scope`,
`pre-existing`, `none`. `REFUTATION_GROUNDS` is **derived from the schema enum**
rather than re-typed — two copies of one list, where one instructs the model and
the other judges its answer, would drift silently and fail safe, so no test would
catch it.

## The one subtle safety property

`pre-existing` — "this defect is in code the PR did not touch" — is only
decidable from a **complete** changed-file list. Truncate the list and an absent
path reads as "untouched" when it was merely cut off. That is the single way this
list could fail **open** and drop a real finding.

So `changedFileContext` reports `authoritative: false` for a missing, malformed,
**or truncated** list, and the prompt then withdraws `pre-existing` as a ground
outright. Hitting the cap costs the availability of one *ground*; it can never
cost a finding.

**This adds a second reason the incremental-review PR must keep
`--changed-files` cumulative.** The first (a shrinking list could un-require, via
`lensApplies`, a lens that failed an earlier round) is already noted in the plan;
this is a different consumer with a different failure mode.

## Measurement

`refutedHighConfidence` used to *be* the drop count. It is now an upper bound on
it, so leaving it alone would have shipped a number in the PR summary whose label
had quietly stopped being true. `dropped` is reported alongside it:

```
- Refuted: 3 (2 high-confidence, 1 dropped)
```

The **gap** is the instrument for this PR — confident refutations the gate
declined to act on. Zero gap means the grounding requirement is costing nothing;
a large one means it is doing the work. Ledger entries written before this change
have no `dropped` field and coerce to 0, which reads correctly rather than merely
safely: under the old rule those drops *were* `refutedHighConfidence`.

## Expected regression: more rounds, not fewer

This is a precision/recall trade taken deliberately toward recall. Strictly more
verdicts now keep their finding, so more findings survive to the gate, so more
fix rounds — on the cost axis, in the wrong direction, on purpose.

The convergence detection from #570 is the counterweight: a PR that starts
re-raising the same findings pages a human at round 3 instead of burning to
`MAX_REVIEW_ROUNDS`. **If this PR causes a page with reason
`re-raising the same findings`, that is the pair of changes working, not a bug.**

## Linked Issues

## Author checklist

## Verification

Locally: `pnpm verify:self` green (11/11); `agent:tests` **107 tests** (was 105 —
new coverage for `isDroppingVerdict` and `changedFileContext`).

Beyond the suite, the rendered verifier prompt was **executed** against four
changed-file shapes — complete, exactly at the cap, one over, junk-only — and
inspected. The `pre-existing` bullet flips to the withdrawal wording at exactly
the cap+1 boundary, the listed block truncates to 200, and the header reports the
true total rather than the listed count. `maxTurns: 8` was confirmed to reach the
SDK options object.

**What none of that covers:** whether the verifier actually *finds* the code. It
now depends on Grep/Glob succeeding in a checkout, which no unit test exercises.
The live signals on the next few autonomous PRs are the `refutedHighConfidence`
vs `dropped` gap, round counts, and `Total-cost`.

## Risk Assessment

- **Data/security risk:** none. No permission or trust-model change — the
  verifier's tool allow-list, `settingSources: []`, and `permissionMode` are
  untouched, and the changed-file list is fenced as DATA like every other
  model-facing input. The gate itself gets *stricter*: every rule change here
  moves in the keep-the-finding direction.
- **Rollback plan:** revert. Pure script change, no state, no migration. Note
  that reverting restores the two-field drop rule, so `dropped` would go back to
  equalling `refutedHighConfidence`.

## Notes for Reviewers

  session and reviewed by me.
- **Where I'd push hardest as a reviewer:** `VERIFIER_MAX_TURNS = 8` is a guess.
  It has to cover a Grep, two or three Reads, and a verdict. Too low and the
  verifier runs out of budget and errors — which fails toward *keeping* the
  finding, so the failure is safe but silent and looks like a real confirmation.
  Worth revisiting against actual turn counts once this has run live.
- The verifier prompt is more prescriptive than the old one (an explicit
  ground→condition mapping). That is deliberate: each branch corresponds to an
  enum value the trusted script checks, so vague phrasing there produces verdicts
  `isDroppingVerdict` silently rejects.

## Summary by CodeRabbit

* **New Features**
  * Review findings are now removed only when refutations are high-confidence and supported by verified repository evidence.
  * Review summaries now report findings that were refuted separately from those actually removed.
  * Verifier checks now re-evaluate findings against the current repository state.

* **Documentation**
  * Added guidance on independent verification, evidence requirements, changed-file context, metrics, and prompt validation.

* **Bug Fixes**
  * Improved handling of incomplete or truncated changed-file lists.
  * Preserved compatibility with older review records missing the new removal metric.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -412,9 +412,24 @@ Components:
   separate UNPRIVILEGED `deps` job so no install runs with the secrets), and
   returns findings (schema-requested, then locally shape-validated + fail-safe
   severity-normalized) classified `critical`/`major`/`minor`/`nit`. A per-finding
-  **verifier** subagent then tries to refute each blocking finding and drops it
-  ONLY on a high-confidence explicit refute — any uncertainty keeps the finding
-  (fails toward blocking, so the false-positive lever can't swallow a real bug). The
+  **verifier** subagent then tries to refute each blocking finding. It is
+  deliberately **not given the diff**: the lens that raised the finding reasoned
+  from that diff, so a verifier reading the same diff inherits its misreadings
+  and confirms them — the correlated-error failure of naive review panels. It
+  instead re-establishes the facts from the repository itself (Read/Grep/Glob
+  against the branch checkout, capped at 8 turns), is told to distrust the
+  finding's quoted evidence, and receives only the cumulative changed-FILE list
+  so it can tell new code from pre-existing. Dropping is **grounded, not
+  asserted** (`isDroppingVerdict`): the verdict must be an explicit `refuted`, at
+  high confidence, naming one of `not-present | already-guarded | out-of-scope |
+  pre-existing`, AND citing at least one `file:line` it actually read. Anything
+  less — including a bare high-confidence refute, which used to be enough — keeps
+  the finding (fails toward blocking, so the false-positive lever can't swallow a
+  real bug). `pre-existing` is withdrawn whenever the changed-file list is
+  missing or was truncated, since an absent path would otherwise read as
+  "untouched" when it was merely cut off. The metrics comment reports
+  `refutedHighConfidence` and `dropped` separately; the gap between them is the
+  count of confident refutations the gate declined to act on. The
   **trusted orchestrator** (run from a `main` checkout, via the shared
   `scripts/agent/severity.mjs` rule) computes each lens's conclusion — the
   subagents only classify — and the job records one unforgeable
@@ -445,10 +460,11 @@ Components:
     2. **Cross-round re-check.** Each round persists its blocking findings in the
        per-lens check run's `output.text`; the next round reads the latest prior
        `agent-review-<lens>` findings (`--prior-findings`) and re-verifies each
-       against the *current* diff with the same biased-to-keep refute pass. A
-       prior finding survives unless it is confidently *resolved* — so it can't
-       vanish just because a later fresh pass missed it (only because it was
-       actually fixed). Unresolved priors merge (deduped) into the round's findings.
+       against the *current repository* with the same biased-to-keep refute pass.
+       A prior finding survives unless it is *resolved* on grounded evidence — so
+       it can't vanish just because a later fresh pass missed it (only because it
+       was actually fixed, cited). Unresolved priors merge (deduped) into the
+       round's findings.
 
     Both lower false-negative odds; neither makes the panel safe to self-promote —
     the human review gate stays the backstop.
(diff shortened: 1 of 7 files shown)


harrykim8672 (approved):
Thanks for your contribution.

PR #574, opened 2026-07-28
Invert the lens rubrics to coverage-first

hwisoo-kim:
## Summary

Invert the four lens rubrics from *"only report what you're sure about"* to
*"report everything, the verifier filters"*, and give findings a `confidence`
field so doubt has somewhere to go other than the severity scale.

Script + rubric only, no workflow edits. This is the largest expected recall win
in the series.

## Why

The audit's sharpest single number: **zero `critical` findings across nineteen
agent PRs.** Not one. The scale has a top rung nothing ever reaches.

The cause is in the rubrics. All four closed with a conservatism clamp —

> *"When unsure, downgrade."*
> *"Use critical/major ONLY with concrete, cited evidence."*
> *"When unsure whether something is concrete or taste, mark it minor."*

— while the schema gave a lens exactly **one** field to express doubt with:
`severity`. So doubt and impact were the same number, and every uncertain
`critical` came back as a `minor`.

That instruction was a reasonable local response to a schema gap. It is also a
documented failure mode: Anthropic's guidance for Opus 4.7+ says severity filters
in the prompt make the model **investigate just as thoroughly and then decline to
report** — precision rises, measured recall falls, and the work was done either
way.

The panel had two precision stages (this clamp, then the verifier) and no recall
stage at all.

## What changed

**Two axes instead of one.** `FINDING` gains a required `confidence` of
`high | medium | low`:

| field | means | gates? |
|---|---|---|
| `severity` | impact **if the finding is real** | **yes** — `BLOCKING = {critical, major}` |
| `confidence` | how sure the lens is | **no** |

Each rubric now has a *Coverage first* section, a *Severity — impact, not
certainty* section, and a *Confidence — certainty, separately* section. The
operative line: **never downgrade severity to express doubt.** A crash on a real
path is `critical` even at low confidence.

## The clamp that wasn't in the rubrics

Worth flagging, because the plan for this change named four files and this was
not one of them.

`runLens` appends its own closing instruction **after** the rubric and the diff:

> Return ONLY the structured verdict. Use critical/major severity ONLY for a
> concrete, defensible violation with cited evidence; taste → minor/nit.

It is the **last thing the lens reads**, so it wins ties against everything above
it. Rewriting only the four `.md` files would have shipped every lens a
coverage-first rubric followed immediately by its own contradiction — and the
measured result would have read as *"the rubric inversion didn't work."*

Both now say the same thing, with a comment at that call site explaining why they
have to stay in sync.

The one part worth keeping was **"taste → minor/nit"**. That is a judgement about
a finding's *kind*, not about certainty, so it survives in both places — now
stated explicitly as such, since that distinction is what the whole change turns
on.

## Two things deliberately NOT done

**Gating still reads `severity` only.** Gating on confidence is the obvious next
thought and it's wrong: it would rebuild the same clamp inside the trusted
script, one layer down, where it's harder to see. A `low`-confidence `critical`
blocks exactly like a `high`-confidence one. Filtering is the verifier's job, and
as of #573 the verifier has to ground its refutations.

**Confidence is not shown to the fix agent.** It already isn't —
`renderSummaryMd`'s `section()` prints only `file` and `summary` — so this needed
no code. But it's a property to keep on purpose rather than by accident: a fixer
that can see `confidence: low` on a blocking finding has been handed an argument
for skipping it.

## Measurement

New row in the PR summary comment:

```
- Confidence of raised (does NOT gate): 4 high, 2 medium, 1 low, 0 unrated
```

`unrated` is load-bearing, and it's the **opposite** convention to
`normalizeSeverity` (unknown → `major`). Severity gates the merge, so an
unparseable value must resolve to something safe. Confidence gates nothing, so
it's purely a measurement — and coercing a missing value into a real bucket would
hide precisely what this exists to detect: a lens that stopped emitting
confidence and went back to expressing doubt through severity.

**Two signals to watch after this lands:**

1. **`critical` becomes non-zero.** If it stays at zero, the decoupling didn't
   take and the cause is elsewhere.
2. **`medium`/`low` actually get used.** An all-`high` distribution means the
   lens isn't using the new axis and the clamp effectively survived.

## Expected regression: more findings, more rounds

Third PR in a row pushing recall up and round count with it — and the largest of
the three. Both counterweights are already merged: **#573**'s grounded
independent verifier is a genuinely better filter than a prompt asking the
reporter to self-censor, and **#570** pages a human at round 3 rather than
burning to `MAX_REVIEW_ROUNDS`.

**The ordering was load-bearing, not incidental.** Shipping this before #573
would have left a window with the clamp removed and no better filter in place.

## Linked Issues

## Author checklist

## Verification

Locally: `pnpm verify:self` green (11/11); `agent:tests` **112 tests** (was 110).

Two checks beyond the suite:

- **The assembled lens prompt was rendered end-to-end** and read top to bottom —
  rubric, diff, closing block — to confirm the rubric and the wrapper now agree.
  This is the check that found the `runLens` clamp; reading the diff would not
  have.
- **The new rubric guard is mutation-tested.** Re-adding `When unsure,
  downgrade.` to `correctness.md` turns it red; removing it turns it green again.
  A guard whose subject is prose is easy to write in a form that can never fail.
  It also asserts the rubric set it checks equals the manifest's lens ids, so a
  fifth lens can't ship with a clamp unnoticed.

**What none of this covers: whether the lenses actually behave differently.** No
test here exercises a model. The real evidence is the two counts above on the
next few autonomous PRs, plus the offline three-run agreement comparison.

## Risk Assessment

- **Data/security risk:** none. No permission, gating, or trust-model change. The
  *security* lens rubric is reworded but its lane is untouched, and it stays
  blocking and wildcard-scoped.
- **The real risk is noise, and it's on purpose.** More findings reach the fix
  loop; some will be wrong. That's the trade — the alternative is the status quo
  where a real `critical` is filed as a `minor` and nobody sees it. If round
  counts blow up, #570 pages rather than burning budget.
- **Rollback plan:** revert. Prompt text plus one optional-in-practice schema
  field; no state, no migration. Note that reverting the rubrics without
  reverting `runLens` (or vice versa) recreates the contradiction, so revert the
  commit whole.

## Notes for Reviewers

  session and reviewed by me.
- **Where I'd push hardest as a reviewer:** the `design-fit` rubric was the
  awkward one. Its `major`/`minor` line is genuinely about a finding's *kind*
  (unmet acceptance criterion vs. taste), not certainty, so "never downgrade to
  express doubt" needed a carve-out saying so explicitly. If that reads as
  muddled, it's the section to push on — it's the one place where the clean
  severity/confidence split doesn't map perfectly onto the lens's existing scale.
- `confidence` is `required` in the schema so a lens can't collapse the axes back
  together by omission. Anything that arrives unrated anyway is counted as
  `unrated` rather than assumed confident.

## Summary by CodeRabbit

* **New Features**
  * Review findings now report separate severity and confidence ratings.
  * Review summaries include confidence distribution data, including unrated findings.
  * Review guidance now emphasizes reporting all potential issues, including uncertain findings, with supporting evidence.

* **Bug Fixes**
  * Severity is no longer reduced solely because confidence is low; critical and major findings remain blocking based on severity.

* **Documentation**
  * Added guidance explaining the coverage-first review model and its reporting rules.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -411,7 +411,21 @@ Components:
   never loaded — the workflow also strips `.claude/` and installs the SDK in a
   separate UNPRIVILEGED `deps` job so no install runs with the secrets), and
   returns findings (schema-requested, then locally shape-validated + fail-safe
-  severity-normalized) classified `critical`/`major`/`minor`/`nit`. A per-finding
+  severity-normalized) classified `critical`/`major`/`minor`/`nit`, each with a
+  separate `confidence` of `high`/`medium`/`low`. The two axes are deliberately
+  independent: **`severity` is impact if the finding is real, `confidence` is how
+  sure the lens is**, and the rubrics tell lenses never to downgrade severity to
+  express doubt. Before this split the only channel for doubt WAS severity — the
+  rubrics closed with *"when unsure, downgrade"* — which is the documented
+  anti-pattern where a model investigates just as thoroughly and then declines to
+  report, and is why the panel recorded zero `critical` findings in its first
+  nineteen PRs. The rubrics are now **coverage-first**: report everything,
+  including uncertain findings, and let the verifier filter. **Gating is
+  unchanged and reads `severity` only** (`BLOCKING = {critical, major}`) — gating
+  on confidence would rebuild the same clamp inside the trusted script. The
+  confidence distribution is reported per PR so the axis can be seen to be in
+  use; it is not shown to the fix agent, which must address every blocking
+  finding regardless of how sure the lens was. A per-finding
   **verifier** subagent then tries to refute each blocking finding. It is
   deliberately **not given the diff**: the lens that raised the finding reasoned
   from that diff, so a verifier reading the same diff inherits its misreadings
--- a/docs/tasks/active/20260728-coverage-first-rubrics-lessons.md
+++ b/docs/tasks/active/20260728-coverage-first-rubrics-lessons.md
@@ -1,0 +1,69 @@
+# Lessons: coverage-first rubrics
+
+## A prompt has a last line, and the last line wins
+
+The plan for this change listed four files: the four lens rubrics. Rewriting
+those four would have produced a green PR, a good diff, and roughly no effect —
+because `runLens` appends its own closing instruction *after* the rubric and the
+diff, and that instruction said the opposite of the new rubrics.
+
+Nothing in the rubric files points at it. You find it by asking "what does the
+model actually receive, in order?" rather than "which files does this change
+name?" For prompt work those are different questions, and only the first one is
+about behaviour.
+
+Generalisation worth keeping: when a prompt is **assembled** from a template plus
+a data file, changing the data file is half the job. Print the assembled string
+and read it top to bottom, the way the model does.
+
+## One field cannot carry two meanings
+
+The rubrics said *"when unsure, downgrade"* because there was nowhere else for
+doubt to go — the schema had `severity` and nothing else. That instruction was a
+reasonable local response to a schema gap, and it quietly destroyed the severity
+scale: nineteen PRs, zero `critical` findings.
+
+The fix is not better wording. It is a second field. Once `confidence` exists,
+*"never downgrade severity to express doubt"* becomes actionable advice instead
+of a demand with no alternative.
+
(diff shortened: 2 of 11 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #576, opened 2026-07-28
Add a blast-radius lens for out-of-diff impact

hwisoo-kim:
## Summary

Add a fifth review lens whose entire job is **out-of-diff impact**, and give the
correctness and security lenses a call-site mandate for guards in their own lane.

## Why

Every false-negative defence the panel has re-reads the **same artifact**.
Sampling runs a lens twice on one diff. Cross-round re-check re-verifies findings
already raised. Run either a hundred times and a defect outside the diff stays
invisible, because neither changes *what is being read*.

One shipped on exactly that shape. #548 added a read-only guard to the docs
editor, and `EditorAPI.paste()` reached the same mutation without passing through
it. The correctness and security lenses each reviewed that diff **twice** and
passed it both times — not carelessly; the bypassing line was never in the diff.
CodeRabbit caught it by looking at the repository.

The diff tells you *what changed*. It never tells you *who depended on it*.

## The method is the lane

*"Consider wider impact"* is unfalsifiable advice — a reviewer satisfies it by
thinking hard and reporting nothing, which is the exact failure #574 was written
to remove. So the rubric leads with procedure:

1. Identify what the diff changes about the interface a **caller** sees.
2. `Grep` for every **other** reference to that symbol.
3. Decide whether each still holds.
4. Cite the broken site by `file:line`.

And it states the completion condition outright: **"If you finish without running
`Grep`, you have not done this review."** A procedure can be observed not to have
happened; an attitude cannot.

The lane is bounded just as hard in the other direction — line-level logic inside
the diff → correctness, auth design → security, tests and spec fit → their own
lenses. Without that, a fifth general reviewer just doubles cost and noise. The
rubric carries the test: *a finding you can state without leaving the diff belongs
to another lens.*

**`correctness.md` and `security.md` also gain the mandate** for guards in their
own lane. The overlap is deliberate: an added-but-bypassable gate reads as
covered, which is worse than no gate, and a crash path reachable around a new
check is a correctness bug on any reading.

## Adding one manifest row found the one list the manifest doesn't drive

Every consumer derives its lens set from `lenses.json` — the panel builds
`required_checks` from it, `set-state.mjs` prefix-matches `agent-review-`. One did
not: `DEFAULT_REVIEW_CHECKS`, the ready gate's fallback.

It **could not be tested where it lived** — `mark-ready.mjs` is a CLI with
top-level `process.exit` and cannot be imported at all. Untestable plus
hand-maintained is the combination that rots. It moves to `checks.mjs`, which
already exists for *"pure check-run gate logic, shared by mark-ready.mjs and its
tests"*, and `checks.test.mjs` now asserts it equals the blocking lenses in the
real manifest. A sixth lens can no longer be added without a red test.

**The gap was narrow, to be clear:** the promotion path always passes the
manifest-derived set, so a missing entry could never let a real promotion
through — it only made the local preview require a check the panel never posted.

## Two comments this change made wrong

Both are precisely what the new lens reports — a consumer of a changed contract
left describing the old one:

- `agent-review-panel.yml` said *"re-read by 4 lenses"*.
- `mark-ready.mjs` claimed the empty-required-set case was unreachable *"because
  every lens is `**`-scoped"* — **already false** since #564 gave design-fit and
  test-adequacy path globs. Corrected to name the invariant that actually holds
  (correctness and security are still blocking at `**`, asserted against the
  manifest in `review-panel.test.mjs`) rather than a count that will re-rot.

## Deliberately deferred: `.github/**` scope

Workflow changes have blast radius too — this workflow's outputs feed
`mark-ready.mjs` and `review-round-guard.mjs`, and a changed output shape would
break them silently. That is squarely this lens's class.

Left out anyway: it widens the lens past `test-adequacy`'s established convention
in the same PR that introduces it (two decisions in one), workflow PRs already
require a human push so a human is in the loop on exactly those changes, and it
costs a lens-round on every workflow PR when this PR already adds ~25% to panel
calls.

Asserted **explicitly false** in the test so the choice is visible and extending
it is a conscious one-line edit. Worth revisiting once this lens has produced
finding-rate data.

## Cost

Five lenses × 2 samples = **2 more SDK calls per round**, plus a verifier call per
blocking finding it raises. Roughly **+25% on panel cost.**

Third consecutive recall-increasing PR, so rounds rise too. #570's convergence
page and #573's grounded verifier remain the counterweights.

## Linked Issues

## Author checklist

## Verification

Locally: `pnpm verify:self` green (11/11); `agent:tests` **116 tests** (was 113);
`agent-review-panel.yml` re-parses as YAML.

Two things worth calling out:

- **#574's rubric guard caught this change before I wrote a line of test code.**
  It asserts the rubric set *equals the manifest* — so adding the lens failed the
  suite until `blast-radius.md` existed **and** was coverage-first. The fifth lens
  inherited that discipline mechanically rather than by my remembering to apply
  it. Written speculatively one PR ago; paid off on the next one.
- **The new drift guard is mutation-tested.** Removing
  `agent-review-blast-radius` from `checks.mjs` — simulating "added the lens,
  forgot the list" — turns `checks.test.mjs` red.

**What none of it covers: whether the lens finds anything.** No test runs a model.
The signal is whether it raises findings the other four miss on the next few
autonomous PRs — and specifically whether it would catch the `EditorAPI.paste()`
bypass, which is worth replaying offline against #548's diff.

## Risk Assessment

- **Data/security risk:** none. New lens uses the same read-only tool allow-list,
  `settingSources: []`, and `permissionMode` as the existing four. The
  `mark-ready.mjs` change only *adds* a required check — it can make the gate
  stricter, never looser.
- **The real risk is cost and noise**, quantified above. `gating: "blocking"` from
  day one is a deliberate call: an advisory lens for a bug class that already
  shipped past review would just be a comment nobody acts on. If it proves noisy,
  flipping it to `"advisory"` is a one-field change in `lenses.json`.
- **Rollback plan:** revert. Remove the manifest row, the rubric, and the
  `checks.mjs` entry — the test asserts those three stay consistent, so a partial
  revert fails loudly rather than silently.

## Notes for Reviewers

  session and reviewed by me.
- **Where I'd push hardest as a reviewer:** whether this should be blocking on day
  one, and whether the lane boundary against `correctness` is crisp enough in
  practice. On paper *"can you state it without leaving the diff?"* is a clean
  test; with two lenses both told to check call sites, I'd expect some duplicate
  findings in the first few runs. `dedupeFindings` collapses identical
  file+summary pairs, but a rewording slips through — so the first real signal to
  watch is whether blast-radius and correctness report the same bug twice.
- Moving `DEFAULT_REVIEW_CHECKS` between files is the one change here that isn't
  strictly required by the feature. I judged it in scope because this PR is the
  event that exposed the drift, and because leaving it would mean shipping a
  hand-maintained list I'd just proven goes stale. Happy to split it out if you'd
  rather.

## Summary by CodeRabbit

* **New Features**
  * Added a blocking “Blast-radius” review lens to identify affected call sites, bypasses, and out-of-diff regressions.
  * Expanded correctness and security reviews to check related call sites for unguarded access.

* **Bug Fixes**
  * Improved review readiness checks to stay synchronized with the configured blocking review lenses.
  * Clarified default and explicitly empty required-check behavior.

* **Documentation**
  * Added guidance and lessons for applying blast-radius reviews, evidence requirements, and coverage expectations.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -177,7 +177,8 @@ jobs:
           # `.interp` / `.tokens` are IDE/tooling files with no reference
           # anywhere in src, package.json, or the build — nothing loads them at
           # runtime, so no lens can act on their contents. On PR #521 they were
-          # 17.2 KB of a 75.6 KB diff, re-read by 4 lenses x 2 samples x 5 rounds.
+          # 17.2 KB of a 75.6 KB diff, re-read by every lens x 2 samples x 5
+          # rounds (4 lenses then, 5 now — the multiplier only grows).
           #
           # The generated `.ts` files are deliberately KEPT, even though they are
           # the larger half: they are executable code shipped in the sheets
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -394,7 +394,7 @@ Components:
   fork-originated `workflow_run` events are rejected),
   ONE orchestrator process (`scripts/agent/review-panel.mjs`, Claude Agent SDK)
   spawns a FRESH read-only subagent per **lens** — `correctness`, `security`,
-  `design-fit`, `test-adequacy` (declared data-drivenly in
+  `design-fit`, `test-adequacy`, `blast-radius` (declared data-drivenly in
   `scripts/agent/lenses/lenses.json` + one rubric `.md` each). The reviewed
   artifact is the branch diff against `main`, minus ANTLR generated *tooling*
   artifacts (`packages/sheets/antlr/*.interp|.tokens` — nothing loads them at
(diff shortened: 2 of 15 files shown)


harrykim8672:
@claude review

harrykim8672 (approved):
Thanks for your contribution

PR #578, opened 2026-07-28
Extract the agent SDK wrapper behind a required tool grant

hwisoo-kim:
## Summary

Moves `askStructured`, `classifyResult` and `withRetry` out of
`scripts/agent/review-panel.mjs` into a new `scripts/agent/ask.mjs`, and makes
`allowedTools` a **required, validated** parameter instead of a hardcoded array.

`assertAllowedTools` refuses `Bash`, `BashOutput`, `KillShell`, `Write`, `Edit`,
`NotebookEdit`, `WebFetch`, `WebSearch` and `Task` whatever the caller passes,
and returns a frozen copy. `review-panel.mjs` now grants exactly
`REVIEW_TOOLS = ["Read","Grep","Glob"]` at both call sites.

Also exports `CITATION` from `review-panel.mjs`.

## Why

A second consumer of this SDK plumbing is arriving — an autonomous
issue-hunting pipeline (planned as harness Phase 26). The tempting way to share
`askStructured` is to widen its hardcoded `allowedTools` for every caller, which
would quietly grant the review lenses capabilities they must never have: they
read an untrusted diff, so `Bash` would turn injected text into execution.

Making the grant a required parameter with a deny-list inverts that: widening a
grant becomes impossible rather than merely discouraged, and a consumer that
genuinely needs to execute something has to do it in its own trusted code from
data the model returned. The read-only invariant was previously a hardcoded
array with no test; it is now stated in one place and covered by eight.

`CITATION` is exported for the same reason `REFUTATION_GROUNDS` is derived from
its schema rather than hand-written: a copied regex that drifted would silently
change what counts as evidence in a gate.

This is deliberately a standalone refactor PR ahead of the new subsystem — a
pure change to a CODEOWNERS-protected, safety-critical file reviews far better
alone than bundled with a new feature.

## Linked Issues

Fixes #

No issue filed — this is a prerequisite refactor for the issue-hunting pipeline.

## Author checklist

## Verification

Skip reason (if applicable): n/a — awaiting CI. Locally `pnpm verify:self`
passed all 11 lanes; this change touches only `scripts/agent/*.mjs`, which no
package imports.

Behavior-preservation evidence beyond the lanes:

- **Export surface diff vs `origin/main` is exactly `+CITATION`** — the one
  intentional addition (compared by importing both revisions and diffing
  `Object.keys`).
- **`review-panel.test.mjs` is untouched** and all 27 tests still pass. It
  imports `classifyResult`/`withRetry` from `review-panel.mjs`, which is why
  they are re-exported there; leaving that file unmodified is the evidence.
- **Lazy SDK import preserved** — all 121 agent tests pass with the
  `@anthropic-ai/claude-agent-sdk` specifier hard-blocked by a resolver hook,
  confirming neither module resolves it at import time.
- **The SDK does not mutate `allowedTools`** (it destructures and zod-validates
  it), so returning a frozen array is safe.

## Risk Assessment

- **User-facing risk:** none. No package imports `scripts/agent/`; nothing ships.
- **Data/security risk:** net reduction. The read-only tool grant for reviewer
  subagents was an untested hardcoded array; it is now validated on every call
  with a deny-list that cannot be overridden by a caller. No change to
  `permissionMode: "dontAsk"` or `settingSources: []`.
- **Rollback plan:** revert the commit. The change is self-contained to three
  files in `scripts/agent/` and has no migrations, config, or state.

## Notes for Reviewers

- **AI assistance:** this PR was written with Claude Code (Opus 5) in an
  interactive session — design, implementation, tests and this description.
  Every line was reviewed by me before commit. This is *not* an autonomous
  pipeline run: `WAFFLEBASE_AGENT_AUTONOMOUS` was unset, so the branch is
  deliberately **not** named `agent/*` and the PR is not agent-managed (see
  below).

- **Branch naming is deliberate.** `agent-review-panel.yml`'s gate sets
  `managed = true` unconditionally for any branch starting with `agent/`. Using
  that prefix for a human-authored PR would hand branch ownership to the cloud
  fix loop mid-review, which is the single-writer hazard
  `docs/design/harness-engineering.md` warns about for the local front door.
  Hence `harness/`.

- **This is a fork PR**, so the agent pipeline's back half cannot run on it
  regardless of prefix — the review-panel gate requires
  `head_repository.full_name == github.repository`. `@claude review` still works
  (it is read-only and explicitly supports forks) if you want the lens panel's
  advisory opinion; `@claude loop` does not, as it needs a same-repo branch the
  fixer can push to.

- **Two small judgement calls worth a look:**
  1. An empty `allowedTools: []` is *rejected*, not treated as "no tools" — the
     SDK would run an agent that cannot read anything. `auth-smoke.mjs`
     legitimately wants `[]` for a pure auth probe, which is why it calls the SDK
     directly and does not come through this wrapper; noted in `ask.mjs`.
  2. `askStructured` takes a `label` (default `"agent"`) purely so the thrown
     error keeps reading `review query API error …` for the panel. It has no
     behavioral effect; it exists to keep existing log output identical.

- **Follow-up work:** the issue-hunting pipeline itself (harness Phase 26),
  which is what needs this seam. It will import `askStructured` with the same
  read-only grant and execute model-proposed `argv` arrays in its own trusted
  code rather than holding a shell.

## Summary by CodeRabbit

- **New Features**
  - Added a standardized structured agent request flow with enforced read-only tool permissions.
  - Introduced strict validation for allowed tool lists and improved result normalization (success vs no-output vs API/quota failures).
  - Implemented targeted retries with exponential backoff for eligible failures.
  - Centralized reviewer orchestration to reuse the same permissioning and retry/classification logic.

- **Tests**
  - Added unit tests covering tool allowlisting, request validation, result classification scenarios, and retry timing/behavior.

### DIFF
--- a/scripts/agent/ask.mjs
+++ b/scripts/agent/ask.mjs
@@ -100,6 +100,47 @@ export function assertAllowedTools(allowedTools) {
 }
 
 /**
+ * A session/usage limit resets on a fixed schedule, often hours out, so no
+ * in-run retry can clear it. This is the ONE thing that overrides an otherwise
+ * retryable status, so the pattern is deliberately narrow and anchored.
+ *
+ * It replaced `/session limit|usage limit|quota|rate limit|resets?\b/i`, which
+ * over-matched badly in both directions:
+ *   - `resets?\b` matched `ECONNRESET`, `read ECONNRESET` and `connection reset
+ *     by peer` (word boundary at end-of-string), turning ordinary transport
+ *     blips into "give up immediately".
+ *   - `rate limit` made every plain 429 non-retryable, contradicting the very
+ *     docblock above it — a rate limit is the canonical thing you retry.
+ *   - bare `quota` is ambiguous enough to catch unrelated messages.
+ * Dropping those is safe in the cheap direction: if a genuine limit is now
+ * retried, the cost is a few seconds of backoff before the same failure is
+ * reported correctly. Whereas mis-classifying a transient error as permanent
+ * pages a human for nothing, which is the expensive mistake.
+ */
+const SESSION_LIMIT_RE = /\b(?:session|usage)\s+limit\b/i;
+
+/**
+ * Is another attempt at this HTTP status plausibly worth making?
+ *
+ * Keyed on the status the API actually returned rather than on prose, because
+ * the message text is a human string that varies by provider and version while
+ * the status is a contract. Absent status means a transport/network failure
+ * (`fetch failed`, `ECONNRESET`) — those are the most retryable of all.
+ *
+ * 4xx other than 408/429 are deterministic: a 401 from a bad token or a 400 from
+ * a malformed schema will fail identically three times, so retrying only delays
+ * the real error. That is a change in behavior from the previous regex-only rule,
+ * and an intentional one.
+ */
+function isRetryableStatus(status) {
+  if (status == null) return true; // no status → transport/network → transient
+  const s = Number(status);
+  if (!Number.isFinite(s)) return true; // unparseable → treat as transient
+  if (s === 408 || s === 429) return true; // request timeout / rate limited
+  return s >= 500; // 500/502/503/529 overload → transient; other 4xx → not
+}
+
+/**
  * Classify an SDK `result` message. The SDK reports API/quota failures as
  * subtype "success" with `is_error: true` (+ `api_error_status`, and a human
  * `result` string like "You've hit your session limit · resets 3:30pm (UTC)"),
@@ -107,8 +148,11 @@ export function assertAllowedTools(allowedTools) {
  *   { ok:true, output }                                  — real structured verdict
  *   { ok:false, kind:'api-error', status, detail, retryable } — API/quota failure
  *   { ok:false, kind:'no-output', detail, retryable:false }   — ran but no verdict
- * A session/usage-limit resets on a fixed schedule (often hours out), so it is
- * NOT retryable in-run; any other API error (plain 429/529/overload/network) is.
+ *
+ * Retryability = "could another attempt succeed?", decided by `isRetryableStatus`
+ * and overridden to false only for an explicit session/usage limit. A plain 429
+ * IS retryable even when its message mentions rate limiting; a 429 that says
+ * "session limit" is not.
  */
 export function classifyResult(message) {
   const m = message || {};
@@ -117,8 +161,9 @@ export function classifyResult(message) {
   }
   if (m.is_error || m.api_error_status || m.terminal_reason === "api_error") {
     const detail = typeof m.result === "string" && m.result ? m.result : "";
-    const isQuota = /session limit|usage limit|quota|rate limit|resets?\b/i.test(detail);
-    return { ok: false, kind: "api-error", status: m.api_error_status ?? null, detail, retryable: !isQuota };
+    const status = m.api_error_status ?? null;
+    const retryable = !SESSION_LIMIT_RE.test(detail) && isRetryableStatus(status);
+    return { ok: false, kind: "api-error", status, detail, retryable };
   }
   return { ok: false, kind: "no-output", status: null, detail: `subtype=${m.subtype}`, retryable: false };
 }
--- a/scripts/agent/ask.mjs
+++ b/scripts/agent/ask.mjs
@@ -2,14 +2,19 @@
 // a model session, so the read-only tool invariant is stated once and can be
 // tested, rather than living as a hardcoded array inside a single caller.
 //
-// Extracted from review-panel.mjs unchanged except for one thing: `allowedTools`
-// is now a REQUIRED, VALIDATED parameter instead of a hardcoded
-// `["Read","Grep","Glob"]`. That matters because a second consumer is arriving
-// (the issue hunter), and the tempting way to share this function is to widen
-// the tool list for everyone. `assertAllowedTools` makes that impossible: the
-// deny-list below is refused no matter who asks, so a caller that needs to
-// execute something must do it in its own trusted code rather than by handing
-// the capability to a model.
+// Extracted from review-panel.mjs, which called the SDK inline. Two changes come
+// with the move, both about making the read-only claim real rather than assumed:
+//
+//   1. The tool grant is a REQUIRED, VALIDATED parameter checked against the
+//      `PERMITTED_TOOLS` allow-list below, instead of a hardcoded array. Sharing
+//      a wrapper invites widening it for everyone; an allow-list makes that a
+//      rejected call rather than a one-word edit.
+//   2. The session now sets `tools` as well as `allowedTools`. Only the former
+//      restricts what exists; the latter just pre-approves. See the call site.
+//
+// A consumer that genuinely needs to execute something must do it in its own
+// trusted code, acting on data the model returned — never by holding the
+// capability itself.
 //
 // SDK: @anthropic-ai/claude-agent-sdk (imported lazily so the pure helpers here
 // are unit-testable without the dependency installed). Verified against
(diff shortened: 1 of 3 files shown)


hwisoo-kim:
@claude review

hwisoo-kim:
## Review disposition

Verified every finding against the pinned SDK (0.3.217) rather than the docs. Two were right in a way that mattered more than the write-up suggested, several are real but pre-existing, and a few don't hold.

### Fixed in e06076457

**The headline finding is correct, and it was worse than described.** `allowedTools` is an auto-*approve* list, not a restriction — `sdk.d.ts:1341-1348` says outright *"To restrict which tools are available, use the `tools` option instead."* So the exact-string deny-list didn't merely fail to match `Bash(git diff:*)`: because entries are permission **rules** on an approve-list, that string would have **pre-approved shell execution past `permissionMode: 'dontAsk'`**. Same for `" Bash"` (trimmed for the emptiness check, compared untrimmed) and `mcp__server__exec`.

The deny-list was also just wrong for this SDK. Checked against `sdk-tools.d.ts`: it named `Task`, `BashOutput` and `KillShell`, none of which exist in 0.3.217's tool schema, while omitting `Agent` — which spawns subagents that inherit the parent's tools, i.e. precisely the respawn escape the doc comment claimed was impossible — plus `REPL`, `Workflow`, `CronCreate`, `RemoteTrigger`, `Artifact`, `Mcp`.

Both fixes land together: `FORBIDDEN_TOOLS` → a `PERMITTED_TOOLS` **allow-list**, and the session now sets `tools` (the real restriction) alongside `allowedTools`. Scoped rules, MCP names, whitespace variants and every future tool are now refused because they're absent, not because someone predicted them. Non-canonical spellings are rejected rather than normalized — a validator that repairs its input is guessing, and a capability gate is the wrong place to guess.

Also fixed: `label` now attributes the no-output error; `REVIEW_TOOLS` is exported so a test can assert the panel's real grant; the `hunt-probe.mjs` references are gone; and the `CITATION` export + test are **removed** — design-fit is right that it's unused in-tree, and it belongs in the PR that adds the consumer.

**Tests.** The self-referential point was well made — the deny-list test looped over the list it validated, so a removal deleted behavior and assertion together. Now pinned to literals, plus coverage for each bypass class (scoped-rule, MCP, whitespace), the real SDK tool names a deny-list would have missed, and `askStructured` itself: deleting its validation call previously left the suite green. I deliberately did **not** add a "valid grant reaches the SDK" test — it opens a real model session, which on a credentialed runner means burning a paid session from a unit suite.

### Valid, but pre-existing — not this PR

These are all in `classifyResult`, moved **verbatim** from `review-panel.mjs`. They're real, and I'd like to fix them, but not in a PR whose contract is "behavior-preserving extraction" — changing the panel's retry and gating semantics deserves its own reviewable diff:

- `resets?\b` matches `ECONNRESET` and `"connection reset by peer"`, flipping transient network failures to `retryable: false`. **Confirmed real** — the pattern matches at a word boundary at end-of-string.
- `"rate limit"` in the non-retryable set contradicts the docblock directly above it.
- `structured_output` is checked before `is_error`, so an errored session carrying output is returned as a verdict — fail-open in a fail-closed gate.
- `SDKResultError` subtypes (`error_max_turns`, `error_max_budget_usd`) carry `is_error` with no `result`/`api_error_status`, so a deterministic failure is retried 3× and reported as "unknown".
- No timeout/abort; `schema` unvalidated; `maxTurns` silently dropped when non-numeric.

Happy to take these as a follow-up — say the word and I'll open it.

### Doesn't hold

- **"No CI workflow runs `scripts/agent/*.test.mjs`."** It does. `.github/workflows/ci.yml:41` runs `pnpm verify:self`, whose first lane is `agent:tests` (`scripts/verify-self.mjs:16`: `cd scripts/agent && node --test *.test.mjs`). These tests gate every PR.
- **"Frozen array may crash if the SDK mutates `allowedTools`."** The correctness lens's own verifier refuted this, and I confirmed it independently — the SDK destructures and zod-validates, never mutates. Moot regardless: the validator no longer freezes its return.
- **"Bash/Write/WebFetch stay *available* to the subagent."** Overstated as written. `'dontAsk'` is documented as *"deny if not pre-approved"* (`sdk.d.ts:2063`), so bare unlisted names were denied at execution. They were *visible in context*, which is a real injection-surface concern and is what `tools` now fixes — but the exploitable path was the scoped-rule bypass, not the bare names.
- Three findings were double-reported (the `hunt-probe.mjs` nit, the exact-string critical, the deny-list-shape major), which inflates the counts.

`pnpm verify:self` green, 126 agent tests pass — including with the SDK specifier hard-blocked, confirming the lazy import survives.

hwisoo-kim:
## CodeRabbit: `classifyResult` retryability — validated and fixed in `0bbf9209d`

**Valid.** Confirmed empirically rather than by reading, because `resets?\b` is subtle — it matches at the end-of-string word boundary:

| `result` detail | old verdict |
|---|---|
| `ECONNRESET` | **NOT retryable** ❌ |
| `read ECONNRESET` | **NOT retryable** ❌ |
| `connection reset by peer` | **NOT retryable** ❌ |
| `rate limit exceeded, please retry` | **NOT retryable** ❌ |
| `You've hit your session limit · resets 3:30pm (UTC)` | NOT retryable ✅ (correct) |
| `overloaded_error` / `fetch failed` | retryable ✅ (correct) |

So an ordinary transport blip made `withRetry` give up on the first attempt and the panel report a false `PANEL_INFRA_ERROR`. And `rate limit` contradicted the docblock immediately above it, which named plain 429 as retryable — a rate limit is the canonical retry-with-backoff case.

### What changed

Retryability now comes from `api_error_status`, which is a contract, rather than from a human message string that varies by provider and version:

- **no status** (transport: `fetch failed`, `ECONNRESET`) → retryable
- **408 / 429 / 5xx** → retryable
- **other 4xx** (400/401/403/404/422) → **not** retryable. New behavior, intentional: a 401 from a bad token fails identically three times, so retrying only delays the real error.

One text override survives, narrowed and anchored to `/\b(?:session|usage)\s+limit\b/i`. Status alone genuinely cannot decide this case — a plain rate limit and a session limit **both arrive as 429** — but a session limit resets hours out, so no in-run backoff clears it.

Dropped bare `quota` and the reset-word match. That's safe in the cheap direction: a genuine limit now costs a few seconds of backoff before failing correctly, whereas mis-reading a transient error as permanent pages a human for nothing.

### Evidence

The four pre-existing `classifyResult` tests in `review-panel.test.mjs` pass **unchanged**, including the `#548` session-limit 429 case — that's the evidence this narrows the match without losing the semantics it existed for. Five new regression tests cover the ECONNRESET family, rate-limit-mentioning 429s, the session-limit override, the 4xx/5xx split, and a stringified status. 130 agent tests green, `verify:self` green.

### Skipped, with reasons

Still-open findings from the earlier panel review that I am deliberately **not** folding in here — all pre-existing in code this PR moved verbatim, and each changes panel gating or control flow enough to deserve its own reviewable diff:

- `structured_output` is checked before `is_error`, so an errored session carrying output is returned as a verdict — fail-open in a fail-closed gate. **This is the one I'd prioritize.**
- `SDKResultError` subtypes (`error_max_turns`, `error_max_budget_usd`) carry `is_error` with no status, so they land in the retryable bucket and a deterministic failure is retried 3× and reported as "unknown".
- No timeout/abort on a stalled session; `schema` unvalidated; `maxTurns` silently dropped when non-numeric.

Happy to take those as a follow-up.

**Note for reviewers:** this branch was rebased onto `main` at #579 by a parallel session while I was working. I replayed only the new commit onto that tip (`rebase --onto`) rather than force-pushing over it — `scripts/agent/` content was byte-identical across the rebase, so nothing was lost.

harrykim8672 (approved):
Thanks for your contribution

PR #581, opened 2026-07-29
Add the incremental-review decision logic, inert by default

hwisoo-kim:
## Summary

Land the decision logic for reviewing only the **delta** since a lens's last
verdict, plus the module that replaces the inline prior-findings script.

> **This PR changes nothing.** Every default preserves today's full-diff review.
> Workflow wiring is a follow-up — see *Why this is split* below.

## Why

The reviewed artifact is `git diff origin/main...HEAD`, recomputed every round, so
an 8-round PR reads round-1 code eight times (≈64 lens-reads on #548). Both
endpoints of `...` also move: a human `git merge main` changed the artifact with no
semantic change to the branch and **flipped a lens verdict**.

## Why this is split at the script boundary

Not size management. Two reasons, and the second is not a choice — the wiring edits
`.github/workflows/**`, which the agent App cannot push (no `workflows` scope: this
is what blocked #564, and it is deliberate, since `ci.yml` runs on `pull_request`
and a branch that could rewrite workflows could pass CI on its own PR). So the
wiring needs a human push however it is packaged.

The first reason is that the two halves differ in whether they can be *verified*:

- **This half is pure.** `resolveReviewMode` takes every git fact as an argument,
  so its entire decision table is a unit test.
- **The wiring half cannot be tested locally at all.** It depends on `external_id`
  surviving a round trip through the checks API, on `git merge-base --is-ancestor`
  against real branch history, and on a `workflow_run` context nothing local
  reproduces.

Landing the logic first means the untestable part arrives as a small, reviewable
diff instead of buried in a thousand-line one. Given a mistake here means
**unreviewed code gets promoted**, that seemed worth the extra round-trip.

**Follow-up PR:** stamp `external_id`, compute the git facts, `-U15` on narrowed
rounds, replace the inline prior-findings step, add `checks: read` to on-demand.

## Where the state lives, and why not `output.text`

Each `agent-review-<lens>` check run's **`external_id`** carries
`{"v":1,"reviewed":…,"base":…,"since":…,"mode":…}`.

Every candidate location sits behind the same `checks:write` boundary — the author
agent lacks it and cannot forge any of them — so the choice is purely about
**failure modes**. `output.text` is disqualified: this workflow trims it to fit a
60k limit **by dropping findings**. It is *designed* to lose data, and a control
field that decides whether code gets reviewed must not live in a lossy channel.

## The one rule

**Every path fails to `full`.** Reviewing code twice costs tokens; reviewing it
zero times ships a bug. There is no symmetry, so there is no case where "probably
fine" resolves to `incremental`.

Ten distinct reasons, with **correctness hazards checked before policy caps** so a
force-pushed branch reports `force-push-or-rewrite` rather than the
`delta-too-large` that also happened to fire. Three are additions to what the plan
sketched, because a fused reason is a worse diagnostic:

| reason | why it forces full |
|---|---|
| `lens-state-divergence` | lenses disagree on what they last reviewed — one missed a round, so the newest pointer would skip commits the laggard never saw |
| `no-new-commits` | re-run on an already-reviewed SHA: the delta is empty, and `review-panel.mjs` fails closed on an empty diff, so narrowing would turn a harmless re-run into an outage |
| `invalid-input` | a caller passing junk gets `full`, never a throw |

## The recall regression is real, and mitigated rather than hidden

Carry-forward re-checks findings already **raised**. It cannot surface a defect
whose root cause is round-1 code that only became *reachable* through round-3
code. Narrowing the diff makes that defect invisible.

Three mandatory mitigations. For `renderScopeNote`, three clauses are
load-bearing:

1. earlier commits were already reviewed (the saving), **but the lens still owns
   defects this delta newly exposes in them**;
2. the **complete** working tree is its cwd, so it should read outside the delta;
3. the changed-file list is **cumulative**, not the delta's files.

Clause 1 alone reads as *"old code is someone else's problem"* — which is exactly
the regression. Plus a forced full round every 3 rounds, and wider diff context on
narrowed rounds (wiring PR).

**Honest accounting: ~2× saving, not ~8×.** A third of rounds stay full by design,
and any rebase or merge resets to full.

## Two fail-opens deliberately not created

- **`--changed-files` stays cumulative in every mode.** Fed the delta,
  `lensApplies` could mark a narrow-glob lens inapplicable in round N, the workflow
  drops inapplicable lenses from `required_checks`, and a lens that **failed in
  round 2** would silently stop being required in round 3 — promoting with an
  unresolved blocker. Unreachable today; the constraint exists so it stays that
  way.
- **Incremental with no usable `--since-sha` is refused.** Otherwise a lens gets a
  partial diff with no scope note — reviewing a fragment while believing it is the
  whole PR. It throws instead.

## `prior-findings.mjs`

Replaces ~40 lines of inline `github-script` (and the second copy #558 would have
needed for on-demand) with the same behaviour, under test. Inline YAML JavaScript
is untestable and un-lintable.

Two GitHub API contracts are load-bearing here, both easy to lose in a rewrite and
both already learned once in this repo — **the first draft of this PR lost both,
and the review caught it** (see *Review round 1* below):

- the `commits/{sha}/check-runs` **list response omits or truncates `output.text`**,
  so each selected run is re-fetched by id;
- that endpoint is object-wrapped, so `--paginate` needs `--slurp` or the
  concatenated pages are invalid JSON.

`collectPrior` takes its API function as an argument. A module extracted *because*
inline YAML cannot be tested should not keep its only API path behind a real `gh`
binary; all four failure paths are covered, and each guard is mutation-tested.

Its fail direction is deliberately the *opposite* of most of this pipeline: prior
findings can only re-raise a blocker, never clear one, so a read error degrades to
"carry fewer findings" rather than failing the round, and every `catch` is
per-commit or per-lens. A throw here would turn one lens's bad JSON into a
panel-wide outage. The two exceptions are bad *arguments* — an unusable PR number
or unreadable lens manifest exits 2, because a broken invocation must not look like
a legitimately empty first round.

## Deviation from the plan

No `countCompletedRounds`. `rounds.mjs` already has `groupReviewRounds`, so
`roundIndex` is `groupReviewRounds(...).length` at the wiring layer, injected as a
plain number. A second round-counter is exactly the hand-maintained duplicate this
series keeps having to fix — see `DEFAULT_REVIEW_CHECKS` in #576.

## Linked Issues

## Author checklist

## Verification

Locally: `pnpm verify:self` green (11/11); `agent:tests` **161 tests** (148 on
`main` after #578).

**Inertness is proven by rendering the prompt, not by reading the source.** I
extracted `runLens`'s assembly from `origin/main`, rendered it beside this branch's
`buildLensPrompt` on identical inputs, and diffed: **byte-identical, 1012 bytes
each**. With a scope note they differ, and the note lands *before* the diff. The
test now pins that exact string, so any later edit to the assembly fails it — the
first version of this test grepped the source for two expressions and could not
observe the property at all.

**A bug the tests caught during development**, worth knowing because it recurred:
`= {}` as a parameter default fires only for `undefined`, so
`resolveReviewMode(null)` threw — contradicting its own documented no-throw
contract. `renderScopeNote` had the **identical** bug, and its test stayed green
because it only passed `undefined`. Both fixed; both tests now sweep the whole junk
class.

**What none of this covers: the wiring.** No local test exercises `external_id`
round-tripping, `--is-ancestor`, or the real round counter. Until the follow-up
lands, this code is inert — which is the intended state for merging it.

## Review round 1 — what was wrong

The review panel found real defects. Fixed, with the mechanism in each case:

| finding | why it mattered |
|---|---|
| `output.text` read from the check-runs **list** response | GitHub omits/truncates it there. Carry-forward would have silently become **zero findings** — indistinguishable from a clean round, and it disables the #521 re-check. Both the inline version *and* `review-round-guard.mjs` re-fetch per run for exactly this reason. |
| `--paginate` without `--slurp` on the object-wrapped endpoint | Invalid JSON past one page of check runs. This repo verified and documented it once already. |
| one outer `try` around the whole commit loop | A single failed call zeroed all five lenses — the opposite of the per-lens granularity the docstring claimed. Now per-commit and per-lens. |
| `renderScopeNote` ended in `.filter((l) => l !== "")` | Deleted every deliberate blank line: heading glued to the paragraph, closing DATA line swallowed as a lazy list continuation. **Same filter that ate the verifier prompt earlier in this series**, and substring assertions cannot see it. |
| the inertness test was a regex over source | Could not observe the byte-identical-prompt property it existed to guard. |
| a **false claim** I made in three places | I wrote that the inline version wrapped the loop in one `try`. It does not — it has a `try` per lens. I invented a defect to justify an extraction that was already justified. Corrected in the docstring, the commit message and above. |

Plus: tie-break on check-run `id` instead of response order; pre-parsed states now
validated through the same path as strings; `states` accepted keyed by lens id
*or* check name (the natural composition with `latestLensRuns` would otherwise
report `no-prior-state` forever and never narrow); `Object.hasOwn` so a lens id of
`constructor` can't pick up the prototype; a null-prototype `parseArgs`; and a new
fail-closed throw for `--since-sha` without `--review-mode incremental`.

Pinning the `external_id` width instead of asserting `< 255` — which was
constant-true — showed the docs said ~170 where it is **183**.

**Declined, with reasons:** refactoring `rounds.mjs::groupReviewRounds` onto
`latestLensRuns` (it answers a different question — latest per lens *within* a
commit — and feeds the round-guard page path; pulling a gate file into an inert
change trades a real risk for a cosmetic one), and checking the diff artifact
itself for narrowing (this script receives a file and cannot tell a narrowed diff
from a full one, so the flag pair is the only signal there is).

## Risk Assessment

- **Data/security risk:** none. No permission or trust-model change. `external_id`
  sits behind the same `checks:write` boundary as the verdicts themselves, and
  `latestLensRuns` filters on `app.slug === "github-actions"` so no *other App* can
  speak for a lens. To be precise about what that filter is and is not: the slug
  identifies the App, not the workflow, so any workflow **in this repo** that
  declared `checks: write` would share it. Today only the panel does, and no
  author-controlled workflow can gain it, because a branch cannot edit
  `.github/workflows/**`. That push restriction is the boundary; the filter narrows
  to it rather than replacing it.
- **Behaviour risk while unwired: none by construction**, and that is the claim the
  inertness test defends.
- **Rollback plan:** revert. Two new files plus three unused CLI flags; nothing
  reads the state until the wiring PR.

## Notes for Reviewers

  session and reviewed by me.
- **Where I'd push hardest as a reviewer:** the `fullEvery = 3` and
  `maxDeltaLines = 400` defaults are judgement calls with no data behind them yet.
  `fullEvery` directly sets how long a delta-exposed defect can hide, and it is
  also the single biggest lever on the saving — 3 is the conservative end. Both are
  parameters rather than constants so the wiring PR can tune them once there are
  real round counts to look at.
- **The `no-new-commits` case** is the one I am least sure belongs as `full` rather
  than a no-op. Re-running the panel on an already-reviewed SHA is arguably nothing
  to do at all, but "review it again" is the safe reading and matches what happens
  today.

## Summary by CodeRabbit

* **New Features**
  * Added incremental review support that can narrow each round’s review scope while retaining safeguards for full reviews when needed.
  * Added carry-forward handling for prior blocking findings across review rounds.
  * Added review-state tracking and scope notes to clarify which changes are being reviewed.

* **Documentation**
  * Added design guidance, implementation notes, lessons learned, and task tracking for incremental review.

* **Tests**
  * Added comprehensive coverage for review-state handling, scope selection, prompt formatting, and prior-findings recovery.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -513,6 +513,48 @@ Components:
     All three lower false-negative odds; none makes the panel safe to
     self-promote — the human review gate stays the backstop.
 
+  - **Incremental review (scope narrowing, `scripts/agent/review-state.mjs`).**
+    The reviewed artifact is `git diff origin/main...HEAD`, recomputed every
+    round, so an 8-round PR reads round-1 code eight times. Both endpoints of
+    `...` also move: a human `git merge main` changed the artifact with no
+    semantic change to the branch and flipped a lens verdict.
+
+    Each lens's **`external_id`** carries what it last reviewed —
+    `{"v":1,"reviewed":…,"base":…,"since":…,"mode":…}`. Every candidate location
+    sits behind the same `checks:write` boundary (the author agent cannot forge
+    any of them), so the choice was about failure modes: `output.text` is
+    disqualified because the panel trims it to fit a 60k limit by *dropping
+    findings* — it is designed to lose data, and a field that decides whether code
+    gets reviewed must not live in a lossy channel.
+
+    `resolveReviewMode` is fully pure (every git fact injected) and **fails to
+    `full` on any doubt**, reporting which: `no-prior-state`, `lens-state-gap`,
+    `lens-state-divergence`, `no-new-commits`, `git-facts-unavailable`,
+    `force-push-or-rewrite`, `merge-in-range`, `periodic-rebaseline`,
+    `delta-too-large`, `invalid-input`, or `ok`. Correctness hazards are checked
+    before policy caps so the reported reason names a real problem when one
+    exists. Reviewing code twice costs tokens; reviewing it zero times ships a
+    bug — there is no case where "probably fine" resolves to `incremental`.
+
+    **The recall regression is real, not hypothetical.** Carry-forward re-checks
+    findings already *raised*; it cannot surface a defect whose root cause is
+    round-1 code that only became reachable via round-3 code. Three mitigations,
+    all mandatory: a **scope note** (`renderScopeNote`) telling the lens it still
+    owns defects the delta newly exposes in earlier code, that the full working
+    tree is its cwd, and that the changed-file list is cumulative; a **forced full
+    round** every 3 rounds plus on any of the hazards above; and wider diff
+    context on narrowed rounds. The honest saving is therefore **~2x, not ~8x**.
+
+    `--changed-files` stays **cumulative** in every mode. Fed the delta instead,
+    `lensApplies` could mark a narrow-glob lens inapplicable in round N, the
+    workflow would drop it from `required_checks`, and a lens that FAILED in round
+    2 would silently stop being required in round 3 — promoting with an unresolved
+    blocker. Unreachable today; the constraint exists so it stays that way.
+
+    CLI defaults to `full`, so all three callers behave byte-identically until the
+    workflow opts in. The on-demand path must never narrow: it records no check
+    runs, so it can never write a pointer back.
+
     **API/quota error classification.** The Agent SDK reports API failures as a
     `result` message with `subtype:"success"` but `is_error:true` (+ `api_error_status`,
     e.g. a 429 "You've hit your session limit"), so "success" alone is not proof the
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -514,18 +514,35 @@ Components:
     self-promote — the human review gate stays the backstop.
 
   - **Incremental review (scope narrowing, `scripts/agent/review-state.mjs`).**
+    ⚠️ **Decision logic only — NOT yet wired, and therefore not yet saving
+    anything.** Nothing stamps `external_id`, nothing calls `resolveReviewMode`,
+    and the panel still reviews the full cumulative diff every round. The workflow
+    half (stamping the pointer, computing the git facts, `-U15` on narrowed rounds,
+    and swapping in `scripts/agent/prior-findings.mjs`) is a separate change; it touches
+    `.github/workflows/**`, which the agent App cannot push, so it needs a human
+    push either way. Read the rest of this entry as the design the scripts
+    implement, not as current behaviour.
+
     The reviewed artifact is `git diff origin/main...HEAD`, recomputed every
     round, so an 8-round PR reads round-1 code eight times. Both endpoints of
     `...` also move: a human `git merge main` changed the artifact with no
     semantic change to the branch and flipped a lens verdict.
 
-    Each lens's **`external_id`** carries what it last reviewed —
+    Each lens's **`external_id`** is to carry what it last reviewed —
     `{"v":1,"reviewed":…,"base":…,"since":…,"mode":…}`. Every candidate location
     sits behind the same `checks:write` boundary (the author agent cannot forge
     any of them), so the choice was about failure modes: `output.text` is
     disqualified because the panel trims it to fit a 60k limit by *dropping
     findings* — it is designed to lose data, and a field that decides whether code
     gets reviewed must not live in a lossy channel.
+
+    `latestLensRuns` filters check runs on `app.slug === "github-actions"`. That
+    narrows the writer to a workflow **in this repository**, not to the panel
+    specifically — the slug identifies the App, and any workflow here that
+    declared `checks: write` would share it. Today only the panel does, and no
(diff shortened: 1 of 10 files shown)


harrykim8672:
@claude review

harrykim8672 (approved):
Thanks for your contribution

PR #582, opened 2026-07-29
Route each review lens to the file classes it can act on

dlgpdmsly2:
## Summary

Adds a second, independent axis to the review panel. `appliesWhen` decides whether
a lens **runs**; a new `scopeClasses` decides what it **reads**.

Every changed file is classified — `code`, `code-adjacent`, `policy`,
`design-spec`, `prose` — and each lens receives only the diff hunks for the
classes it declares. A new cheap `docs` lens (sonnet-5, one sample) reviews prose.

| class \ lens | correctness | security | design-fit | test-adequacy | blast-radius | docs |
|---|:-:|:-:|:-:|:-:|:-:|:-:|
| code | ✓ | ✓ | ✓ | ✓ | ✓ | |
| code-adjacent | ✓ | ✓ | ✓ | ✓ | ✓ | |
| policy | ✓ | ✓ | ✓ | ✓ | ✓ | |
| design-spec | | ✓ | ✓ | | | |
| prose | | ✓ | | | | ✓ |

`security` deliberately reads every class. Every other lens has a subject-matter
lane, so narrowing what it reads only costs it findings in its own lane.
`security`'s lane is "anything hostile in this diff", which is not a property of
any one file class — so routing must not relocate it. Asserted in a test.

`policy` — `.github/**`, `.claude/**`, `CLAUDE.md`, `AGENTS.md`,
`CONTRIBUTING.md`, `harness.config.json`, `scripts/agent/lenses/*.md` — also goes
to every lens that can act on it. These define the pipeline's own privileges, and
they are tiny next to the prose this change targets, so there is nothing to gain
by routing them. Narrowing them would have left a workflow-only PR gated by
`security` alone.

`appliesWhen` and the deliberately-unfiltered `--changed-files` list are
untouched, so `required_checks` is computed exactly as before.

## Why

Every lens received the entire diff. The pipeline's own agents attach a
`docs/tasks/active/<slug>-todo.md` + `-lessons.md` pair to nearly every PR, so
that narration was re-read by five opus lenses at two samples each while
contributing nothing to a correctness, security, test, or blast-radius verdict.
On #546 it was 93% of the diff.

**This is routing, not skipping.** Design specs stay with design-fit;
agent-governing files (`CLAUDE.md`, `AGENTS.md`, `CONTRIBUTING.md`,
`scripts/agent/lenses/*.md`, `.github/**`) stay with security and design-fit as
`policy`; markdown that tests parse is read as code. All prose stays with
security, which is opus and always applicable, so injection and secret coverage
of narration is unchanged from today — the `docs` lens is additive there, not a
replacement.

The fail-safe direction is deliberate: `prose` is the only class routed off the
code lenses, so it requires an explicit glob match, and anything unrecognized
falls through to `code` and is reviewed by everyone. A new kind of file cannot
silently take the cheap path.

Measured on three real agent PRs — diff lines fed to lenses, summed across the
panel:

| PR | shape | before → after | reduction |
|---|---|---|---|
| #546 | 69 prose / 23 code | 552 → 253 | 54% |
| #549 | 107 prose / 36 spec / 301 code | 2664 → 1791 | 33% |
| #548 | 160 prose / 31 spec / 791 code | 5892 → 4337 | 26% |

That counts diff lines, not tokens or dollars — rubrics, system prompts, repo
reads and the verifier all sit outside the measure. The saving concentrates on
prose-heavy PRs, which is the shape it targets.

## Linked Issues

None — this is a cost-reduction change to the review pipeline, not an issue fix.

## Author checklist

## Verification

Locally, the `agent:tests` lane passes: `cd scripts/agent && node --test
*.test.mjs` → 143/143, including seven new tests. The slicer was also run over
the three real PR diffs above, asserting byte-identical round-tripping.

## Risk Assessment

- **User-facing risk:** none — this is pipeline-internal and ships no product code.
- **Data/security risk:** the review gate itself is the surface, so the safety
  argument is asserted in tests rather than in prose:
  - **Coverage invariant** — every file class must be in the `scopeClasses` of at
    least one *blocking* lens, and that lens must also *apply* to a diff made only
    of that class. A lens that reads a class but never runs on it is not coverage.
  - **`required_checks` is unchanged.** Slicing is a pure transform of the diff
    body. `lensApplies` still reads the same unfiltered changed-file list, so a
    lens that failed an earlier round cannot stop being required.
  - **Empty slice ⇒ `skipped` AND `applicable: false`.** A lens can apply and have
    nothing in scope (correctness on a docs-only PR). The workflow builds
    `required_checks` from `blocking && applicable` and blocks on any conclusion
    other than `success`, mapping `skipped → neutral` — so reporting
    `applicable: true` here would create a required check that can never go green,
    deadlocking every docs-only PR. Guarded by a test.
  - **The global empty-diff guard still fails closed.** Only a per-lens slice may
    be empty.
  - **Scope is decided from the cumulative changed-file list, never from the
    round's diff** — this rebases onto #581, where under `--review-mode
    incremental` the diff is only the delta since the last reviewed commit.
    Deciding scope from the diff would make a lens's applicability oscillate
    between rounds, and an inapplicable lens leaves `required_checks`: a round
    that only fixed a typo in a task file would have retired a correctness
    finding raised in an earlier round without anyone resolving it. That is the
    failure `--changed-files` is kept cumulative to prevent, and routing must not
    reintroduce it one axis over. So a lens with nothing new *this round* keeps
    gating and gets an empty diff — detection is skipped, the prior-round
    re-check still runs and is what resolves the earlier finding. Full-mode
    gating is unchanged for every PR shape.
  - **Slicer robustness.** Path resolution reads only each block's header region:
    a `.patch`/`.diff` fixture's content lines legitimately begin with `+++ b/`
    once the addition's leading `+` is counted, and scanning the whole block would
    let a fixture's payload rename it and route a real file to the wrong lens.
    Quoted paths (`"a/my file.md"`) are refused rather than guessed → `code`.
    Hunk bytes are preserved exactly, so the `file:line` a finding cites stays
    valid. A lens with no `scopeClasses` receives the whole diff, so an
    un-migrated manifest entry fails toward more review.
  - The verifier is unaffected — it no longer receives the diff at all.
- **Rollback plan:** revert the commit. Or, without a revert, deleting
  `scopeClasses` from a lens restores its full-diff behaviour, and dropping the
  `docs` entry from the manifest removes its check run (both are manifest-only
  edits).

## Notes for Reviewers

- **AI assistance:** this PR was written with Claude Code — the classifier,
  slicer, `docs` rubric, and tests.
- **No workflow change.** Adding a lens to the manifest self-wires its check run
  and required status. One exception the plan missed and the repo's own guard
  caught: `checks.mjs:DEFAULT_REVIEW_CHECKS` is the single lens list that does not
  derive from the manifest (`mark-ready.mjs` is a CLI with a top-level
  `process.exit`, so it cannot be imported by a test). `checks.test.mjs` failed
  until `agent-review-docs` was added there.
- **Two existing guards now enumerate the manifest** instead of naming five lenses
  literally. The no-clamp guard previously asserted its hardcoded list *equalled*
  the manifest, so adding a lens failed it as a name mismatch rather than
  actually checking the new rubric.
- **The `docs` lens is blocking, but its teeth are in the severity scale.** Its
  rubric caps conventions, structure, and wording at `nit` — non-blocking, and not
  even sent to the verifier — and reserves `major`/`critical` for steering text,
  secrets, and dangerous instructions. It runs at `samples: 1`: three of its four
  clauses are pattern recognition over a small flat surface, where a second sample
  mostly adds disagreement, and union turns disagreement into findings.
- **`DEFAULT_REVIEW_CHECKS` now lists only always-applicable blocking lenses.**
  It is the fallback for `mark-ready.mjs` with no `--require-checks` — the local
  preview `spec-to-pr.mjs` suggests; the promotion path always passes the
  manifest-derived set. Listing a path-scoped lens there makes the gate
  unsatisfiable whenever that lens legitimately skips, since a skipped lens posts
  `neutral` and `checkPassed` wants `success`. A residual case that pre-dates this
  change is documented in `checks.mjs`: "always applicable" is not "always has
  something to review", so a prose-only PR skips correctness and cannot satisfy
  that fallback. Unchanged for the real gate.
- **`docs` runs with `maxTurns: 8`**, matching `VERIFIER_MAX_TURNS`. A tighter
  bound is a trap: exhausting turns produces no structured output, which
  `classifyResult` reads as a retryable API error (`is_error` with no
  `api_error_status`, and `isRetryableStatus(null)` is `true`), so it retries,
  fails identically each time, and at `samples: 1` fails the lens closed as an
  *infrastructure* error — a red required check and a page. That mislabelling is
  reachable today through the verifier, which survives it by keeping the finding;
  left alone here.
- **In-flight PRs** won't show the new `agent-review-docs` check until their next
  review round.
- CODEOWNERS routes `scripts/agent/**` to `@wafflebase/maintainers`.
- UI changes: none.

## Summary by CodeRabbit

* **New Features**
  * Added documentation-focused review coverage for user-facing prose, READMEs, changelogs, and task documentation.
  * Review checks now route each change to the applicable review areas, reducing unnecessary evaluations for unrelated files.
  * Added support for scoped review execution with per-area turn limits.

* **Bug Fixes**
  * Prevented documentation-only or out-of-scope changes from incorrectly blocking required checks.
  * Improved handling of skipped and incremental reviews.

* **Tests**
  * Expanded coverage for file classification, diff routing, scope handling, and skip behavior.

### DIFF
--- a/scripts/agent/checks.mjs
+++ b/scripts/agent/checks.mjs
@@ -21,6 +21,7 @@ export const DEFAULT_REVIEW_CHECKS = [
   "agent-review-design-fit",
   "agent-review-test-adequacy",
   "agent-review-blast-radius",
+  "agent-review-docs",
 ];
 
 /** Latest run of `name` concluded success? Missing → false. */
--- a/scripts/agent/checks.mjs
+++ b/scripts/agent/checks.mjs
@@ -6,22 +6,35 @@
  * Review checks the ready gate requires when the caller passes no
  * `--require-checks` — i.e. the local preview `spec-to-pr.mjs` suggests. The
  * promotion path in agent-review-panel.yml always passes the manifest-derived
- * set, so this list cannot let a real promotion through.
+ * `blocking && applicable` set, so this list cannot let a real promotion through.
+ *
+ * It holds only the blocking lenses that apply to EVERY diff (`appliesWhen` of
+ * `**`). This caller has no panel output, so it cannot know which lenses were
+ * applicable, and `checkPassed` counts a `neutral` (skipped) check as NOT
+ * passed — so a path-scoped lens listed here makes the gate unsatisfiable for
+ * every PR that legitimately skips it (test-adequacy on a docs-only PR, docs on
+ * a code-only PR). Requiring the always-on lenses keeps the fallback meaningful
+ * while staying satisfiable on ordinary code PRs (mark-ready still fails closed
+ * on an empty set).
+ *
+ * KNOWN GAP, pre-dating file-class routing and not fixed here: "always
+ * applicable" is not the same as "always has something to review". A lens can
+ * apply and still be skipped when nothing in its `scopeClasses` changed, so a
+ * prose-only PR skips correctness and this default gate cannot be satisfied.
+ * The same hole existed before routing (test-adequacy on a docs-only PR). The
+ * real promotion path is unaffected — agent-review-panel.yml always passes the
+ * manifest-derived `blocking && applicable` set. Fixing it properly means
+ * teaching this path that a neutral required check is a skip rather than a
+ * failure, which is a change to the gate's semantics, not to this list.
  *
  * It lives here rather than in `mark-ready.mjs` for one reason: mark-ready is a
  * CLI with top-level `process.exit`, so a test cannot import it, and this is the
- * ONE lens list that does not derive itself from `lenses/lenses.json`. Adding a
- * lens and forgetting this list is a silent, invisible drift. From here
- * `checks.test.mjs` asserts it against the real manifest, so the next lens
- * cannot be added without updating it.
+ * ONE lens list that does not derive itself from `lenses/lenses.json`. Drift is
+ * silent, so `checks.test.mjs` asserts it against the real manifest.
  */
 export const DEFAULT_REVIEW_CHECKS = [
   "agent-review-correctness",
   "agent-review-security",
-  "agent-review-design-fit",
-  "agent-review-test-adequacy",
-  "agent-review-blast-radius",
-  "agent-review-docs",
 ];
 
 /** Latest run of `name` concluded success? Missing → false. */
--- a/scripts/agent/checks.test.mjs
+++ b/scripts/agent/checks.test.mjs
@@ -10,17 +10,33 @@ import { checkPassed, allRequiredPassed, DEFAULT_REVIEW_CHECKS } from "./checks.
 // added. Assert it against the REAL manifest — this test is the reason the
 // constant lives in checks.mjs at all (mark-ready.mjs is a CLI with top-level
 // process.exit and cannot be imported).
-test("DEFAULT_REVIEW_CHECKS covers every blocking lens in the manifest", () => {
+test("DEFAULT_REVIEW_CHECKS covers every ALWAYS-APPLICABLE blocking lens", () => {
   const HERE = path.dirname(fileURLToPath(import.meta.url));
   const lenses = JSON.parse(readFileSync(path.join(HERE, "lenses", "lenses.json"), "utf8"));
-  const blocking = lenses
-    .filter((l) => String(l.gating ?? "blocking") === "blocking")
-    .map((l) => `agent-review-${l.id}`);
+  const isBlocking = (l) => String(l.gating ?? "blocking") === "blocking";
+  const alwaysOn = (l) => {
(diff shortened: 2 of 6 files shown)


harrykim8672:
@claude review

harrykim8672 (dismissed):
Thanks for your contribution

harrykim8672 (approved):
.

PR #583, opened 2026-07-29
Route review findings by whether this change introduced them

hwisoo-kim:
## Summary

Split the review gate's single question into two, and answer the second with git.

The per-finding verifier asks *"is this a genuinely blocking defect in this
repository?"* — well, and biased to keep on doubt. It has never been able to ask
*"did this PR cause it?"*, and nothing else asked either. This adds a **novelty
gate** (`scripts/agent/novelty.mjs`) that answers provenance deterministically,
and replaces the verifier's keep/drop binary with **lanes**, so a finding that
stops gating is still reported rather than silently vanishing.

`isDroppingVerdict` is **unchanged**. It remains the only path to deletion.

## Why

On #578 — a PR whose contract was "behaviour-preserving extraction" — every
pre-existing bug in a function the diff moved *verbatim* into a new file was
re-reported as introduced-here, confirmed by the verifier, and handed to the fixer.

The verifier was right by its own rule. #573 deliberately stopped giving it the
diff, because a verifier reading the same diff as the lens inherits the lens's
misreadings — the correlated-error failure of naive panels. That independence is
correct and stays. But it leaves the verifier standing in the branch checkout with
no view of the base, where a line a refactor **moved** and a line the change
**wrote** are byte-identical. The `pre-existing` refutation ground is *file*-scoped
("lives in a file this change did not touch"), and a relocated function lives in a
file the PR created, so the ground correctly does not apply.

Two questions were fused into one gate. The verifier answers the first well and
the second not at all.

## How

`git blame -w -C -C -C` follows a line back through moves and copies to the commit
that **wrote** it. A second, independent probe greps the line's text against the
base tree and still answers where blame cannot (shallow clone). Both run in the
trusted script via `execFileSync` — array args, no shell, timeout — so no
capability is granted to a model and `PERMITTED_TOOLS` stays `["Read","Grep","Glob"]`.

Findings then route to a lane:

| lane | meaning | effect |
|---|---|---|
| `blocking` | real, caused by this change | fails the lens check, as today |
| `backlog` | real, but predates the branch | reported with the base location that proves it; **excluded from the check conclusion and from the fixer's checklist** |
| `discarded` | concretely refuted | dropped — via the unchanged `isDroppingVerdict` |

The fixer filter matters: without it a bounded fix round is spent rewriting code
the PR never touched, widening a diff whose contract may be behaviour-preserving.

## The one rule

**Every uncertain path yields `unknown`, and `unknown` keeps the finding
blocking** — today's behaviour exactly. Demotion requires a probe to have
*affirmatively* placed the code before the base; no combination of missing, null,
or failed evidence can produce a demoting origin. That is pinned by a test over
the whole null/false/undefined cross-product. **This gate can only remove a false
blocker. There is no path by which it loses a real one.**

Two consequences worth reviewing specifically:

- git's exit code is an **answer** for some commands (`grep` 1 = no match,
  `--is-ancestor` 1 = no) and an **error** for the rest. They are kept distinct;
  collapsing them would record a broken lookup as a confident negative and hide it
  from `unknownOrigin`, the counter whose only job is to say the gate ran blind.
- the content probe refuses non-distinctive lines. `}` occurs in any tree, and
  probing it would demote every finding that touched one (`isProbeableLine`).

## Verification

Replayed #578's findings in a worktree at the exact commit the panel reviewed
(`8e87d453f`, base `08c7885f9`):

```
relocated   ask.mjs:105  resets?\b matches ECONNRESET       already at …review-panel.mjs:436
relocated   ask.mjs:105  'rate limit' contradicts docblock  already at …review-panel.mjs:436
relocated   ask.mjs:100  structured_output before is_error  already at …review-panel.mjs:431
introduced  ask.mjs:76   deny-list is exact-string          ← the correct critical headline
introduced  ask.mjs:35   FORBIDDEN_TOOLS wrong for 0.3.217
```

All three findings the #578 disposition comment classified as "valid but
pre-existing" demote, each citing the base location that comment assembled by
hand. Both genuinely-new findings stay blocking. No false demotions.

`verify:self` green locally (11/11 lanes). 204 agent tests pass.

## Risk Assessment

- **User-facing risk:** none — this touches the review harness only.
- **Data/security risk:** none new. git runs in the trusted orchestrator via
  `execFileSync` (array args, no shell, timeout), not as a model capability;
  `PERMITTED_TOOLS` is untouched. `.git/config` is not branch-controlled, so
  aliases and diff drivers cannot be injected from a PR branch.
- **Rollback plan:** omit `--base-sha` from the workflow. Every finding then
  routes `unknown` → `blocking` and the panel behaves exactly as it does today.
  The gate is inert by construction without that flag.

## Notes for Reviewers

**AI assistance:** this PR was written by Claude Code in an interactive session —
implementation, tests, and this description. I have not yet read every line
myself, so the author checklist above is deliberately unticked.

**This PR's own panel run will not exercise the new gate.** The orchestrator and
lenses come from a trusted `main` checkout, which does not have `novelty.mjs`
until this merges. First real signal arrives on the *next* PR, via the new
`lensStats.lanes` counters (`blocking` / `backlog` / `unknownOrigin`). A high
`unknownOrigin` would mean the gate ran inert rather than that everything was
genuinely new — that is the number to watch first.

**Follow-up work.** The other #578 buckets need different mechanisms, each its own
diff — absence claims ("no CI workflow runs these tests", where refuting needs an
exhaustive search but confirming needs one hit), wrong-mechanism-but-true-kernel
findings (a binary verdict has nowhere to put them), and duplicates (`dedupeFindings`
keys on `file + lowercased summary`, so the same bug in different words survives
twice). See `docs/tasks/active/20260729-review-novelty-gate-todo.md`.

Stacked on #581, now rebased onto `main`.

## Summary by CodeRabbit

* **New Features**
  * Added a novelty gate that identifies whether reported issues were introduced, relocated, pre-existing, or uncertain.
  * Relocated findings remain visible but no longer block approval.
  * Review summaries now display demoted findings and supporting provenance when available.
  * Review metrics now include blocking, backlog, and unknown-origin outcomes.

* **Bug Fixes**
  * Review results now consistently use the calculated comparison base.
  * Non-gating findings are excluded from persisted blocking results, preventing them from affecting later review rounds.

### DIFF
--- a/.github/workflows/agent-review-on-demand.yml
+++ b/.github/workflows/agent-review-on-demand.yml
@@ -192,9 +192,14 @@ jobs:
 
       # KEEP IN SYNC with the identical step in agent-review-panel.yml.
       - name: Compute diff + changed files (data for the reviewers)
+        id: diff
         run: |
           set -euo pipefail
           git fetch --no-tags origin main
+          # Same endpoint as the `...` diff below, so "what changed" and "what
+          # novelty.mjs treats as already-there" cannot disagree. See the twin
+          # step in agent-review-panel.yml.
+          echo "base=$(git merge-base origin/main HEAD)" >> "$GITHUB_OUTPUT"
           # UNFILTERED on purpose — see the twin step in agent-review-panel.yml
           # for why the changed-file list must never shrink.
           git diff --name-only origin/main...HEAD > /tmp/changed.txt
@@ -330,6 +335,7 @@ jobs:
           --issue-file /tmp/issue.txt
           --changed-files /tmp/changed.txt
           --prior-findings /tmp/prior-findings.json
+          --base-sha "${{ steps.diff.outputs.base }}"
           --repo "$GITHUB_WORKSPACE"
           --lenses-dir .trusted/scripts/agent/lenses
           --out .agent-review
--- a/.github/workflows/agent-review-on-demand.yml
+++ b/.github/workflows/agent-review-on-demand.yml
@@ -327,6 +327,9 @@ jobs:
         continue-on-error: true
         env:
           CLAUDE_CODE_OAUTH_TOKEN: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+          # Via env, never interpolated into the `run:` string — this step holds
+          # the Claude token. Same rule as the twin step in agent-review-panel.yml.
+          BASE_SHA: ${{ steps.diff.outputs.base }}
         # --prior-findings is tolerant of a missing file (existsSync guard in
         # review-panel.mjs → treated as []), so a skipped/failed read step is safe.
         run: >-
@@ -335,7 +338,7 @@ jobs:
           --issue-file /tmp/issue.txt
           --changed-files /tmp/changed.txt
           --prior-findings /tmp/prior-findings.json
-          --base-sha "${{ steps.diff.outputs.base }}"
+          --base-sha "$BASE_SHA"
           --repo "$GITHUB_WORKSPACE"
           --lenses-dir .trusted/scripts/agent/lenses
           --out .agent-review
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -165,10 +165,18 @@ jobs:
 
       # KEEP IN SYNC with the identical step in agent-review-on-demand.yml.
       - name: Compute diff + changed files (data for the reviewers)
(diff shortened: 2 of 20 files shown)


hwisoo-kim:
@claude review

harrykim8672:
@claude review

harrykim8672 (approved):
Thanks for your contribution

PR #587, opened 2026-07-29
Verify absence claims by hunting a counterexample

hwisoo-kim:
## Summary

Split findings into **presence** and **absence** claims and verify them in
opposite directions. Refuting *"this code is wrong"* is a lookup at a location
the finding already names. Refuting *"there is no X"* means **finding an X** — a
search across the repository whose failure is indistinguishable from the thing
genuinely not existing.

The panel ran the first procedure on both, so the bias-to-keep rule that protects
real findings was rubber-stamping false absence claims.

`isDroppingVerdict` is **unchanged**. No new demotion path — see below.

## Why

From #578's disposition, the clearest "doesn't hold" finding was **"No CI
workflow runs `scripts/agent/*.test.mjs`"**. It's false: `ci.yml:41` runs
`pnpm verify:self`, whose first lane is `agent:tests` (`verify-self.mjs:16`).

The verifier confirmed it anyway, and the reason is structural rather than a weak
prompt. To refute it, the verifier must **find** the CI lane: chase `ci.yml` →
`verify:self` → `verify-self.mjs` → `agent:tests`, reading each to confirm. Three
hops against a ceiling of 8 turns. It ran out, couldn't reach a grounded
refutation, and bias-to-keep did the rest.

|  | refuted by | cost |
|---|---|---|
| presence | showing the code isn't as described | a lookup at a named location |
| absence | finding one instance of the thing | a search across the repo |

## How

- `claimType` (`presence`/`absence`, **required** so a lens decides rather than
  defaults by omission) and `searchedFor` — what the lens actually searched
  before concluding something is missing.
- An absence claim gets a **counterexample-hunting prompt**: one instance is
  enough, the thing may live under another name or be reached indirectly ("two or
  three hops from where you started"), and here is exactly what the lens already
  tried — so it searches where the lens did **not**, rather than repeating a
  search that came up empty.
- A **`counterexample`** refutation ground. The existing grounds all describe
  ways a *present* thing fails to be a defect, which is the wrong shape.
- `VERIFIER_MAX_TURNS` becomes `{presence: 8, absence: 20}`. Absence claims are
  the minority, so the extra ceiling is bounded in practice.

## `unresolved` does NOT demote — and that's the point

The plan for this PR had `unresolved` route to a non-gating lane. I dropped that
after #583's review caught the novelty gate demoting an entire legitimate class
of findings, because **the same trap is here and it's bigger.**

Absence claims are not a noise category. *"No test covers this"* is
test-adequacy's entire output contract. *"No validation on this input"* is much
of security. *"No design doc records this module"* was a **correct** #578
finding. Routing unsettleable absence claims off the merge gate would silently
disable large parts of three lenses.

So `unresolved` keeps the finding blocking exactly as `confirmed` does. Its value
is honesty and measurement: *"I searched and couldn't disprove this"* and *"I
checked and it's real"* used to be the same word, which hid how often the
verifier was guessing. The finding is marked *"(verifier could not settle this)"*
in the summary and counted separately.

**The mechanism that actually fixes #578's case is refutation, not demotion.**
That counterexample is real and findable; with 20 turns, a hunt-shaped prompt and
`searchedFor` to search around, it should come back `refuted` on the
`counterexample` ground — dropped by the rule that already exists, with cited
evidence.

## What to watch

`absenceRaised` / `absenceRefuted` / `unresolved` in the metrics comment.

- High `unresolved`, low `absenceRefuted` → absence claims are riding through
  unchecked. Either the ceiling is still too low or lenses are raising claims
  they haven't searched for; an empty `searchedFor` distinguishes the two.
- High `absenceRefuted` → #578's class is being caught where it should be.

## Verification

Both prompt branches are checked by **rendering**, following the precedent
`buildLensPrompt` set — the claim is about the prompt, and a regex over
`review-panel.mjs` cannot observe what the verifier actually reads. Tests assert
an absence prompt offers `counterexample` and `unresolved` and does *not* offer
the presence-only grounds, and that a finding with no `claimType` gets today's
procedure unchanged.

223 agent tests, `verify:self` green locally (11/11).

## Risk Assessment

- **Data/security risk:** none new. No new tool grants, no new execution, and no
  new way for a finding to leave the merge gate. `searchedFor` is model-supplied
  but advisory: it only shapes where the verifier looks, never whether a finding
  blocks.
- **Rollback plan:** a finding with no `claimType` takes the presence path, which
  is byte-identical to today's prompt and turn budget. Reverting the schema field
  alone restores current behaviour.

## Notes for Reviewers

implementation, tests, and this description. I haven't read every line myself
yet, so the author checklist is deliberately unticked.

**Deliberately not in this PR** (each its own diff): the wrong-mechanism/true-kernel
case, where a binary verdict has nowhere to put "the concern is real but the
stated cause is wrong" (#578's "Bash/Write stay available" rode through at full
severity); and duplicates, where `dedupeFindings` keys on `file + lowercased
summary` and the verifier runs per-finding in isolation, so the same bug in
different words survives twice. See
`docs/tasks/active/20260729-absence-claims-todo.md`.

Follows #583.

## Summary by CodeRabbit

* **Improvements**
  * Review findings now distinguish between claims that code is present and claims that code is absent.
  * Absence claims receive deeper verification, including counterexample searches, and may be marked unresolved when verification is inconclusive.
  * Unresolved absence findings remain relevant to review gating and are visibly labeled in severity reports.
  * Review summaries now report absence claims raised, refuted, and unresolved.
* **Documentation**
  * Updated review-panel design documentation and task tracking to describe absence-claim verification and related metrics.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -424,7 +424,12 @@ jobs:
                   // absence claim as an absence claim. Without it a carried
                   // finding defaults to `presence` and gets the wrong procedure
                   // (and the smaller turn budget) on every round after the first.
-                  .map((f) => ({ severity: norm(f.severity), file: trim(f.file, 300), claimType: f.claimType === 'absence' ? 'absence' : undefined, summary: trim(f.summary, 2000), evidence: trim(f.evidence, 2000) }));
+                  // `searchedFor` rides along for the same reason: the original
+                  // lens recorded what it already searched, and without carrying
+                  // it every repeat round tells the verifier to "search from
+                  // scratch", losing the complementary-search value. Advisory —
+                  // never gates (see verifyFinding), so it is safe to carry.
+                  .map((f) => ({ severity: norm(f.severity), file: trim(f.file, 300), claimType: f.claimType === 'absence' ? 'absence' : undefined, searchedFor: Array.isArray(f.searchedFor) ? f.searchedFor.filter((s) => typeof s === 'string' && s.trim()).slice(0, 20).map((s) => trim(s, 300)) : undefined, summary: trim(f.summary, 2000), evidence: trim(f.evidence, 2000) }));
                 // Keep output.text VALID JSON within the 60k limit: NEVER slice the
                 // serialized string (that yields JSON the next round can't parse).
                 // Fields are trimmed above; drop trailing findings until it fits.
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -420,7 +420,11 @@ jobs:
                 let blockers = (Array.isArray(v.findings) ? v.findings : [])
                   .filter((f) => norm(f.severity) === 'critical' || norm(f.severity) === 'major')
                   .filter((f) => f.lane !== 'backlog')
-                  .map((f) => ({ severity: norm(f.severity), file: trim(f.file, 300), summary: trim(f.summary, 2000), evidence: trim(f.evidence, 2000) }));
+                  // `claimType` rides along so the next round verifies an
+                  // absence claim as an absence claim. Without it a carried
+                  // finding defaults to `presence` and gets the wrong procedure
+                  // (and the smaller turn budget) on every round after the first.
+                  .map((f) => ({ severity: norm(f.severity), file: trim(f.file, 300), claimType: f.claimType === 'absence' ? 'absence' : undefined, summary: trim(f.summary, 2000), evidence: trim(f.evidence, 2000) }));
                 // Keep output.text VALID JSON within the 60k limit: NEVER slice the
                 // serialized string (that yields JSON the next round can't parse).
                 // Fields are trimmed above; drop trailing findings until it fits.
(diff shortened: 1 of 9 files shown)


hackerwins:
@coderabbitai review this.

hackerwins:
Addressed all three in 6a2242cb7 (full `verify:fast` green, +2 regression tests).

- **`refutationGround` vs `claimType` (Major)** — `isDroppingVerdict` now enforces the shape the prompt only advised. Added `DROP_GROUNDS_BY_CLAIM` (`absence` → `counterexample`/`out-of-scope`/`pre-existing`; `presence` → `not-present`/`already-guarded`/`out-of-scope`/`pre-existing`) and threaded `claimType` from the two gate paths (`routeFinding`, `verifierTally`). A model emitting the wrong shape — `not-present` for an absence claim, or `counterexample` for a presence claim — can no longer drop the finding. Omitted `claimType` preserves the prior any-ground behaviour; an unrecognized one keeps the finding (conservative).
- **`absenceRefuted` mislabel (Major)** — now counts only `refutationGround === "counterexample"` refutations, so an absence claim demoted via `out-of-scope`/`pre-existing` no longer inflates the metric `metrics.mjs` prints as "refuted by counterexample" and masks the #578 signal. `dropped` still counts those demotions.
- **`searchedFor` carry-forward (Minor)** — now rides the cross-round persist alongside `claimType` (filtered, capped at 20 × 300 chars to respect the 60k budget), so repeat rounds keep the "already searched" list instead of telling the verifier to search from scratch.

hackerwins (approved):
Thanks for your contribution.

PR #589, opened 2026-07-29
Turn on incremental review: stamp, resolve, narrow

hwisoo-kim:
## Summary

**Turns incremental review on.** #581 landed the decision logic inert; this
connects it. The panel now records what each lens reviewed, resolves the next
round's scope from those records, and reviews only the delta when every condition
holds.

> ⚠️ **This PR cannot self-test.** It is authored from a fork, and
> `agent-review-panel.yml` gates on `head_repository.full_name == github.repository` —
> so the workflow half of this change gets no panel run here. See *What is still
> unverified* before merging.

## Read this first if you read nothing else

**Round 1 after merge is still `full`** (`reason=no-prior-state` — nothing has
stamped a pointer yet). **Round 2 is the first that can narrow.** The signal to
watch is the `Review scope:` line in the run summary.

## What each piece does

| file | role |
|---|---|
| `review-scope.mjs` (new) | The caller `review-state.mjs` was written for. Reads each lens's pointer (checks API), measures the range (git), calls `resolveReviewMode`, emits `mode`/`since`/`reason`/`rounds`. |
| `gh-checks.mjs` (new) | The two check-run API contracts, stated once. |
| `review-state.mjs` | `agreedReviewedSha` exported — see *the two phases*. |
| `review-panel.mjs` | `--head-sha`; `panelEntry` owns the write rule. |
| `agent-review-panel.yml` | Scope step, narrowing step, the flags, `external_id`. |
| both workflows | Inline prior-findings `github-script` → `prior-findings.mjs`. |

## The two phases, because the inputs are circular

`resolveReviewMode` needs git facts about `since..head`. But `since` is only
knowable *after* reading the pointers — so a caller cannot gather its inputs in one
pass.

`agreedReviewedSha` is now exported, and `resolveReviewMode` calls it internally, so
the two cannot disagree about which pointer is in play. This matters because facts
measured over a *different* range would validate one range and narrow another, and
**`resolveReviewMode` cannot detect that** — it is handed the facts. The test
asserts it by capturing the argv git actually receives, not by trusting the shape.

A documented caller contract the caller cannot mechanically satisfy is a latent bug.

## The flag and the diff come from the same step

Whichever workflow step rewrites `/tmp/pr.diff` is the step that emits
`--review-mode`. They cannot disagree about what the panel is looking at.
`review-panel.mjs` cannot check this itself — it receives a *file*, not a range.

When narrowing does not happen those outputs are empty, the panel gets no flag, and
the existing full diff stands. That is why the diff step is left **byte-identical**
to the on-demand workflow's copy of it.

**Fail directions differ deliberately:**

| step | on failure | why |
|---|---|---|
| Resolve review scope | tolerated → full review | Only *computes*. `continue-on-error` covers the class the script's own `catch` cannot — `main().catch` never sees an import-time crash. A bug in a cost optimiser must not block a PR. |
| Narrow the diff | **fails the job** | *Mutates* the reviewed artifact. Tolerating a death between the `mv` and the output write would hand a lens a fragment to review as if it were the whole PR — the one outcome this design must not produce. |
| Read prior findings | tolerated → carry none | Priors can only re-raise a blocker, never clear one. |

## The write rule is a function, not a convention

`panelEntry` attaches the pointer only when a lens produced a real verdict. A
crashed, quota-failed or skipped lens stamps nothing, so the next round finds a
state gap and reviews in full. **Coverage is proven by the presence of a pointer**
rather than asserted beside it — which is why `resolveReviewMode` needs no separate
"did this lens actually review?" input.

It is a function because the first version was a rule three `panel.push` sites were
trusted to follow, guarded by a test that scanned the source for which one carried
the field. **That test passed** when the pointer was also attached to the
fail-closed entry by a separate `entry.reviewState = ...`. It was watching the
syntax I happened to write. Both remaining mutations now fail.

## Three implementations → one

The last review noted that #581 took the prior-findings read from two copies to
three. This closes that: `gh-checks.mjs` holds the two contracts, both workflows'
inline copies are gone, and `parseArgs` moved there rather than becoming a fifth
copy. `review-round-guard.mjs` is deliberately **not** refactored onto it — it feeds
the paging path, and pulling a gate file in here trades a real risk for a cosmetic
one.

On-demand already had `checks: read` (#558), and it never narrows: it records no
check runs, so it could never write a pointer back.

## Verification

`pnpm verify:self` green (11/11) · **225 tests** (224 on `main`) · both workflows
parse.

**Stubs prove the branches; one real run proves the call.** A misspelled git flag
yields `null`, resolves to `full`, and is a silent permanent no-op that no stub can
catch — so all three were run for real:

- **`gitFacts` against real history.** `origin/main~4..origin/main` →
  `isAncestor: true, hasMergeInRange: false, deltaLines: 4521`; reversed →
  `isAncestor: false`; nonexistent sha → all `null`; same sha → `0`.
- **`review-scope.mjs` end-to-end against the real API** (#581): enumerated its
  commits, fetched check runs with `--slurp` **without a parse error** — the
  contract this series has broken twice — and emitted
  `mode=full reason=no-prior-state`.
- **The narrowing path with real git**, stubbed API (nothing local can write
  `external_id`): default cap → `delta-too-large` on a real 1670-line range; raised
  cap → `incremental` with the correct `since`; `fullEvery: 1` →
  `periodic-rebaseline`.

## What is still unverified

- **`external_id` round-tripping through the checks API.** Nothing local can write
  one, and this PR gets no panel run (fork). The first real exercise is the next
  agent PR after merge.
- **Blast radius if it misbehaves:** the narrowing step either produces a delta
  *and* the flag, or fails the job. So the realistic failure is "rounds keep
  reviewing in full" — the safe direction — rather than an under-reviewed PR.
- If you would rather this ran the panel on itself first, it needs to be pushed to
  a base-repo branch instead of the fork; say so and I will redo it that way.

## Linked Issues

## Author checklist

## Verification checklist

## Risk Assessment

- **Data/security risk:** no permission or trust-model change. `external_id` sits
  behind the same `checks:write` boundary as the verdicts. `latestLensRuns` filters
  `app.slug === "github-actions"`, which narrows the writer to a workflow **in this
  repo** — the slug identifies the App, not the workflow. Only the panel declares
  `checks: write`, and no author-controlled workflow can gain it because a branch
  cannot edit `.github/workflows/**`. That push restriction is the boundary; the
  filter narrows to it rather than replacing it.
- **Rollback:** revert. Two new files plus workflow steps; reverting restores
  full-diff review immediately, and stale pointers are then simply never read.

## Notes for Reviewers

  session and reviewed by me.
- **Where I'd push hardest:** `fullEvery = 3` and `maxDeltaLines = 400` still have
  no data behind them. `fullEvery` sets how long a delta-exposed defect can hide
  *and* is the biggest lever on the saving. Both are parameters, tunable once there
  are real round counts.
- **Open interaction with #582** (`feat/file-class-routing`): it lets a lens review
  with an empty per-lens slice, which produces a valid verdict and therefore stamps
  a pointer while running no detection pass. Sound while the classification is
  right, but it makes `scopeClasses`/`classifyFile` retroactive — widening a lens's
  scope would apply only to commits after the pointer, where today it self-heals.
  Noted in the design doc; the fix is a scope digest in the state, the same
  mechanism `REVIEW_STATE_VERSION` already provides for shape changes.


hackerwins:
Closing this PR — the branch is corrupted, not conflict-resolvable.

**Commit history** is a repeating loop of throwaway commits (`base` / `extract classifyResult into moved.mjs`, 13 of them), with no coherent "stamp, resolve, narrow" change.

**Diff vs `main`**: 2997 files **removed** (effectively the entire repo — `.github/`, `CLAUDE.md`, `.gitignore`, `scripts/agent/review-panel.mjs`, …) and 3 files **added** (`banner.mjs`, `moved.mjs`, `old.mjs`). The `CONFLICTING` state comes from this branch deleting files that `main` has since modified; "resolving" it would mean either deleting the whole tree or discarding the branch entirely.

This looks like an incremental-review pipeline test artifact that got committed by mistake. The intended "Turn on incremental review: stamp, resolve, narrow" work should be redone on a fresh branch off current `main`. Closing rather than force-fixing so the history doesn't carry the destructive diff.

PR #590, opened 2026-07-30
Turn on incremental review: stamp, resolve, narrow

hwisoo-kim:
## Summary

**Turns incremental review on.** #581 landed the decision logic inert; this
connects it. The panel now records what each lens reviewed, resolves the next
round's scope from those records, and reviews only the delta when every condition
holds.

> **Replaces #589**, whose branch was polluted by another session's smoke-test
> commits. Same work, cherry-picked clean onto current `main` (post-#582/#587/#588)
> as a single commit.
>
> ⚠️ **This PR cannot self-test.** It is authored from a fork, and
> `agent-review-panel.yml` gates on `head_repository.full_name == github.repository` —
> so the workflow half of this change gets no panel run here. See *What is still
> unverified* before merging.

## Read this first if you read nothing else

**Round 1 after merge is still `full`** (`reason=no-prior-state` — nothing has
stamped a pointer yet). **Round 2 is the first that can narrow.** The signal to
watch is the `Review scope:` line in the run summary.

## What each piece does

| file | role |
|---|---|
| `review-scope.mjs` (new) | The caller `review-state.mjs` was written for. Reads each lens's pointer (checks API), measures the range (git), calls `resolveReviewMode`, emits `mode`/`since`/`reason`/`rounds`. |
| `gh-checks.mjs` (new) | The two check-run API contracts, stated once. |
| `review-state.mjs` | `agreedReviewedSha` exported — see *the two phases*. |
| `review-panel.mjs` | `--head-sha`; `panelEntry` owns the write rule. |
| `agent-review-panel.yml` | Scope step, narrowing step, the flags, `external_id`. |
| both workflows | Inline prior-findings `github-script` → `prior-findings.mjs`. |

## The two phases, because the inputs are circular

`resolveReviewMode` needs git facts about `since..head`. But `since` is only
knowable *after* reading the pointers — so a caller cannot gather its inputs in one
pass.

`agreedReviewedSha` is now exported, and `resolveReviewMode` calls it internally, so
the two cannot disagree about which pointer is in play. This matters because facts
measured over a *different* range would validate one range and narrow another, and
**`resolveReviewMode` cannot detect that** — it is handed the facts. The test
asserts it by capturing the argv git actually receives, not by trusting the shape.

A documented caller contract the caller cannot mechanically satisfy is a latent bug.

## The flag and the diff come from the same step

Whichever workflow step rewrites `/tmp/pr.diff` is the step that emits
`--review-mode`. They cannot disagree about what the panel is looking at.
`review-panel.mjs` cannot check this itself — it receives a *file*, not a range.

When narrowing does not happen those outputs are empty, the panel gets no flag, and
the existing full diff stands. That is why the diff step is left **byte-identical**
to the on-demand workflow's copy of it.

**Fail directions differ deliberately:**

| step | on failure | why |
|---|---|---|
| Resolve review scope | tolerated → full review | Only *computes*. `continue-on-error` covers the class the script's own `catch` cannot — `main().catch` never sees an import-time crash. A bug in a cost optimiser must not block a PR. |
| Narrow the diff | **fails the job** | *Mutates* the reviewed artifact. Tolerating a death between the `mv` and the output write would hand a lens a fragment to review as if it were the whole PR — the one outcome this design must not produce. |
| Read prior findings | tolerated → carry none | Priors can only re-raise a blocker, never clear one. |

## The write rule is a function, not a convention

`panelEntry` attaches the pointer only when a lens produced a real verdict. A
crashed, quota-failed or skipped lens stamps nothing, so the next round finds a
state gap and reviews in full. **Coverage is proven by the presence of a pointer**
rather than asserted beside it — which is why `resolveReviewMode` needs no separate
"did this lens actually review?" input.

It is a function because the first version was a rule three `panel.push` sites were
trusted to follow, guarded by a test that scanned the source for which one carried
the field. **That test passed** when the pointer was also attached to the
fail-closed entry by a separate `entry.reviewState = ...`. It was watching the
syntax I happened to write. Both remaining mutations now fail.

## Three implementations → one

The last review noted that #581 took the prior-findings read from two copies to
three. This closes that: `gh-checks.mjs` holds the two contracts, both workflows'
inline copies are gone, and `parseArgs` moved there rather than becoming a fifth
copy. `review-round-guard.mjs` is deliberately **not** refactored onto it — it feeds
the paging path, and pulling a gate file in here trades a real risk for a cosmetic
one.

On-demand already had `checks: read` (#558), and it never narrows: it records no
check runs, so it could never write a pointer back.

## A pointer means the lens LOOKED, not merely that it reported

#582 (per-lens diff slicing) merged while this was in flight, turning a hazard I
had only documented into a live one.

`lensReviewPlan` can return an **empty slice** — the PR contains files a lens reads,
but none changed in this round's delta — so the lens runs **zero detection samples**
and still produces a valid verdict. Stamping that pointer would claim the lens
looked at everything up to here when it looked at nothing.

The round itself is fine (there was nothing in its lane to see). The damage is that
`scopeClasses`/`classifyFile` become **retroactive**: widening a lens's scope later
would apply only to commits *after* the pointer, where today it self-heals because
every round re-reads the whole diff.

So `panelEntry` requires `ranDetection`, derived from `ok.length > 0` — the samples
that actually returned — rather than from a flag that could drift from them.
**Cost:** one forced-full round after any round where a lens had an empty slice.
A scope digest inside the state would avoid that cost and is the follow-up if those
rounds prove expensive; a permanent invisible hole is not worth trading for them now.

## Verification

`pnpm verify:self` green (11/11) · **333 tests** (331 on `main`) · both workflows
parse.

**Stubs prove the branches; one real run proves the call.** A misspelled git flag
yields `null`, resolves to `full`, and is a silent permanent no-op that no stub can
catch — so all three were run for real:

- **`gitFacts` against real history.** `origin/main~4..origin/main` →
  `isAncestor: true, hasMergeInRange: false, deltaLines: 4521`; reversed →
  `isAncestor: false`; nonexistent sha → all `null`; same sha → `0`.
- **`review-scope.mjs` end-to-end against the real API** (#581): enumerated its
  commits, fetched check runs with `--slurp` **without a parse error** — the
  contract this series has broken twice — and emitted
  `mode=full reason=no-prior-state`.
- **The narrowing path with real git**, stubbed API (nothing local can write
  `external_id`): default cap → `delta-too-large` on a real 1670-line range; raised
  cap → `incremental` with the correct `since`; `fullEvery: 1` →
  `periodic-rebaseline`.

## What is still unverified

- **`external_id` round-tripping through the checks API.** Nothing local can write
  one, and this PR gets no panel run (fork). The first real exercise is the next
  agent PR after merge.
- **Blast radius if it misbehaves:** the narrowing step either produces a delta
  *and* the flag, or fails the job. So the realistic failure is "rounds keep
  reviewing in full" — the safe direction — rather than an under-reviewed PR.
- If you would rather this ran the panel on itself first, it needs to be pushed to
  a base-repo branch instead of the fork; say so and I will redo it that way.

## Linked Issues

## Author checklist

## Verification checklist

## Risk Assessment

- **Data/security risk:** no permission or trust-model change. `external_id` sits
  behind the same `checks:write` boundary as the verdicts. `latestLensRuns` filters
  `app.slug === "github-actions"`, which narrows the writer to a workflow **in this
  repo** — the slug identifies the App, not the workflow. Only the panel declares
  `checks: write`, and no author-controlled workflow can gain it because a branch
  cannot edit `.github/workflows/**`. That push restriction is the boundary; the
  filter narrows to it rather than replacing it.
- **Rollback:** revert. Two new files plus workflow steps; reverting restores
  full-diff review immediately, and stale pointers are then simply never read.

## Notes for Reviewers

  session and reviewed by me.
- **Where I'd push hardest:** `fullEvery = 3` and `maxDeltaLines = 400` still have
  no data behind them. `fullEvery` sets how long a delta-exposed defect can hide
  *and* is the biggest lever on the saving. Both are parameters, tunable once there
  are real round counts.
- **`fullEvery`/`maxDeltaLines` aside, the #582 interaction is now handled, not
  noted** — see *A pointer means the lens looked* above. The remaining judgement
  call there is whether the forced-full rounds it costs are worth avoiding with a
  scope digest later.

## Summary by CodeRabbit

* **New Features**
  * Automated review panels now support incremental rounds, focusing on changes since the last completed review when conditions are suitable.
  * Review results retain accurate progress between rounds, helping prevent previously reviewed areas from being overlooked.
  * Prior blocking findings can be carried into subsequent review rounds.

* **Bug Fixes**
  * Improved resilience when review data or service responses are incomplete or unavailable.

* **Documentation**
  * Updated guidance and task records to explain incremental review behavior, safeguards, and failure handling.

### DIFF
--- a/.github/workflows/agent-review-on-demand.yml
+++ b/.github/workflows/agent-review-on-demand.yml
@@ -280,45 +280,25 @@ jobs:
       # findings". Best-effort (continue-on-error) and empty when there's no prior
       # panel history (e.g. a non-agent PR). This review posts NO check runs, so it
       # only READS priors and never contributes new ones.
+      #
+      # The SAME trusted script the gating panel runs, which is the point: this was
+      # the second inline copy of that logic, and two copies of a gate input is how
+      # one of them rots. Both API contracts it depends on are non-obvious and fail
+      # silently as "carried 0 findings" — see gh-checks.mjs.
+      #
+      # This review NEVER narrows its own diff: it records no check runs, so it can
+      # never write a pointer back, and a round that reviewed only a delta with no
+      # pointer to show for it would be invisible to the gating panel. Hence no
+      # review-scope step here and no --review-mode flag below.
       - name: Read prior findings (align with the gating panel)
         continue-on-error: true
-        uses: actions/github-script@v7
         env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR: ${{ needs.authorize.outputs.pr }}
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          script: |
-            const fs = require('fs');
-            const owner = context.repo.owner, repo = context.repo.repo;
-            const pr = Number(process.env.PR);
-            const manifest = JSON.parse(fs.readFileSync('.trusted/scripts/agent/lenses/lenses.json', 'utf8'));
-            const wantNames = new Set(manifest.map((l) => `agent-review-${l.id}`));
-            const commits = await github.paginate(github.rest.pulls.listCommits, { owner, repo, pull_number: pr, per_page: 100 });
-            const latest = {}; // check name -> { id, t }
-            for (const c of commits) {
-              const runs = await github.paginate(github.rest.checks.listForRef, { owner, repo, ref: c.sha, per_page: 100 });
-              for (const r of runs) {
-                if (!wantNames.has(r.name) || !r.app || r.app.slug !== 'github-actions') continue;
-                // Only COMPLETED runs carry final findings in output.text — an
-                // in-progress/queued run (e.g. a panel round running right now) has
-                // a newer started_at but no findings yet, and would otherwise shadow
-                // the last completed run. Rank by completed_at (fallback started_at).
-                if (r.status !== 'completed') continue;
-                const t = new Date(r.completed_at || r.started_at || 0).getTime();
-                if (!latest[r.name] || t > latest[r.name].t) latest[r.name] = { id: r.id, t };
-              }
-            }
-            const prior = [];
-            for (const [name, { id }] of Object.entries(latest)) {
-              const lens = name.replace(/^agent-review-/, '');
-              let text = '';
-              try { const { data } = await github.rest.checks.get({ owner, repo, check_run_id: id }); text = (data.output && data.output.text) || ''; } catch {}
-              let findings = [];
-              try { findings = JSON.parse(text); } catch {}
-              if (Array.isArray(findings)) for (const f of findings) if (f && typeof f === 'object') prior.push({ ...f, lens });
-            }
-            fs.writeFileSync('/tmp/prior-findings.json', JSON.stringify(prior));
-            core.info(`prior findings carried into on-demand review: ${prior.length}`);
+        run: >-
+          node .trusted/scripts/agent/prior-findings.mjs "$PR"
+          --lenses .trusted/scripts/agent/lenses/lenses.json
+          --out /tmp/prior-findings.json
 
       - name: Run review panel
         id: panel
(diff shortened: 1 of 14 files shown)


harrykim8672 (approved):
Thanks for contributing.

PR #591, opened 2026-07-30
Collapse restatements of one defect into a single finding

hwisoo-kim:
## Summary

Collapse **restatements** of one defect into a single finding. `dedupeFindings`
only catches byte-identical summaries, so the panel reports one bug several times
whenever a lens describes it in different words.

Merging is **not** deletion — every collapsed wording rides along, is rendered,
and reaches the fixer. The only thing removed is the count inflation.

## Why

From #578's disposition comment: *"Three findings were double-reported (the
`hunt-probe.mjs` nit, the exact-string critical, the deny-list-shape major), which
inflates the counts."*

Each pair describes one defect in different words, so each gets a different
`file + lowercased summary` key and both survive. The verifier can't help: it
judges one finding at a time in isolation, so it structurally cannot notice it's
confirming the same bug twice — and it bills for each copy.

## No model call — the metric already exists and is already calibrated

The plan for this step called for one `askStructured` call per lens. That turned
out to be unnecessary.

`findingSimilarity` in `scripts/agent/rounds.mjs` was built for the
non-convergence detector: gated on same lens + same file, using the **overlap**
coefficient — chosen over Jaccard precisely because *"the panel restates one
defect at different levels of detail, and Jaccard penalises the extra specifics in
the longer wording"*. Its threshold was calibrated against real panel output
(#564's design-fit lens emitting four wordings of one defect).

Measured against #578's real data — the three double-reports its disposition
called out, plus the `CITATION` pair, versus four genuinely distinct findings from
the same lens and file:

| | score range |
|---|---|
| same defect, different wording | **0.308 – 0.933** |
| distinct defects, same lens+file | **0.000** |

The distinct pairs score exactly 0 because `MIN_SHARED_TOKENS = 3` zeroes them, so
the real gate is "share ≥3 meaningful tokens **and** overlap ≥ 0.3". Clean
separation on real data, deterministic, free.

`DEFAULT_SIMILARITY` is imported, not redefined — a second hand-tuned copy would
be exactly the drift `REFUTATION_GROUNDS` (derived from its schema) and `CITATION`
(extracted to a leaf module) both exist to avoid.

## Merging is not deletion

This is what makes it safe without a model in the loop, and it's the principle the
novelty gate landed on after this PR series' first review:

- every collapsed wording rides along in `mergedFrom`, is rendered in the check
  body, and reaches the fix agent — so a wrong merge is **visible**, and the fixer
  still reads both descriptions;
- the survivor takes the cluster's **highest** severity, so #578's `critical`
  restated as a `major` cannot be masked;
- the survivor **gates** if any member gated (the rule `dedupeFindings` already
  has), so merging can never turn a check green;
- `unsettled` propagates, and a lane is never invented on a finding that had none.

Single-linkage, not compare-to-representative: four wordings collapse to one only
if a new wording can join on matching **any** member. Membership is decided before
a representative is chosen, and the representative is picked by a total order
(gating → severity → has-evidence → longer summary → lexicographic), so the result
doesn't depend on input order — asserted in both directions.

## Stated limitation

Two wordings of one defect that share no vocabulary score 0 and stay separate.
That leaves the count inflated, which is the status quo — the conservative
direction. `clusters.collapsed` will show whether the residue justifies a paid
pass later.

## Also fixes a bug in the merged #587

Found by **rendering** the output rather than trusting the unit tests.

`normalizeFindings` rebuilt every finding as exactly
`{severity,file,summary,evidence}`, dropping everything the orchestrator annotates
after the lens produced it. `renderSummaryMd` renders from `classify()`'s output,
so **the "verifier could not settle this" marker shipped in #587 never appeared in
a single check body** — the field was gone before the renderer looked for it.

It now coerces `severity` and preserves the rest, with a regression test on the
field set and an end-to-end one on the rendered body.

## Verification

333 agent tests, `verify:self` green locally (11/11). The clustering tests use
**verbatim** #578 summaries rather than invented strings, because the threshold is
calibrated against real panel wording and synthetic text wouldn't exercise that.

## Risk Assessment

- **Data/security risk:** none new. No new tool grants, no new execution, no model
  call added, and no new way for a finding to leave the merge gate.
- **Rollback plan:** drop the `clusterFindings(...)` wrapper in `main()`. Every
  other change is additive (`mergedFrom` rendering, counters), and with no
  clustering there is nothing to render.

## Notes for Reviewers

implementation, tests, and this description. I haven't read every line myself yet,
so the author checklist is deliberately unticked.

One pre-existing source-grep test needed loosening for the **second** time and for
the same reason: it pinned `const merged = dedupeFindings`, which legitimately
became `clusterFindings(dedupeFindings(...))`. Its own docblock argues the point —
*"a guard over an implementation detail reports refactors as breakage"* — so it now
anchors on the binding rather than on which functions produce it.

**Last remaining #578 bucket, not in this PR:** wrong mechanism / true kernel. A
binary confirmed/refuted verdict has nowhere to put "the concern is real but the
stated cause is wrong", so such a finding rides through at full severity ("Bash/
Write stay available" was overstated but had a true core). Needs a
`confirmed-corrected` verdict that may only downgrade. See
`docs/tasks/active/20260730-restatement-clustering-todo.md`.

Follows #583 and #587.

## Summary by CodeRabbit

* **Improvements**
  * Review results now consolidate repeated descriptions of the same defect, reducing inflated finding counts.
  * Consolidated findings retain the highest severity and appropriate gating behavior.
  * Review summaries and fix guidance can show alternate descriptions associated with a consolidated finding.
  * Unresolved status is preserved across merged findings.

* **Metrics**
  * Summaries now report how many repeated descriptions were combined.

* **Documentation**
  * Added guidance on restatement clustering, its safeguards, limitations, and monitoring metrics.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -523,7 +523,7 @@ jobs:
                   // it every repeat round tells the verifier to "search from
                   // scratch", losing the complementary-search value. Advisory —
                   // never gates (see verifyFinding), so it is safe to carry.
-                  .map((f) => ({ severity: norm(f.severity), file: trim(f.file, 300), claimType: f.claimType === 'absence' ? 'absence' : undefined, searchedFor: Array.isArray(f.searchedFor) ? f.searchedFor.filter((s) => typeof s === 'string' && s.trim()).slice(0, 20).map((s) => trim(s, 300)) : undefined, summary: trim(f.summary, 2000), evidence: trim(f.evidence, 2000) }));
+                  .map((f) => ({ severity: norm(f.severity), file: trim(f.file, 300), claimType: f.claimType === 'absence' ? 'absence' : undefined, searchedFor: Array.isArray(f.searchedFor) ? f.searchedFor.filter((s) => typeof s === 'string' && s.trim()).slice(0, 20).map((s) => trim(s, 300)) : undefined, mergedFrom: Array.isArray(f.mergedFrom) && f.mergedFrom.length ? f.mergedFrom.slice(0, 5).map((m) => ({ severity: norm(m.severity), summary: trim(m.summary, 500) })) : undefined, summary: trim(f.summary, 2000), evidence: trim(f.evidence, 2000) }));
                 // Keep output.text VALID JSON within the 60k limit: NEVER slice the
                 // serialized string (that yields JSON the next round can't parse).
                 // Fields are trimmed above; drop trailing findings until it fits.
@@ -801,7 +801,7 @@ jobs:
               for (const f of findings) {
                 if (!f || typeof f !== 'object') continue;
                 const file = typeof f.file === 'string' && f.file.trim() ? f.file.trim() : '(file not specified)';
-                (byFile[file] ||= []).push({ lens, severity: f.severity, summary: f.summary, evidence: f.evidence });
+                (byFile[file] ||= []).push({ lens, severity: f.severity, summary: f.summary, evidence: f.evidence, mergedFrom: f.mergedFrom });
               }
             }
             // BOUND the checklist. Each lens persists up to 60k chars of findings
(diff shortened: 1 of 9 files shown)


hwisoo-kim:
@coderabbitai full review

harrykim8672:
@claude review

harrykim8672 (approved):
Thanks for your contribution

PR #592, opened 2026-07-30
Stop a git hook's GIT_DIR from redirecting novelty's git reads

hwisoo-kim:
## Summary

While preparing #588 the tip of my working branch was silently replaced **three
times** by `novelty.test.mjs`'s 4-file fixture tree, discarding ten commits from the
branch pointer. One of those states reached #588, where it rendered as **deleting all
3019 files**. Nothing warned: tests passed, and the only symptom was the diff stat.

**Root cause: git exports `GIT_DIR` into every hook it runs, and `cwd` does not win
against it.** This repo's `.githooks/pre-push` runs `pnpm verify:self`, whose
`agent:tests` lane reaches `novelty.mjs` and its fixtures. Under the hook, those git
calls ran with `cwd` at the fixture and `GIT_DIR` at the real checkout — and git obeys
`GIT_DIR`.

That single fact explains every symptom, including the two I could not previously
account for:

| Symptom | Explanation |
|---|---|
| Only ever during `git push` | Hooks are the only context that sets `GIT_DIR` |
| Never reproducible by hand | `GIT_DIR` is unset in an interactive shell |
| 4-file tree on a **real** branch | Work tree = fixture, git dir = real repo |
| Silent | Every command succeeded; it just used the wrong repo |

Reproduced deterministically — with `GIT_DIR` set the way a hook sets it, exactly five
`[real git]` tests fail:

```
$ GIT_DIR=$(git rev-parse --absolute-git-dir) node --test *.test.mjs
✖ noveltyOf [real git]: verbatim-moved code is `relocated` and demotes
✖ noveltyOf [real git]: genuinely new code in the same new file is `introduced`
✖ noveltyOf [real git]: an untouched line is `pre-existing` and does NOT demote
✖ noveltyOf [real git]: a line the change did not add stays pre-existing in a MODIFIED file
✖ noveltyOf [real git]: a line that is only a SUBSTRING of a base line stays gating
ℹ pass 316   ℹ fail 5
```

## This is a production bug, not just a test bug

`novelty.mjs` is the novelty gate the review panel routes findings through. Because its
`git()` passed `cwd` but not the environment, **any invocation under a git hook silently
blames the wrong repository** — and the answers still come back well-formed, so a
finding gets demoted or kept on blame from an unrelated tree. A gate that is quietly
answering about the wrong repo is worse than one that is loudly off, which is the
argument `baseResolves` already makes in its own docstring.

`git()` now strips every git-location variable so `cwd` alone selects the repository.
**Deleting rather than pinning is deliberate:** `repo` may be a linked worktree, where
`.git` is a *file* holding a `gitdir:` pointer, so a computed `GIT_DIR: repo/.git` would
be wrong exactly where it matters most.

## The three changes

1. **`novelty.mjs`** — `git()` strips `GIT_DIR`, `GIT_WORK_TREE`, `GIT_COMMON_DIR`,
   `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY`, `GIT_ALTERNATE_OBJECT_DIRECTORIES`,
   `GIT_CEILING_DIRECTORIES`. This is the root-cause fix.
2. **`novelty.test.mjs`** — fixtures pin `GIT_DIR`/`GIT_WORK_TREE`/`GIT_CEILING_DIRECTORIES`
   per call, assert the resolved toplevel equals the fixture *before* writing any commit,
   and stop discarding stderr. Two new tests: one points `GIT_DIR` at a *different* repo
   and asserts `noveltyOf`'s answer is unchanged; one builds a fixture inside the real
   checkout and proves commits stay contained.
3. **`verify-self.mjs`** — reads `HEAD` before the lanes and after each one, failing if a
   lane moved it *even when that lane exited 0*. This covers the class rather than this
   one test, and names the offending lane instead of leaving a reflog to excavate.

## Reviewer's guide

- **`repoScopedEnv()` in `novelty.mjs` is the change that matters.** Everything else is
  containment. Please sanity-check the variable list for omissions.
- **`readHead()` returns `null` outside a repo** (vendored copy, exported tarball) and the
  check is then skipped rather than failing. Confirm you agree with that direction — the
  alternative is failing closed, which breaks non-git consumers.
- **The escape test asserts isolation *before* committing.** That ordering is
  load-bearing: a broken pin must trip an assertion rather than mutate the branch.

## Linked Issues

Fixes #

No issue filed — root cause is identified and fixed here. Happy to file one retroactively
for the record if you'd rather it be tracked.

## Author checklist

## Verification

- [x] `verify:self` — all 11 lanes green
- [x] `agent:tests` — **322 pass both with and without a hook-style `GIT_DIR`.** Before
      this change it passed only without; that asymmetry was the bug.
- [x] **Pushed with the `pre-push` hook enabled** — the exact path that clobbered the
      branch three times. `HEAD` unchanged, remote tip matches local, diff stat correct.
- [x] **The `verify:self` guard was proven to fire**, not merely to compile: run against a
      stand-in lane that commits and exits 0, it reported
      `✗ fake:clobber MOVED HEAD: d7a43a587 -> 12e0f8485`, printed the recovery command,
      and exited non-zero.
- [ ] verify:integration — n/a; touches only `scripts/`, which no package imports

## Risk Assessment

- **User-facing risk:** none. Nothing ships; no package imports `scripts/`.
- **Behaviour change:** `noveltyOf` now answers about `repo` in environments where it
  previously answered about `GIT_DIR`. That is the fix, but it does mean novelty verdicts
  under a hook or a `GIT_DIR`-setting CI step will change — from wrong to right.
- **Toolchain risk:** `verify-self.mjs` gates every push and CI run, so a bug here blocks
  pushes. Mitigated by exercising the whole file end-to-end (lane run, report write,
  summary, non-zero exit), not just by type-checking it.
- **Rollback plan:** revert. Three files, no config, no state, no migrations.

## Notes for Reviewers

- **AI assistance:** diagnosis and all three changes written with Claude Code (Opus 5).

- **I got the diagnosis wrong twice before getting it right**, and the sequence is worth
  recording. First I blamed the `pre-push` hook and "fixed" it with `--no-verify` — the
  reflog then showed the commits landing minutes *before* the push, so the correlation was
  real and my causal direction was invented; `--no-verify` fixed nothing. Then I concluded
  the fixture code was correct and the cause unknowable, and proposed containment alone.
  Only when the hook failed a lane that passed standalone did the actual mechanism surface.
  The lesson recorded in the lessons file is that "correlates with X" is not "caused by X",
  and that a discrepancy between two runs is more informative than either run alone.

- **`extractFailureSummary` is unreliable and cost me time here.** For the failing lane it
  reported `✔ classifyResult: distinguishes verdict / api-error / no-output` — a *passing*
  line, matched because its `/\b(FAIL|ERROR|error|Error|…)\b/` alternation hits the
  substring `error` in `api-error`. Worth tightening separately; I left it alone to keep
  this diff on one subject.

- **`tasks:archive` has a side effect worth knowing about.** Running it swept 46 completed
  task files from other branches into `docs/tasks/archive/`, including one that looks like
  live work (`20260729-absence-claims-todo.md`). That is correct behaviour per
  `tasks-archive.mjs:94` — a task stays active only while its todo has an unchecked box —
  the backlog had simply accumulated. I reverted it rather than bundle it here, but someone
  should run it deliberately.

## Summary by CodeRabbit

- **New Features**
  - Added safeguards to detect unexpected branch changes during verification and stop subsequent checks with clear failure reporting.
  - Improved Git repository isolation for automated checks, preventing commands from affecting unintended repositories.

- **Bug Fixes**
  - Prevented inherited Git settings from redirecting operations to another repository.
  - Improved diagnostics for repository and fixture-related failures.

- **Documentation**
  - Added guidance on branch integrity, secure Git execution, and preventing test fixtures from modifying surrounding repositories.

- **Tests**
  - Added coverage for repository isolation, environment handling, linked worktrees, and branch-integrity failures.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -190,6 +190,20 @@ Hook scripts live in `scripts/hooks/`.
 | `pnpm verify:entropy` | Dead-code (knip) + doc-staleness entropy gate |
 | `pnpm verify:self` | Runner: `verify:fast` + builds + chunk budgets + entropy; generates `.harness-reports/` JSON |
 
+**Branch-integrity gate.** `verify:self` records `HEAD` before the lanes and
+re-reads it after each one. A lane that moves `HEAD`, or leaves it unreadable,
+fails **even if it exited 0** — a lane can corrupt the repository and still
+report success. The lane report carries `status: "fail"` with a non-zero
+`exitCode` even when the lane's own exit code was 0, so no consumer is handed a
+report that reads as both failed and successful.
+
+The gate detects *net* movement of `HEAD` only. A lane that commits and resets
+back, rewrites another ref, or mutates the index or stash is **not** covered;
+the narrow guarantee is stated deliberately, because overclaiming invites
+trusting a green run. It exists because a test fixture escaping its temp
+directory replaced a working branch's tip three times, and one of those states
+reached a public PR as a diff appearing to delete every file in the repo.
+
 ### Integration Lanes (require database)
 
 | Command | Purpose |
@@ -497,7 +511,17 @@ Components:
   `git grep` against the base tree is a fallback for the second question only
   when move-aware blame cannot answer (shallow clone) — never an override of it.
   All probes run in the trusted script via async `execFile` (array args, no
-  shell, timeout); no capability is granted to a model. The result routes each
+  shell, timeout); no capability is granted to a model. Every git invocation is
+  built by `scripts/agent/git-env.mjs`, because **`cwd` and `-C` do not decide which repository
+  git operates on — the environment does, and it wins.** git exports its location
+  variables into every hook it runs, and `pre-push` runs `verify:self`, which
+  reaches this module; unscoped, a probe under a hook answers about whatever
+  `GIT_DIR` names. `repoScopedEnv` (reads) strips every location and behaviour
+  variable and caps discovery at the repo's parent, so a non-repo path fails
+  loudly instead of answering about an ancestor. `fixtureGitEnv` (writes, tests
+  only) additionally pins `GIT_DIR`/`GIT_WORK_TREE`; pinning `GIT_DIR` alone is
+  insufficient, because an inherited `GIT_INDEX_FILE` still makes `git add`
+  rewrite another repository's index and leaves it corrupt. The result routes each
   blocking finding to a **lane** (`routeFinding`): `blocking` gates as before,
   `backlog` is reported with the base location that proves the code already
   existed but does not gate, is **filtered out of the fixer's checklist**, and is
--- a/docs/tasks/active/20260729-autonomous-issue-hunting-lessons.md
+++ b/docs/tasks/active/20260729-autonomous-issue-hunting-lessons.md
@@ -167,3 +167,36 @@ that "is set" but may not authenticate.
 Dominated by **3.1M cache-read tokens** billed at 0.1× — without prompt caching the
 same run would be roughly an order of magnitude more. `metrics.mjs`'s existing
 `TOKEN_WEIGHTS` (`cacheRead: 0.1`) already models this correctly.
+
+## 13. A test that shells out to git can eat your branch, silently
+
+The tip of this task's branch was replaced three times by `novelty.test.mjs`'s
+4-file fixture tree, discarding ten commits from the branch pointer. One of those
+states reached PR #588, where it rendered as deleting all 3019 files.
+
+`cwd` does not confine git. In a directory that is not itself a repository, git
+walks **up** and finds the surrounding checkout — so a fixture whose `init` did not
+take effect commits into the real repo, on whatever branch is current. Work tree at
+the fixture plus git dir at the real repo yields exactly the observed shape: a
+4-file tree committed onto a live branch. Confinement needs explicit `GIT_DIR` /
+`GIT_WORK_TREE`, which disable discovery outright.
+
+Three things made it expensive rather than trivial:
+
+- **It succeeds.** An escaped fixture does not error; it commits somewhere else.
+  Tests stay green and nothing warns.
+- **`stdio: ["ignore","pipe","ignore"]` discarded stderr**, so git's own account of
+  what it did was thrown away. This became forensics — commit author identity,
+  reflog timestamps, commit-pair counts — instead of a stack trace.
+- **I diagnosed it wrong first.** All three losses coincided with `git push`, so I
+  blamed the `pre-push` → `verify:self` hook and "fixed" it with `--no-verify`. The
+  reflog then showed the commits landing ~5 minutes *before* the push. The
+  correlation was real and the causal direction was invented; `--no-verify` fixed
+  nothing and the hazard stayed live.
+
+Two habits earned: **read `git diff --stat origin/main...HEAD` before every push**
+(one second, and it is the only visible symptom), and **tag before recovering**
+(`archive/hunt-tier1-verified` is why ten commits were recoverable).
+
+And a smaller one, paid for twice: `git reset --hard` to undo a *test* also
+discards uncommitted work. Commit first, then experiment.
--- a/scripts/agent/git-env.mjs
+++ b/scripts/agent/git-env.mjs
@@ -1,0 +1,149 @@
+/**
+ * Environments for shelling out to git safely.
+ *
+ * `cwd` and `-C` do NOT decide which repository git operates on — the
+ * environment does, and it wins. That matters because **git exports its
+ * location variables into every hook it runs**, and this repo's `pre-push`
+ * hook runs `pnpm verify:self`, which reaches these scripts. Under a hook,
+ * a command that looks repo-scoped silently operates on whatever `GIT_DIR`
+ * names instead.
+ *
+ * That is not hypothetical. A test fixture inheriting these variables
+ * replaced a working branch's tip three times, and one of those states
+ * reached a public PR as a diff appearing to delete every file in the repo.
+ * Read paths answer about the wrong repository; write paths corrupt it.
+ *
+ * Two environments, because reads and writes need different strictness:
+ *
+ * - `repoScopedEnv(root)` — for READING an existing repository. Strips every
+ *   location variable so `cwd` selects the repo, and caps discovery so it
+ *   cannot ascend past `root`.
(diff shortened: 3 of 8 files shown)


hwisoo-kim:
@claude review

Root cause is identified and fixed, so the highest-value review is on whether the fix is *complete*:

1. **`repoScopedEnv()` in `novelty.mjs`** — is the stripped-variable list missing anything that can redirect git's repository resolution? That list is the whole fix.
2. **Does any other `scripts/` code shell out to git without scoping its environment?** The same class of bug would be latent there. `review-panel.mjs`, `metrics.mjs`, `spec-to-pr.mjs` and `hunt*.mjs` are the candidates.
3. **The `verify-self.mjs` HEAD guard** deliberately fails a lane that exited 0. Confirm that is right, and that `readHead()` returning `null` outside a repo should skip rather than fail.

Context worth having: this began as a silent branch rewrite that reached #588 as a diff appearing to delete all 3019 files. Please be skeptical of the claim that the mechanism is fully explained — I got the diagnosis wrong twice before this.

hwisoo-kim:
Verified every blocking claim before acting. **The critical finding was right**, two claims were wrong, and two of my own regression tests turned out to prove nothing. Fixed in 54a825b29.

## Critical — confirmed empirically, and worse than reported

The fixture pinned `GIT_DIR`/`GIT_WORK_TREE` but left `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY`, `GIT_COMMON_DIR` and `GIT_ALTERNATE_OBJECT_DIRECTORIES` inherited. Reproduced against a victim repo:

```
victim index checksum before: 6d5599df0b1a0d5c4bfc0d53549d0a7f
victim index checksum after:  e41e080c7655cfd8b99a4ada176324b4
$ git -C victim status --short
fatal: unable to read 88230b77509ff79ca2df38043a0d08f23bfd9a60
```

Not merely rewritten — left **corrupt**, because the index now references objects in the fixture's store. Pinning `GIT_DIR` alone does not stop it. Fixed by stripping all of them on the write path.

## You were right that both my regression tests were vacuous

This is the finding I'm most glad you caught, because I had presented these as proof.

- **The `GIT_DIR` test** compared two `makeRepo()` fixtures. Confirmed: identical content, identical identity, whole-second timestamps → **identical shas** (`c54374da5…` for both), so a leaked read returned the same answer. Now points `GIT_DIR` at an *empty* repo where `baseSha` cannot resolve: scoped → `relocated`, leaked → `unknown`.
- **The escape test** never set a hostile ambient environment, so it passed with the pinning removed. Now sets `GIT_DIR` + `GIT_WORK_TREE` + `GIT_INDEX_FILE` at a victim repo and asserts the victim's tip, index checksum, and index *readability* are all unchanged.

Both are now **mutation-tested** — each fails when its own fix is reverted:

```
cripple fixtureGitEnv  → ✖ fixture git cannot escape…        (27 pass, 1 fail)
cripple repoScopedEnv  → ✖ noveltyOf answers about `repo`…   (27 pass, 1 fail)
restored               → 28 pass, 0 fail
```

## `assertIsolated` was tautological — confirmed

`--show-toplevel` echoes the pinned `GIT_WORK_TREE`:

```
GIT_DIR=r1/.git GIT_WORK_TREE=fixture  rev-parse --show-toplevel   → …/fixture   (blind)
GIT_DIR=r1/.git GIT_WORK_TREE=fixture  rev-parse --absolute-git-dir → …/r1/.git  (detects)
```

Now asserts `--absolute-git-dir`, which is what decides where commits land.

## Both `GIT_CEILING_DIRECTORIES` findings — confirmed

Ceiling entries must be **strict ancestors** of the search directory, so my `GIT_CEILING_DIRECTORIES: dir` was inert. And deleting the ceiling did leave discovery free to ascend. Both fixed by capping at the **parent**, which I verified finds the repo, refuses to climb past it, fails loudly on a non-repo path, and still resolves a linked worktree's `gitdir:` pointer.

## Shared helper — you were right that a one-off left siblings exposed

Three lenses flagged this independently. Env logic now lives in `scripts/agent/git-env.mjs`, used by `novelty.mjs`, `verify-self.mjs`'s guard, and `hunt.mjs:367/377`. It also strips the config-injection and pathspec-magic variables the security lens listed, plus numbered `GIT_CONFIG_KEY_n`/`VALUE_n`.

## Guard fixes — all confirmed

Now runs **after** the exitCode split so a genuinely failing lane keeps its diagnosis; an unreadable `HEAD` **fails** rather than passing; the report carries a non-zero `exitCode` so `summarize-ci.mjs` can't render "fail (exit 0)"; the recovery hint leads with `git status && git reflog` and warns that resetting discards uncommitted work. `readHead` delegates to the scoped `readHeadSha`, and the docstring now claims only **net** `HEAD` movement.

`readHeadSha` is now tested — including a **linked-worktree fixture**, so the deletion-over-pinning rationale is verified rather than asserted.

## Two claims rejected

**`.trusted/scripts/agent/novelty.mjs` is "the copy CI actually executes"** — there is no `.trusted/` directory in this repo. `git ls-tree -r origin/main | grep -c '^\.trusted/'` → **0**. Nothing to fix.

**The `grepWholeLine` demotion bypass** — pre-existing behaviour, not introduced here. Your own test-adequacy lens flagged a sibling finding as relocated code for this reason. Worth its own issue; I'd rather not expand this diff into the demotion algorithm.

## Deliberately not addressed

`novelty-escape-*` in `.gitignore` — solved differently: the fixture now lives under `.harness-reports/` (already gitignored at `.gitignore:173`), so it's still inside a checkout, which is the condition that makes discovery dangerous, without an interrupted run leaving a nested `.git` for `git add -A` to commit as a gitlink.

`spec-to-pr.mjs`'s five git calls are left alone. They run at the repo root with no `repo` argument, so there's no scoping mismatch to fix, and the base-sha interaction you describe is real but is a behaviour question about which repo the panel should measure — worth deciding deliberately, not as a side effect of this PR.

## Verification

- **330 agent tests pass both with and without a hook-style `GIT_DIR`** (322 → 330; +8 for `git-env.mjs`)
- `verify:self` — 11/11 lanes green; `verify:entropy` clean
- Pushed **with the `pre-push` hook enabled**; `HEAD` unchanged, remote tip matches
- Guard re-verified firing after being moved past the exitCode split
- Both new invariants recorded in `docs/design/harness-engineering.md` (lane contract + novelty probe contract), as design-fit asked

@claude review

harrykim8672 (approved):
Thanks for your contribution

PR #531, opened 2026-07-23
Nest a pasted table into the table cell the caret is in (issue #528)

semimikoh:
## Summary

Pasting a table while the caret is **inside a table cell** did nothing (issue
#528). Now it nests the pasted table into that cell, matching the **Insert
Table** command and Google Docs / Word.

This PR lands both the design note and its implementation on the same branch.

**Root cause:** `insertBlocks` (`packages/docs/src/view/text-editor.ts`) inserted
a single pasted `table` block with the body-only `getBlockIndex` +
`insertBlockAt`. `EditContext` has no `'cell'` value and `getContextBlocks()`
returns only body/header/footer, so a cell target was never resolved and nothing
was inserted. The **Insert Table** command already nests via `isInCell` + a
cell-aware insert (`insertTableInCell`); the paste path just wasn't wired to it.

**Implementation:**

- `Doc.insertBlockAfter(siblingId, block)` — region/cell-aware insertion,
  delegating to the store's `findBlockInAnyArray`-backed `insertBlockAfter`
  (inserts after the sibling wherever it lives: body, header, footer, or cell).
- `cloneBlockWithFreshIds(block)` — deep-clone a (possibly nested) table with
  every block id regenerated, so a paste is independent of its source and shares
  no cell ids. This also mitigates the shared-id misroute described in #333.
- `insertBlocks`' single-table branch: when the caret is in a cell, nest via
  `insertBlockAfter`; otherwise keep the body path. The pasted table always gets
  fresh ids, and the caret lands in its first cell.

## Why

The in-cell nesting capability already existed (`insertTable` uses it); the paste
path simply wasn't wired to it. Reusing the existing `isInCell` / cell-aware
insert primitives keeps the change small and consistent with `insertTable`,
rather than reworking `EditContext` to add a first-class cell context.

Distinct from #333, which is about input being misrouted *after* pasting an
already-nested table into the body; this issue is that pasting a table into a
cell didn't nest at all.

## Linked Issues

Fixes #528

## Author checklist

- [x] I searched existing issues and PRs and confirmed this is not a duplicate.
- [x] I read every line of this diff and can explain why each change is necessary.
- [x] Changes follow the conventions in the touched packages; any deviations are explained in *Why* above.
- [x] If AI tools assisted with this PR, I noted where in *Notes for Reviewers* below.

## Verification

- [ ] verify:self — CI comment shows ✅
- [ ] verify:integration — CI comment shows ✅

Local `pnpm verify:fast` runs green.

## Test plan

[...]

Issue #528, opened 2026-07-23
Pasting a table into a table cell isn't nested


semimikoh:
### What happened:

Pasting a table while the caret is **inside a table cell** does not nest the
table into that cell. The pasted table does not appear in the clicked cell —
the paste has no visible effect.

Creating a nested table with the **Insert Table** command works correctly (the
new table nests inside the cell) — only the **paste** path is affected. So the
in-cell nesting capability exists; paste just doesn't use it.

### What you expected to happen:

Pasting a table into a table cell should **nest it into that cell**, matching:

- what the **Insert Table** command already does when the caret is inside a cell, and
- Google Docs / Word behavior.

The pasted table should appear inside the clicked cell, with the caret landing
in the first cell of the pasted table.

### How to reproduce it (as minimally and precisely as possible):

1. Open a document on https://wafflebase.io/
2. Insert a table (call it the outer table)
3. Somewhere else, insert another table, select it, and copy it (Cmd + C)
4. Click into a cell of the outer table (caret is now inside the cell)
5. Paste (Cmd + V)
6. **Expected:** the copied table nests inside the clicked cell
7. **Actual:** nothing happens — the table is not inserted into the cell

### Anything else we need to know?:

- Localized to the paste path. `insertBlocks` in
  `packages/docs/src/view/text-editor.ts` handles a single pasted `table` block
  with the **body-only** `getBlockIndex` + `insertBlockAt`. These don't account
  for the caret being inside a cell — `EditContext` is only
  `'body' | 'header' | 'footer'` (no cell context), and `getContextBlocks()`
  returns only body/header/footer blocks. So a cell target isn't resolved and
  the table isn't inserted into the cell.
- By contrast, the `insertTable` command already detects the cell via
  `layout.blockParentMap` and routes to `insertTableInCell`, so nesting works
  there. The fix is to apply the same cell detection in the paste path
  (`insertBlocks`).
- This is **distinct from #333**, which is about input being misrouted *after*
  pasting an already-nested table into the body. This issue is that pasting a
  table into a cell doesn't nest at all.
- AI assistance: root-cause analysis and phrasing with Claude Code (Claude Opus
  4.8). I observed the behavior myself.

### Environment:

- Operating system: macOS Sonoma 14.6
- Browser and version: Chrome (arm64)

PR #597, opened 2026-07-30
Extend paste-into-cell design note to cover #333

### LINKED ISSUE #333: Pasted nested tables misroute text input to the original table
### What happened:

When a table containing nested tables is copied and pasted, text input on the pasted table is misrouted. The pattern depends on the nesting depth of the cell being clicked:

- **In the innermost (deepest) nested table of the pasted copy**: the caret cannot enter the cell at all — clicking does nothing and no input is accepted. Typing is impossible.
- **In all levels above the deepest one**: the caret **does** enter the clicked cell of the pasted copy (visual selection and caret position look correct), but typed characters are still inserted into the corresponding cell of the **original** table — not the pasted one the user is clicking on. So the caret is in the right place visually, but the input destination is wrong.

This happens with both 2-level and 3-level nesting:

- **2-level nesting** (table → table): the level-2 table on the pasted copy rejects the caret entirely; the level-1 cells accept the caret but route input to the original.
- **3-level nesting** (table → table → table): the level-3 table on the pasted copy rejects the caret entirely; levels 1 and 2 accept the caret but route input to the original.

So the rule appears to be: the **deepest nested table** of the paste never receives focus, and **all levels above it accept the caret visually but leak the actual input to the source table**.

This suggests that the pasted nested tables share internal cell identifiers (or event target references) with the source, and that identifier handling fails entirely for the deepest level.

Each table produced by paste should be a fully independent instance with its own cell identifiers and event handlers. Clicking and typing into any cell of the pasted table (at any nesting level) should:

- Place the caret in the clicked cell of the pasted table.
- Insert text into that pasted cell only, with no effect on the original table.

**2-level case:**

2. Insert a table (level 1)
3. Inside a cell of the level 1 table, insert another table (level 2)
4. Select the outermost level 1 table and copy it (Cmd + C)
5. Paste it into another location in the document (Cmd + V)
6. Click into a cell of the pasted **level 2** table → the caret does not enter, no input is accepted
7. Click into a cell of the pasted **level 1** table (outside the level 2 table) and type → the caret is visually placed in the pasted cell, but typed characters appear in the corresponding cell of the **original** level 1 table

**3-level case:**

1. Same setup but add a level 3 table inside a level 2 cell
2. Copy and paste the outermost level 1 table
3. Click into a cell of the pasted **level 3** table → the caret does not enter, no input is accepted
4. Click into a cell of the pasted **level 2** table and type → caret is visually placed in the pasted level-2 cell, but characters appear in the **original** level 2 cell
5. Click into a cell of the pasted **level 1** table and type → caret is visually placed in the pasted level-1 cell, but characters appear in the **original** level 1 cell

- Across all levels above the deepest, the caret is visually correct on the pasted copy but the input destination is the original. So this is not a focus / selection bug at those levels — it's specifically an input-routing bug.
- The "deepest level rejects caret entirely, all upper levels route input to the original" pattern is consistent across 2-level and 3-level nesting. Worth checking whether 4+ level nesting follows the same rule.
- Single-level (non-nested) table copy should be tested as a control — if that one works correctly, the regression is specific to nested-table handling.
- The fact that the deepest level fails differently from the upper levels (no caret vs. wrong input destination) suggests two related bugs in the same code path: identifier cloning during paste, and event-target / focus resolution for the innermost table.
- Filing this for visibility — I'm not planning to take this on myself right now.
- AI assistance: Drafted with help from Claude Code (issue structure and pattern analysis across nesting levels). I reproduced every scenario myself and confirmed every detail in this report.

### Environment:

- Operating system: macOS Sonoma 14.6.1
- Browser and version: Chrome 148.0.7778.96 (arm64)

semimikoh:
## Summary

Design note for issue #333: pasting a table that contains **nested** tables
misroutes input on the pasted copy — the deepest nested level rejects the
caret outright, and every level above it visually accepts the caret but
routes typed characters into the corresponding cell of the **original**
table instead.

This turns out to share its root cause and its fix with #528 (already
merged, PR #531): the multi-block paste path's fresh-id regeneration only
covered a cloned block's own top-level id, not ids nested inside its table
cells, so a pasted nested table's inner cell blocks kept the source's ids.
Rather than opening a new design doc, this folds #333 into the existing
`docs-paste-table-into-cell.md` note per `docs/design/README.md`'s guidance
to fold single-PR notes into the relevant doc when they land on the same
code path.

This is the **design-note-first step** (no code yet) — opening for
direction before pushing the implementation. Implementation is already
drafted locally and will follow on this same branch.

## Why

`Doc.getBlock` / `findBlockInCells` resolved the shared ids back to the
source table for every level above the deepest (visually-correct caret,
wrong input target); the deepest level's cell wasn't in `_blockParentMap`
yet (only rebuilt at the next layout pass) and so couldn't resolve at all,
which is why it rejected the caret outright.

The note proposes: use `cloneBlockWithFreshIds` (already used for the
single-table paste branch) for the multi-block branch's middle-block clones
too — its `refreshBlockIds` already recurses through
`tableData.rows[].cells[].blocks[]`, so this alone fixes the id-sharing —
plus a `walkCellsForBlock` recursive fallback in `findBlockInCells` for
blocks not yet in `_blockParentMap` (e.g. the paste's own middle blocks).

## Test plan

No code changes yet — design note only. The note's Testing Strategy covers
the implementation that will follow: a model-level regression test
(`Doc.getBlock` resolves a block inside a freshly pasted nested table that
is not yet in `_blockParentMap`), plus manual re-verification of #333's
2-level and 3-level repro steps in the `packages/docs` standalone demo.

## Notes for Reviewers

AI tools assisted with this PR: the root-cause investigation (tracing the
id-sharing bug through `cloneBlockWithFreshIds` and `findBlockInCells`) and
the note draft were produced with Claude Code (Claude Sonnet 5). I reviewed
the root-cause analysis and every line of the note.

Fixes #333.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved paste-into-table-cell behavior for nested tables and multi-block clipboard content, preventing misrouting and crashes.
  * Enhanced deep block lookup after paste to avoid identifier conflicts and ensure the correct content is targeted.
* **Documentation**
  * Updated the paste-into-cell design documentation and added a nested-table paste routing task note.
* **Demo**
  * Added a temporary “Table” toolbar button that inserts a 2×2 table at the current caret position.

### DIFF
--- a/docs/design/docs/docs-paste-table-into-cell.md
+++ b/docs/design/docs/docs-paste-table-into-cell.md
@@ -15,11 +15,13 @@ Docs pastes all three blocks together into the cell. Today the paste path is
 body-only: the single-table case does nothing, and the multi-block case
 **crashes**.
 
-The fix for that (#528) turns out to also fix a second, related bug (#333):
-pasting a table that contains nested tables leaves the pasted copy's inner cell
-blocks sharing ids with the source, so clicking/typing into the pasted copy's
-nested table(s) silently edits the source instead. Both bugs live on the same
-paste path and are fixed by the same change, so this note covers both.
+Fixing that (#528) surfaced a second, related bug (#333): pasting a table that
+contains nested tables leaves the pasted copy's inner cell blocks sharing ids
+with the source, so clicking/typing into the pasted copy's nested table(s)
+silently edits the source instead. The two are separate fixes — `insertBlockAfter`
+chaining fixes #528's crash/no-op; recursive fresh-id cloning plus a
+`findBlockInCells` lookup fallback fix #333's id-sharing — but they land on the
+same `insertBlocks` code path in the same change, so this note covers both.
 
 ## Background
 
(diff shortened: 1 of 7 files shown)


semimikoh:
Apologies — I jumped ahead of our own process here. Per the repo's design-note-first workflow (note PR → maintainer agreement → implementation), I should have waited for direction/agreement on the design note before pushing the implementation. Instead I pushed the implementation commits (`c4e82cdbe`, `4e6b7a4f5`) onto this same branch before that step.

The implementation is up for reference/early review if useful, but I'm treating the design note as not yet agreed, and I'm happy to hold off on anything further (or roll the implementation back to a separate branch) until there's agreement on the approach. Sorry for the noise.

hackerwins (approved):
Thanks for your contribution.

PR #598, opened 2026-07-30
Count verifier errors; stop sending the changed-file list

hwisoo-kim:
## Summary

Two changes, both prompted by investigating why #592's review let a hallucinated
`.trusted/` finding and a pre-existing finding through:

1. **Count verifier errors.** A verifier session that throws currently keeps its
   finding *silently*, so an outage is indistinguishable from a verifier that
   confirmed everything.
2. **Stop sending the changed-file list**, and remove the `pre-existing`
   refutation ground it existed to support.

`isDroppingVerdict`'s rule is otherwise unchanged, and nothing about the gate's
strictness changes.

## Why: #592 shipped 40 unfiltered findings that looked verified

```
02:14:38  novelty gate: on, base d7a43a587...
02:20:28  blast-radius:  failure (infra)   ← never ran
          correctness:   failure (infra)   ← never ran
```

> Review could not run — Claude API/quota error (429): **You've hit your session
> limit · resets 4:30am (UTC)**

The run was 02:14–02:20 and the limit reset at 04:30, so once hit it held for the
entire run. Session limits are deliberately non-retryable (`ask.mjs`
`SESSION_LIMIT_RE`, correctly — they can't clear in-run), so every verifier call
after that point threw. The orchestrator caught each one and kept the finding:

```js
catch { return null; } // error → keep the finding (fail toward blocking)
```

That default is right. But `verifierTally` incremented `sentToVerifier` and then
only counted `refuted`/`unresolved` for *truthy* verdicts — so a `null` and a
`confirmed` were the same thing. The review posted 40 blocking findings with
nothing indicating the filter had never run.

**This is a trust problem, not a safety one.** A verifier outage makes the panel
*more* blocking, not less, because kept findings still gate. The danger is
reading unfiltered output as reviewed output.

## Why the observability was missing entirely

Searching the last 30 PRs for verifier numbers found **exactly one** — #564,
merged the day before #573 introduced grounded refutations. Everything since has
gone through `@claude review`, and that path recorded **nothing**: no check runs,
no stats artifact, no metrics comment. The autonomous panel's `review-panel` job
is `skipped` on every recent run (`managed == false` for un-labelled `harness/`
branches).

So #573, #583 and #587 have all shipped with zero measured effect. Fixed by
uploading the lens-stats artifact from the on-demand workflow and adding a
one-line tally to its comment, with `errored` first because a non-zero value
invalidates every number after it.

## Why the changed-file list goes

Provenance has been the novelty gate's job since #583 — answered from git, not by
asking a model to compare a path against a list. Keeping the `pre-existing`
ground meant re-sending that whole list on every verification to support a
judgement the gate already makes better.

And the ground was a path to **deleting** a finding on provenance, which is the
action #583 established is wrong: a defect whose location is old code is exactly
what the blast-radius lens exists to report. Age demotes to a reported,
non-gating lane; it never deletes. The verifier is now told explicitly not to
judge provenance at all.

Measured, since cost was the reason to look:

| | before | after |
|---|---|---|
| uncached tail per verification | finding **+ the whole changed-file list** | 268 chars — the finding alone |
| cacheable prefix | 98% | 94% |
| grows with changed-file count | yes | no |

The list sat *after* the finding, so it was re-billed uncached on every call, for
every blocking finding, every round. Removing the ground also let
`allowPreExisting` and the whole `verifyOpts` threading go — including the
"the caller MUST pass this or the gate silently re-enables a withdrawn ground"
hazard its own docblock warned about.

## Verification

345 agent tests, `verify:self` green (11/11).

The three tests pinning the old trust flag are replaced by ones asserting
provenance can no longer drop anything — including that a stale caller passing
the removed flag can't resurrect it — plus the #592 shape: every verdict `null`,
`errored == sentToVerifier`, `dropped == 0`. Rendering is asserted by rendering:
the warning must precede both the summary text and the first finding, and the
no-outage path must be byte-identical to the previous output.

## Risk Assessment

- **Data/security risk:** none new. No new tool grants or execution. Strictly
  *removes* a drop path (`pre-existing`) and removes data from the verifier
  prompt.
- **Rollback plan:** the `errored` counter and the rendered note are additive and
  inert at zero. Restoring the changed-file block means re-adding the enum value,
  the prompt block and the `allowPreExisting` gate together — they only make
  sense as a set.

## Notes for Reviewers

implementation, tests, and this description. I haven't read every line myself
yet, so the author checklist is deliberately unticked.

**Still open from the #592 investigation, not in this PR:**

- The `.trusted/` finding class. `.trusted/` is scaffolding the workflow itself
  creates (`checkout path: .trusted`, zero tracked files), and the reviewer's cwd
  is that same workspace — so a finding about it verifies as present, and
  `noveltyOf` returns `unknown` for an untracked path (confirmed by
  reproduction). The structural fix is checking scaffolding out into
  `$RUNNER_TEMP` instead of `$GITHUB_WORKSPACE` so it isn't in the review surface
  at all.
- #591 (open) collapses the 3–4× restatement visible in #592's ~40 findings —
  `readHead() inherits GIT_DIR` appears four times, `GIT_CEILING_DIRECTORIES`
  four times. That also cuts verifier session count, which is the actual driver
  of the session limit above.

Follows #583, #587, #592.

## Summary by CodeRabbit

* **Bug Fixes**
  * Review results now clearly identify findings that could not be verified.
  * Verifier errors are tracked separately from confirmed and filtered findings.
  * Provenance-based refutations no longer incorrectly remove findings from review gates.
  * Review comments preserve verifier status information even when truncated.

* **Documentation**
  * Added guidance on verifier metrics, outage behavior, and review-comment reporting.

* **Tests**
  * Expanded coverage for verifier errors, routing, tallying, and outage messaging.

### DIFF
--- a/.github/workflows/agent-review-on-demand.yml
+++ b/.github/workflows/agent-review-on-demand.yml
@@ -400,10 +400,32 @@ jobs:
               try { md = fs.readFileSync(`.agent-review/${lens.id}/summary.md`, 'utf8'); } catch {}
               return md.trim();
             });
+            // One line of verifier accounting, from the stats the panel already
+            // writes. Without it this path reported NOTHING about the filter —
+            // no check runs, no artifact, no metrics comment — which is why a
+            // session-limit outage that discarded every verdict on #592 was
+            // indistinguishable from a clean full review. `errored` leads
+            // because a non-zero value invalidates every number after it.
+            let tally = '';
+            try {
+              const stats = JSON.parse(fs.readFileSync('.agent-review/review-lens-stats.json', 'utf8'));
+              const sum = (f) => stats.reduce((n, s) => n + (Number(s?.verifier?.[f]) || 0), 0);
+              const sent = sum('sentToVerifier'), errored = sum('errored'), dropped = sum('dropped');
+              const lanes = (f) => stats.reduce((n, s) => n + (Number(s?.lanes?.[f]) || 0), 0);
+              if (sent > 0 || errored > 0) {
+                tally = '\n\n<sub>Verifier: ' + sent + ' sent, ' + dropped + ' dropped, '
+                  + sum('unresolved') + ' unresolved'
+                  + (errored ? ', **' + errored + ' ERRORED (unfiltered)**' : '')
+                  + ' · novelty: ' + lanes('backlog') + ' relocated, ' + lanes('unknownOrigin') + ' unplaceable</sub>';
+              }
+            } catch { /* stats absent (crash before write) — say nothing rather than guess */ }
             const note = '\n\n_Advisory only — this on-demand review records no status checks and does not gate merge; a human approval is still required._';
-            let body = `${marker}\n${header}\n\n${bodies.join('\n\n')}${note}`;
+            let body = `${marker}\n${header}\n\n${bodies.join('\n\n')}${tally}${note}`;
             // GitHub comment hard limit is 65536 chars; trim from the tail if huge.
-            if (body.length > 65000) body = body.slice(0, 64900) + '\n\n… _(truncated)_' + note;
+            // `tally` is re-appended after the cut, like `note`: a body long
+            // enough to truncate is one with many findings, which is exactly when
+            // "were these actually verified?" matters most.
+            if (body.length > 65000) body = body.slice(0, 64900) + '\n\n… _(truncated)_' + tally + note;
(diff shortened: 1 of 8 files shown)


harrykim8672:
@claude review

harrykim8672 (approved):
Thanks for your contribution

Issue #593, opened 2026-07-30
CLI: `docs import --replace --dry-run` is blocked by the confirmation gate


hwisoo-kim:
## Summary

`docs import --replace <id> --dry-run` refuses to run without `--yes`, but
`docs/design/cli.md:714` states the opposite:

> - `docs import --replace`: preview the PUT only; `--yes` is ignored.

The confirmation gate sits *before* the `--dry-run` branch, so a command that by
definition mutates nothing is blocked by a destructive-action prompt. In a
non-interactive context — which is every script, CI job, and agent — it does not
prompt but hard-fails.

## Reproduction

```console
$ wafflebase docs import --replace doc-123 --dry-run missing.docx
{
  "error": {
    "code": "CONFIRMATION_REQ",
    "message": "Pass --yes to confirm replacing document \"doc-123\"."
  }
}
$ echo $?
1
```

Adding `--yes` — the flag the docs say is ignored — is what lets execution past the
gate:

```console
$ wafflebase docs import --replace doc-123 --dry-run --yes missing.docx
{
  "error": {
    "code": "ENOENT",
    "message": "ENOENT: no such file or directory, open 'missing.docx'"
  }
}
```

The second failure is just the missing file. It proves the point: with `--yes` the
command proceeds to read input, without it the gate stops it first.

Note the file need not exist to reproduce — the gate at `:119` precedes `io.readBytes`
at `:143`, so nothing about the input is ever consulted.

## Cause

`packages/cli/src/docs/import.ts`:

```
:117    if (replace) {
:119      if (!yes) {
:120        if (!io.isTTY) {
:125          code: 'CONFIRMATION_REQ',
:133          return { exitCode: 1 };      // ← non-interactive: hard fail
:135        const ok = await io.confirm(…)  // ← interactive: prompts for a dry run
:143      const buf = await io.readBytes(file);
:147      if (dryRun) {                    // ← never reached without --yes
```

`dryRun` is only consulted at `:147`, after the gate has already returned. Both
branches of the gate are wrong for a dry run: the non-TTY path fails a command that
would not have changed anything, and the TTY path asks the user to confirm a
destructive action that is not going to happen.

## Also affects `notes` and `slides`

Same ordering, so presumably the same defect:

- `packages/cli/src/notes/import.ts` — gate at `:114`, `dryRun` at `:139`
- `packages/cli/src/slides/import.ts` — gate at `:146`, `dryRun` at `:174`

[...]

PR #599, opened 2026-07-30
Apply cross-sample agreement after replay, and measure its cost

hwisoo-kim:
## Summary

Cross-sample agreement ran at `hunt.mjs:537`; replay ran at `:567`. So a candidate only one sample proposed was **discarded before the free filter ran**.

Replay costs no tokens — it is trusted code re-running recorded `argv` in a clean room, three times. The old order therefore threw away, for nothing, the only evidence of whether a single-sample candidate was real. The funnel could not answer *"is 2-of-3 agreement too strict?"*, because the candidates that would answer it never ran.

```
before:  explore → agreement → novelty → replay → verify → gate
after:   explore → novelty → replay → agreement → verify → gate
                                 ↑ free filter runs first, so the drop is measured
```

**The reported set is unchanged.** Agreement still gates verification, just later. The cost is probe wall-clock; there are no extra model tokens.

## Why this is worth a diff

The last run funnelled `14 proposed → 4 agreed → 2 reported`. Ten candidates died at agreement and **nobody knows whether any were real** — they were never run. That is the largest unmeasured drop in the pipeline, and it sits directly on the recall/precision tradeoff this subsystem is built around.

The funnel becomes `proposed → unique → novel → reproduced → corroborated → reported`, plus the number that matters:

- **`soloReproduced`** — reproduced deterministically, dropped *only* for lacking a second sample.

High across runs ⇒ agreement is discarding genuine defects and the threshold needs revisiting. Persistently zero ⇒ agreement filters nothing replay would not have caught for free, and it belongs where it is.

## A count can't settle it, so the report shows them

New section, **Reproduced but not corroborated**, listing each with oracle, severity, expected, observed and citations — enough to judge by hand. Non-reproduced solo candidates are excluded: they aren't judgeable and would dilute the signal.

Solo drops are deliberately **not ledgered**. No verifier judged them, so recording them as "seen" would hide a real defect from later runs until the code in scope happened to change — the same trap the unjudged-verdict path already avoids.

## Reviewer's guide

- **The egress boundary moved, and that is the security-relevant part of this diff.** Showing dropped candidates means probe-derived text from a candidate that was never reported now reaches a file destined for a public issue. `extra` is now built from `reported` **and** `dropped` secrets. There is a mutation-tested case for it: revert the widening and `renderReport: redacts secrets carried by DROPPED candidates` fails.
- **`stats.novel` and `stats.reproduced` changed meaning** — they now count over the union rather than the agreed subset. That is intended (the funnel reads honestly top to bottom), but it does mean run-to-run comparison against older `run.json` files is not apples-to-apples for those two keys.
- **Probe wall-clock rises** roughly in proportion to how much larger the union is than the agreed set — ~3.5× more candidates probed and replayed in the last run's shape. No token cost.

## Linked Issues

Fixes #

No issue — follow-up to #588, tightening the measurement rather than the behaviour.

## Author checklist

## Verification

- [x] `agent:tests` — 359 pass (348 before, +11 here)
- [x] The redaction test is **mutation-tested**: it fails with the `extra` widening reverted
- [x] `hunt.mjs preflight` still runs clean (no SDK, no token needed)
- [ ] **A live run has NOT been done yet** — that is the next step, and it is what produces the actual measurement

## Risk Assessment

- **User-facing risk:** none. Nothing ships; `scripts/agent/` is imported by no package.
- **Behaviour risk:** the reported set should be identical, since agreement still gates verification. That is an argument from reading the control flow, **not** something a test proves — the run loop needs live SDK sessions and has no harness. The first live run is the real check.
- **Rollback plan:** revert. One module plus its tests.

## Notes for Reviewers

- **AI assistance:** written with Claude Code (Opus 5).

- **This came out of a design question worth recording.** The proposal was to skip agreement entirely and verify every candidate. Verifying all 14 would cost roughly $11/run instead of $6.20 — affordable — but the evidence points elsewhere: **#593 was reported in one run and refuted by the verifiers in the next**, and I confirmed by hand that the report was right. So verifier unanimity is *already* losing real defects, and verifying more candidates under a unanimous rule mostly moves where the loss happens. Measuring agreement's cost first is the cheaper question, and it is free. Whether to then relax unanimity to 2-of-3 is the follow-up.

- **Neither agreement nor verification protects against a systematic misreading.** Both use the same model, so if it misreads a doc sentence, three samples misread it identically and so does the verifier. Agreement only filters *stochastic* one-off confabulation. Worth being clear-eyed about what the gate does and does not buy.

## Summary by CodeRabbit

* **New Features**
  * Added a report section for candidates that reproduce consistently but lack corroboration.
  * Improved reporting of agreement and reproduction results with clearer funnel statistics.

* **Bug Fixes**
  * Expanded secret redaction to cover candidates excluded from final reports, preventing sensitive probe details from appearing in generated markdown.

* **Tests**
  * Added coverage for solo-reproduced candidate reporting and secret removal from dropped candidates.

### DIFF
--- a/scripts/agent/hunt.mjs
+++ b/scripts/agent/hunt.mjs
@@ -596,7 +596,20 @@ async function cmdRun(args) {
     }
 
     // Probe the UNION of everything any sample proposed, deduped by defect.
-    const unique = dedupeCandidates(samples.flat(), keyOf);
+    //
+    // Surface unlocatable candidates FIRST. `dedupeCandidates` silently skips any
+    // element whose key is "" (an empty key cannot dedupe against anything), so
+    // probing the raw union would erase "no locatable citation" candidates from
+    // the funnel entirely — no stat, no drop-table row — which is exactly the
+    // blindness this file's raw-proposal persistence exists to prevent. Recording
+    // them here keeps that drop measured; dedupe then runs over the locatable rest.
+    const flatProposals = samples.flat();
+    for (const cand of flatProposals) {
+      if (keyOf(cand) === "") {
+        dropped.push({ title: cand.title, why: "no locatable citation — cannot be tracked in the ledger" });
+      }
+    }
+    const unique = dedupeCandidates(flatProposals, keyOf);
     stats.unique += unique.length;
 
     const changedSince = makeChangedSince(repo, charter.codeScope);
@@ -605,11 +618,9 @@ async function cmdRun(args) {
       // already resolved this?" is a question about the defect, and the defect key
       // is the only identity available before a probe has run. The precise
       // argv+observation fingerprint is recorded alongside it as `probeFp`.
+      // `dk` is non-empty for every candidate here: `dedupeCandidates` drops empty
+      // keys, and the unlocatable ones were recorded as dropped just above.
       const dk = keyOf(cand);
-      if (dk === "") {
-        dropped.push({ title: cand.title, why: "no locatable citation — cannot be tracked in the ledger" });
-        continue;
-      }
       if (!isNovel(dk, seen, { changedSince })) {
         dropped.push({ title: cand.title, why: "already seen in a previous run (ledger)" });
         continue;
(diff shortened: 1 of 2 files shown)


harrykim8672:
@coderabbitai full review

hackerwins (dismissed):
Thanks for your contribution.

PR #600, opened 2026-07-30
Give the issue hunter the CLI: run commands, cite the transcript

hwisoo-kim:
## Summary

The issue hunter added in #588 **could not run anything.** Its grant was `Read`/`Grep`/`Glob` and its system prompt said *"you never run commands yourself"* — so it read the source, **predicted** a contradiction, and a trusted runner tested the prediction afterwards.

That removed the only reason to point it at a CLI. A CLI's unique value is being an *executable* interface; a reader that cannot execute could have been aimed at any package in the repo, and it could never be **surprised** — which is the entire product. As a static reader, scoping it to `packages/cli/src/**` was arbitrary too.

Exploration is now empirical. It runs through **one** in-process MCP tool (`scripts/agent/hunt-tool.mjs`) whose entire reachable surface is `node <cli-bin> <argv…>`.

## What is and is not being relaxed

The propose/execute split is **not** the mistake and is **not** abandoned:

| | verdict |
|---|---|
| For **reporting** | Correct — exact replay, shell injection structurally impossible. **Kept.** |
| For **exploration** | Wrong — blindness is the problem. **Replaced.** |

This is **not** `Bash`. The tool can only reach the CLI binary:

- `argv` is a string **array**; `spawnSync` with no `shell`, so `;`, `$(…)`, pipes and globs are inert. Pinned by a test passing `["docs","list","; rm -rf /","$(id)","`whoami`","a\nb","'quoted'","*"]` and asserting every element survives verbatim.
- `assertSafeArgv` hard-denies credential/login/context commands, **and now consults the CLI's own `schema` safety annotations** — `safetyOf` had been accepted by `assertSafeArgv` and passed by nobody, so 39 `read-only`/`write`/`destructive` annotations gated nothing.
- `buildProbeEnv` **replaces** the environment, so the developer's real session, config and workspace are unreachable — a property of the process, not a promise.
- Refusals are **returned as readable text, not thrown**. A thrown error aborts the call and teaches the model nothing; a refusal it can read lets it adapt instead of stalling.

## Evidence is cited, not authored

Every tool call is journalled. A candidate names `probeRefs`/`failingRef` — **indices into that journal** — so a model **cannot describe a reproduction it never performed**. `resolveProbeRefs` converts them back into the `probes`/`failingIndex` shape the gate, replay and report already speak (so nothing downstream changed) and **drops** any candidate whose references don't resolve.

The repro is now a transcript rather than a claim. That is strictly *more* trustworthy than before, not less.

## Three latent gaps closed

1. **`probeBudget.maxProbes` and `totalTimeoutMs` were declared in `charters.json` from the start and nothing ever enforced them** — only `perProbeTimeoutMs` was read.
2. **Exploration had no `maxTurns` at all**; only verifiers were capped.
3. **`safetyOf` was never passed** (above).

All three were harmless while the model couldn't act. With a tool loop they are the difference between a bounded run and an unbounded one.

## Reviewer's guide — two subtleties I'd want checked

**Help requests and `schema` are exempt from the annotation lookup.** `docs --help` reduces to the single word `docs`; the registry only names leaves (`docs.list`, `docs.export`, …); and an unresolvable lookup **fails closed**. Without the exemption the gate would refuse `--help`, the explorer's primary discovery tool, and the session would be spent on refusals. Commander prints help and exits without running any command, so a help request cannot mutate — but this is the judgement call most worth a second opinion. Hard denials are checked *before* `safetyOf`, so `api-keys revoke --help` stays denied.

**Each retry attempt gets a fresh journal, budget and scratch.** I had this wrong first: `withRetry` re-invokes the session, and a model in a second attempt counts its runs from 0 — so a shared journal would resolve `probeRefs: [0]` to the **abandoned** attempt's first command and ship a reproduction of something else.

Also worth a look: scratch is per **session**, not per call, so the model can write a fixture and then import it. Isolation comes from the replaced environment, not from discarding state.

## Linked Issues

Fixes #

No issue — follow-up to #588, redirecting a working pipeline.

## Author checklist

## Verification

- [x] `verify:self` — 11/11 lanes green; pushed **with the `pre-push` hook enabled**
- [x] `agent:tests` — **372 pass** (357 before, +15 for `hunt-tool.mjs`)
- [x] **Every safety test mutation-tested.** Removing the budget check fails 2 tests, the safety check 3, redaction 1, the journal push 1. Two of my tests on #592 passed with their own fix reverted, so a test that certifies nothing while looking like it does is the specific failure mode guarded against here.
- [x] `zod` pinned explicitly and the lockfile regenerated — it resolved only as an SDK transitive dep, and this package installs with `npm ci`
- [ ] **No live run yet.** The pipeline is exercised only by unit tests; the first real run is the acceptance test.

## Risk Assessment

- **This is the largest capability increase in the harness so far**: a model-driven loop that executes a real binary. Mitigations above; the residual risk is that `assertSafeArgv`'s hard-deny list is incomplete, which is why the schema annotations are now wired as a second layer. **Please try to break it.**
- **Egress:** probe output now flows into a model's context *before* reaching a report, and that model's output can reach a public issue. Redaction happens at the tool boundary, covered by a test asserting both the run's own API key and a `Bearer` token are stripped.
- **Cost:** tool loops add turns. Expect **~$8–12/run** against the static $6.17. `maxProbes: 40`, `explorerMaxTurns: 40`, `totalTimeoutMs: 600s`.
- **Rollback:** revert. Self-contained to `scripts/agent/` plus one design-doc section.

## Notes for Reviewers

- **AI assistance:** design and implementation with Claude Code (Opus 5).

- **This PR exists because a reviewer pushed back on my design, and they were right.** I had built a static doc-vs-code comparator and described it as a CLI explorer. The critique — *if it only reads code, why the CLI, and why not the whole repo?* — was correct in both directions, and the plan was rewritten rather than defended.

- **`codeScope` widens to the engine packages.** The CLI is a thin client, so a defect it surfaces often lives in `packages/docs`, `packages/sheets`, `packages/slides` or `packages/backend`. Scoping citations to `packages/cli` forced the hunter to blame the driver for a bug in the engine. `packages/frontend` stays out — unreachable from a CLI — and a test pins that.

- **Two new grounds, and one deliberately absent.** `self-inconsistent` and `schema-annotation-false` are both checkable from a transcript. *"This surprised me"* is **not** a ground, and never will be — it is the slop generator that ended curl's bug bounty.

- **Follow-up:** backend-bound commands (create/import/export/delete, i.e. actually exercising docs/sheets/slides) need Phase C and the `WAFFLEBASE_HUNT_WORKSPACE` safety rail, which must refuse to run against the developer's real workspace. This PR reaches `--help`, `schema`, argument validation and `--dry-run` paths.

- **#599 conflicts trivially with this** (both touch `hunt.mjs`); whichever lands second, I'll rebase.

## Summary by CodeRabbit

* **New Features**
  * Added safe, in-process CLI exploration with isolated workspaces, command validation, execution budgets, output redaction, and persistent session state.
  * Added evidence journaling so findings reference reproducible probe results.
  * Added support for approved tool integrations in agent sessions.
  * Expanded contract and crash analysis across documentation, spreadsheets, presentations, backend, and CLI sources.
  * Added verification grounds for inconsistent observations and inaccurate safety declarations.

* **Bug Fixes**
  * Invalid or unsupported probe references are now rejected.
  * Restricted commands receive readable refusals without interrupting exploration.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1089,6 +1089,34 @@ Not yet built: `round-trip` and `state` charters (need the backend tier and a
 autonomous filing behind `HUNT_FILING_ENABLED` with a mechanical accept-rate kill
 switch, and the formula differential oracle.
 
+**The explorer EXECUTES.** The first cut could not: its grant was Read/Grep/Glob and
+its prompt said *"you never run commands yourself"*, so it read the source,
+predicted a contradiction, and a trusted runner tested the prediction afterwards.
+That removed the only reason to target a CLI — a CLI's unique value is being an
+executable interface — and made surprise impossible, which is the product.
+
+The propose/execute split survives where it earns its keep. It is correct for
+**reporting** (exact replay, no shell injection) and wrong for **exploration**
+(blindness is the problem). So exploration now runs through one in-process MCP tool
+(`scripts/agent/hunt-tool.mjs`): `argv` arrays only, no shell, `assertSafeArgv` plus
+the CLI's own `schema` safety annotations, an environment `buildProbeEnv`
+**replaces**, and an enforced probe budget. Verifiers keep the read-only grant —
+handing them execution would let a verifier confirm a finding with evidence of its
+own making.
+
+Evidence is **cited, not authored**. Every tool call is journalled, and a candidate
+names `probeRefs`/`failingRef` — indices into that journal. A model therefore cannot
+describe a reproduction it never performed, which a model-authored `probes` array
+allowed; `resolveProbeRefs` converts the indices back into the shape the gate,
+replay and report already speak, and drops any candidate whose references do not
+resolve. The repro is a transcript rather than a claim.
+
+Two grounds exist only because the explorer can now act: `self-inconsistent` (two
+paths disagree about the same data) and `schema-annotation-false` (a command
+violates its own declared safety level). Both are checkable from a transcript, which
+is what stops them collapsing into *"this surprised me"* — the ground deliberately
+absent, because it is the slop generator that ended curl's bug bounty.
+
 ### Phase 27: Panel Feedback Corpus
 
 **Principle:** Entropy Management — the panel has been tuned repeatedly with no
--- a/scripts/agent/ask.mjs
+++ b/scripts/agent/ask.mjs
@@ -454,14 +454,10 @@ export async function askStructured({
   // wires `mcpServers`, so the whole session shape stays observable in one place.
   const options = buildSessionOptions({ systemPrompt, model, repo, schema, maxTurns, allowedTools, mcpServers });
 
-  // MCP tool calls are otherwise "effectively unbounded by default"
-  // (sdk.d.ts:477). Each call already carries its own probe timeout, so this is
-  // the backstop for a call that never returns at all. A process-level env side
-  // effect, so it stays OUT of the pure `buildSessionOptions`; `options.mcpServers`
-  // is present only when the grant actually wired a server.
-  if (options.mcpServers && !process.env.MCP_TOOL_TIMEOUT) {
-    process.env.MCP_TOOL_TIMEOUT = String(60_000);
-  }
+  // NOTE: `MCP_TOOL_TIMEOUT` is deliberately NOT set here. It has to be in place
+  // before the SDK is imported, and by this line it already has been — the server
+  // passed in `mcpServers` was built by `createProbeServer` (hunt-tool.mjs), which
+  // imports the SDK itself. That is where the backstop is set.
 
   const sdk = await import("@anthropic-ai/claude-agent-sdk");
   // The one place the real SDK is in hand, so the one place the cache-boundary
--- a/scripts/agent/ask.mjs
+++ b/scripts/agent/ask.mjs
@@ -57,7 +57,39 @@
  */
 export const PERMITTED_TOOLS = Object.freeze(["Read", "Grep", "Glob"]);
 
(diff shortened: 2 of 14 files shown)


hwisoo-kim:
@claude review

This is the largest capability increase in the harness so far — a model-driven loop that executes a real binary — so please weight the review accordingly:

1. **Try to break `hunt-tool.mjs`.** Is there any argv that reaches something other than `node <cli-bin> …`? Is `assertSafeArgv`'s hard-deny list plus the schema annotations sufficient, or is there a `write`/`destructive` command that resolves to `read-only` through the longest-prefix lookup?
2. **The help exemption is my least confident decision.** `isReadOnlyByConstruction` treats any argv containing `--help`/`-h`/`--version`/`-V`, or leading with `schema`, as read-only — because the annotation lookup fails closed and would otherwise refuse `docs --help`. Is that reasoning sound, and is the flag set complete?
3. **Egress.** Probe output now enters a model's context before reaching a report destined for a public repo. Redaction is at the tool boundary — is that the right place, and does it cover every path?
4. **Per-attempt state.** Each `withRetry` attempt rebuilds the journal, budget and scratch. Is there any path where a candidate's `probeRefs` could resolve against a different attempt's journal?

Known gap, stated rather than hidden: no live run yet, so the pipeline is exercised only by unit tests.

hackerwins:
@coderabbitai review this.

harrykim8672 (approved):
Thanks for your contribution

PR #601, opened 2026-07-30
Cluster restatements before the verifier, not after

hwisoo-kim:
## Summary

Follow-up to #591. That PR added `clusterFindings` to collapse restatements of one defect into a single finding, but ran it **after** the verifier — so the verifier still opened a full session (and a `git blame` via `noveltiesFor`) for **every wording** before the collapse. On the #578 shape (four restatement pairs), that was four redundant verifier sessions for defects that clustering then merged to one.

This moves the fresh-pass clustering **before** the verify pass, so the verifier runs once per distinct defect.

## Changes

- **`review-panel.mjs`** — `const detected = clusterFindings(findings)` before the verify pass; verify / `annotateFindings` / `verifierTally` now run over `detected`, not the raw `findings`.
- The merge-step `clusterFindings(dedupeFindings([...kept, ...priorKept]))` **stays** — it now only collapses **cross-pass** restatements (a prior finding the fresh pass re-found in different words).
- **`mergeCluster` flattens `mergedFrom`** instead of overwriting it. A finding can be clustered twice now (fresh pre-verify, then fresh-vs-prior), and without flattening the second pass would drop the wordings the first folded. Each folded member contributes itself + its own prior `mergedFrom`, and the representative's is kept — nothing is lost.
- `raised`/`raisedConfidence` stats stay on raw `findings` (honest raised count); `sentToVerifier` now counts distinct defects.
- Design doc + paired task/lessons files.

## The tradeoff (stated, not hidden)

#591 deliberately clustered **after** verification so it "can never decide what gets verified." This reverses that: the merge now decides what the verifier sees, so a wrong merge means a folded wording is judged only through its representative. Two things bound it:
- the similarity threshold is conservative — the #578 *distinct* pairs score **0.000**, so they stay separate;
- `mergeCluster` elects the **strongest** wording as representative (gating → highest severity → evidence-bearing), so the verifier judges the form most likely to gate, and **every folded wording is still rendered** — a bad merge is visible, not silent.

A candidate follow-up (noted in the task file): re-verify a *refuted* cluster's folded members individually before dropping them, preserving safety in the only dangerous direction while keeping the savings in the common confirmed case.

## Test plan

- New `clusterFindings: re-clustering a merged finding keeps every folded wording` regression test (asserts flattening; pre-fix it dropped a wording).
- Full agent suite: `cd scripts/agent && node --test *.test.mjs` → **358/358 pass**.

Pipeline-internal; ships no product code.

## Summary by CodeRabbit

* **Improvements**
  * Verification now runs on clustered findings, reducing redundant verification sessions for duplicate defects.
  * Clustered findings retain complete wording provenance across repeated clustering passes.
  * Reporting metrics now distinguish raw findings from distinct defects sent for verification.

* **Tests**
  * Added regression coverage to ensure re-clustering preserves all merged wording details and accurate cluster counts.

* **Documentation**
  * Documented the updated verification workflow, behavior, metrics, tradeoffs, and safeguards.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -487,12 +487,15 @@ Components:
   `verifier.absenceRaised`/`absenceRefuted`/`unresolved`, so a high `unresolved`
   against a low `absenceRefuted` is visible as absence claims riding through
   unchecked rather than being silently discounted.
-  A **clustering** pass (`clusterFindings`) then collapses RESTATEMENTS of one
-  defect. `dedupeFindings` only catches byte-identical summaries, so #578 reported
-  the same defect three times over — two wordings of one missing file, the
-  deny-list bug as both a `critical` and a `major`, the deny-list-shape concern
-  twice — and the verifier could not help, since it judges one finding at a time
-  in isolation and bills for each copy. The similarity metric is **not** re-derived:
+  A **clustering** pass (`clusterFindings`) collapses RESTATEMENTS of one defect
+  **before the verifier runs**, so each distinct defect is verified once rather
+  than once per wording. `dedupeFindings` only catches byte-identical summaries, so
+  #578 reported the same defect three times over — two wordings of one missing
+  file, the deny-list bug as both a `critical` and a `major`, the deny-list-shape
+  concern twice — and the verifier could not help, since it judges one finding at a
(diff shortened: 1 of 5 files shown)


harrykim8672:
@coderabbitai full review

harrykim8672 (approved):
Thanks for your contribution

Issue #494, opened 2026-07-18
Editing an existing link via "Edit link" inserts a new link at the caret instead of updating it


te-ing:
### What happened:

When editing an existing hyperlink in Docs, the change is not applied to the
existing link. Instead, a brand-new link is inserted at the current click/caret
position, and the original link is left unchanged.

### What you expected to happen:

Clicking "Apply" in the "Edit link" flow should update the existing link
(its URL and/or display text) in place. No new link should be inserted.

### How to reproduce it (as minimally and precisely as possible):

1. In a Docs document, select some text and insert a link
   (e.g. select "example" → Insert link → `https://example.com` → Apply).
2. Click on the linked text.
3. Choose "Edit link".
4. Change the URL (e.g. to `https://example.org`) and click "Apply".
5. Observe: the original link on "example" is unchanged, and a new link is
   inserted at the clicked position instead of the existing link being updated.

### Anything else we need to know?:

If I select the existing link by dragging over it and then change the URL, it works as expected.

### Environment:

- Operating system: macOS 15.6.1(24G90)
- Browser and version: 150.0.7871.128 (arm64)

https://github.com/user-attachments/assets/f22b2160-2f31-4675-ba8b-0bd373ef0725

te-ing:
I’d like to take this issue.

ggyuchive:
Hi @te-ing , Assigned.
Thanks!

Issue #343, opened 2026-06-05
Increasing font size on a mixed-size selection resets to the minimum


semimikoh:
### What happened:

Select text that contains **multiple font sizes** and click the **+** (increase font size) button. Instead of growing each run relative to its own size, the whole selection collapses to the **minimum** font size (effectively "restart from the smallest size").

### What you expected to happen:

Increasing the font size on a mixed-size selection should step **each run relative to its own size** (e.g. 11→12 and 36→37), matching Google Docs — not reset the entire selection to a single value.

### How to reproduce it (as minimally and precisely as possible):

1. Open a document on https://wafflebase.io/
2. On one line, make some text 11pt and some text 36pt
3. Select across both
4. Click the **+** (increase font size) button
5. Expected: each run grows by 1pt relative to its own size
6. Actual: the whole selection drops to the minimum font size

### Anything else we need to know?:

Two-layered cause in `packages/frontend/src/components/text-formatting/font-size-picker.tsx`:

1. For a mixed selection `value` is `undefined` (by design, so the input shows empty) and the draft is empty, so in `step()` `base = value ?? Number(draft)` falls to `Number("") = 0`; `0` is finite, so `clamp(0 + 1)` yields `FONT_SIZE_MIN`.
2. More fundamentally, `onChange: (size: number)` carries a single **absolute** size for the whole selection, so per-run relative stepping cannot be expressed by the current signature.
- AI assistance: drafted with help from Claude Code (repro phrasing and code-path hints). I reproduced the issue myself and confirmed the behavior.

### Environment:

- Operating system: macOS Sonoma 14.6
- Browser and version: Chrome (arm64)

semimikoh:
Found a few issues while testing, so posting them together 😅

hackerwins:
Fixed on `main` by f8c4b6397 (#616), after #358 removed the destructive collapse.

The reset came from `FontSizePicker.step()` falling back to
`value ?? Number(draft)`: on a mixed-size selection `value` is `undefined` and the
draft is empty, so the step computed from `0` and clamped to `FONT_SIZE_MIN`. That
case is now intercepted before the fallback:

```ts
if (value === undefined && draft.trim() === "") { onStepMixed?.(delta); return; }
```

(`packages/frontend/src/components/text-formatting/font-size-picker.tsx:90-105`)

The docs toolbar wires it to per-run stepping —
[...]

PR #602, opened 2026-07-30
Add the backend hunting tier behind a workspace refusal

hwisoo-kim:
> **Stacked on #600.** Base is `harness/executing-explorer`, so review that first — this PR's own diff is the last commit, `dbcb1eb8e`. GitHub will show both until #600 merges.

## Summary

Every charter until now was `mutating: false`, so the worst a probe could do was read. `round-trip` and `state` **create, rename and delete documents.** That is a real capability, and `scripts/agent/hunt-workspace.mjs` is the whole thing standing between it and a developer's real documents.

## The guarantee, stated precisely

Not *"the hunter tries to use a scratch workspace"* — that's a hope. It is: **the run refuses to start** unless a dedicated workspace was named explicitly and cannot be shown to be yours.

`WAFFLEBASE_HUNT_WORKSPACE` is refused when it:

- is unset, empty or whitespace — **there is no default and no inference from your config**, because a workspace a mutating run may destroy has to be something a human typed
- is an obviously-real name (`default`, `main`, `prod`, `production`, `personal`, …), case-insensitively
- equals `WAFFLEBASE_WORKSPACE`
- appears **anywhere** in the resolved `config.yaml` or `session.json`
- …or either file exists but **cannot be read**

That last one is the branch that keeps this from degrading to "usually safe". **Absence of proof is refusal**, because a deleted document does not come back from a reflog.

## Three decisions worth arguing about

**The collision check is a crude substring scan, on purpose.** Parsing YAML means a new dependency or a hand-rolled parser whose bugs are *silent*. A missed collision is unrecoverable; a false one costs you one renamed variable. So the crude check is the correct one.

**It honours `WAFFLEBASE_CONFIG` / `WAFFLEBASE_SESSION` overrides**, and ignoring them would have been the worst hole available: a developer on a non-default config path would get **no collision check at all**, which looks identical to a clean machine. There's a test pinning it.

**Asking the CLI would be more authoritative and doesn't work.** `status` ignores `--format json` and prints prose:

```console
$ wafflebase status --format json
Not logged in. Run `wafflebase login`.
```

That's **ground-truth defect #9** — the one command that would answer the question is one of the things this hunter exists to find. Verified empirically while building this, not assumed. Still unfiled; happy to file it separately.

## Two preconditions that fail differently on purpose

| Precondition | Outcome | Why |
|---|---|---|
| `checkStack` — `GET /health`, 2s | **Skip** the charter | Environmental and benign; mirrors `review-panel.mjs:671-674` |
| Workspace refused | **Skip**, loudly | A config error you must fix, not an environment you lack |

Neither is silent. The report renders **Charters that did NOT run** *immediately after the funnel*, because "found nothing" and "never executed" are otherwise the same zero — and only one of them means the code is fine. A test pins that this section precedes the reassuring "No candidates reported" text.

`checkStack` never throws: a connection refusal, a non-2xx, and an abort all return `{up:false}`. Throwing would abort the whole run and take the backend-*free* charters with it, which is the opposite of the intent.

## Cleanup can only see what it created

Seeded documents are named `hunt-<runId>-<n>`. `hunt.mjs cleanup --run <id>` deletes **that exact prefix and nothing else**, and prints every decision — deleted, failed, and *left alone*.

- **`--run` is mandatory**, so there's no "delete anything fixture-shaped" mode to reach by accident.
- The run id is in the prefix, so concurrent runs can't delete each other's fixtures (which would look exactly like flake).
- `isSeeded` is **exact-prefix, not `contains`** — a document you named `my-hunt-abc123-notes` is never matched. Pinned by a test, and mutation-tested.

## Reviewer's guide

- **`hunt-workspace.mjs` is the file to read.** Everything else here is plumbing. All seven refusals are mutation-tested — each one fails with its guard removed.
- **One existing test was replaced, deliberately.** A blanket *"no charter mutates"* assertion was true only while tier 1 was the only tier. It's now the property that actually protects data: **`mutating ⟹ needsBackend`**, because `needsBackend` is what routes a charter through *both* gates. A charter marked mutating but not needsBackend would skip the workspace refusal entirely and run against whatever the ambient config resolves to.
- **The mutating charters' rubrics carry the naming rule**, injected with the live seed prefix. A document named anything else is invisible to cleanup and orphaned forever.

## Linked Issues

Fixes #

No issue — Phase C of the hunting pipeline (#588 → #600 → this).

## Author checklist

## Verification

- [x] `agent:tests` — **389 pass** (372 before, +17 here)
- [x] **All seven workspace refusals mutation-tested**, plus `isSeeded` weakened to `contains` and the config override ignored — every one produces failures
- [x] `checkStack` verified against the **real down stack** on this machine, plus injected non-2xx / refused / aborted branches
- [x] `cleanup` refuses without `--run`, refuses `default` as a workspace, and refuses with no backend — all three exercised by hand
- [x] `verify:entropy` clean; `preflight` reports the new tier's prerequisites as `WARN`, not `ok`
- [ ] **No live backend run.** Docker isn't up on this machine, so `round-trip`/`state` have never executed. Gate before believing them: `cleanup` provably removes every seeded document, a stack-down run skips, and a run pointed at a real workspace is refused.

## Risk Assessment

- **This PR introduces the first destructive capability in the harness.** Mitigations above. The residual risk is a workspace that passes every check but is still real — most plausibly a developer who sets `WAFFLEBASE_HUNT_WORKSPACE` to something genuine that happens to appear nowhere in their config. I could not close that without asking the CLI, which is broken. **Please try to find a path through.**
- **Blast radius if the rail fails:** documents deleted in whatever workspace resolves. Unrecoverable. This is why absence of proof refuses.
- **Rollback:** revert. Self-contained to `scripts/agent/` plus one design-doc section.

## Notes for Reviewers

- **AI assistance:** design and implementation with Claude Code (Opus 5).

- **Scope, honestly stated.** This completes the CLI hunter. It does **not** deliver what was originally asked for: the CLI cannot type into a document (`docs.content` is `read-only`, and no command inserts text), so "act like a user and see what breaks" is unreachable here by design. Only `sheets.cells.set` genuinely mutates content. A **UI hunter** driven by Playwright over the existing `verify:frontend:interaction` + Docker Chromium lanes is the next piece, and it's the one that matches the original intent — most open bugs (#494, #343, #333) are rendering and interaction issues no CLI can reach.

- **Docker is required to validate this**, and it isn't running here. I built and tested everything that can be tested without it — which is all of the safety logic, deliberately, since that's the part that must be right before the first live run.

## Summary by CodeRabbit

* **New Features**
  * Issue hunting can now execute CLI probes and use observed results as evidence.
  * Added Round-trip and State hunting checks for data fidelity and lifecycle problems.
  * Added backend and workspace safety checks, with clear reporting for skipped checks.
  * Added scoped cleanup for test data created during a hunt.
* **Bug Fixes**
  * Improved candidate validation to reject unsupported or unverifiable findings.
  * Added safeguards against unsafe commands, workspace collisions, leaked secrets, and excessive probing.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1116,10 +1116,11 @@ Done criteria for Tier 1 (met): a local run reports at least one defect that a
 human independently confirms, with **zero** reports a human judges wrong, and the
 funnel explains every dropped candidate.
 
-Not yet built: `round-trip` and `state` charters (need the backend tier and a
-`WAFFLEBASE_HUNT_WORKSPACE` safety rail), the rolling GitHub-issue report,
-autonomous filing behind `HUNT_FILING_ENABLED` with a mechanical accept-rate kill
-switch, and the formula differential oracle.
+Not yet built: the rolling GitHub-issue report, autonomous filing behind
+`HUNT_FILING_ENABLED` with a mechanical accept-rate kill switch, and the formula
+differential oracle. (The `round-trip`/`state` backend charters and the
+`WAFFLEBASE_HUNT_WORKSPACE` safety rail shipped — see the backend-tier section
+below.)
 
 **The explorer EXECUTES.** The first cut could not: its grant was Read/Grep/Glob and
 its prompt said *"you never run commands yourself"*, so it read the source,
@@ -1149,6 +1150,44 @@ violates its own declared safety level). Both are checkable from a transcript, w
 is what stops them collapsing into *"this surprised me"* — the ground deliberately
 absent, because it is the slop generator that ended curl's bug bounty.
 
+**Backend tier, and the refusal that gates it.** `round-trip` (metamorphic
+identities: import→export, batch ≡ N×set, `--pages` partitions, `--dry-run` mutates
+nothing) and `state` (create→delete→list, rename twice, delete twice) are
+`needsBackend: true` and `mutating: true`. Their oracles are self-evidencing — a
+broken relation needs no written promise — so they set `requiresDocCitation: false`
+while still demanding a code citation, or a finding would be unactionable.
+
+Two preconditions run before a token is spent, and they fail differently on purpose.
+A missing stack (`checkStack`, `GET /health`, 2s) SKIPS the charter, mirroring the
+review panel's treatment of its own missing preconditions. A refused workspace also
+skips, but is a configuration error rather than an environmental one. Neither is
+silent: the report renders **Charters that did NOT run** immediately after the
+funnel, because "found nothing" and "never executed" are otherwise the same zero and
+only one of them means the code is fine.
+
+`scripts/agent/hunt-workspace.mjs` is the whole guarantee. `WAFFLEBASE_HUNT_WORKSPACE`
+must be set explicitly — there is no default and no inference from the developer's
+config, because a workspace a mutating run may destroy has to be something a human
+typed. It is refused if it is an obviously-real name, if it equals
+`WAFFLEBASE_WORKSPACE`, if it appears anywhere in the resolved config.yaml or
+session.json, or if either file exists but cannot be read. Absence of proof is
+refusal, since a deleted document does not come back from a reflog. The collision
+check is a deliberately crude substring scan honouring `WAFFLEBASE_CONFIG` /
+`WAFFLEBASE_SESSION` overrides: parsing YAML would mean a new dependency or a
+hand-rolled parser whose bugs are silent, and a missed collision is unrecoverable
+while a false one costs one renamed variable.
+
+Asking the CLI for its resolved workspace would be more authoritative and does not
+work: `status` ignores `--format json` and prints prose. That is ground-truth defect
+\#9 — the one command that would answer the question is one of the things this
+hunter exists to find.
+
+Every seeded document is named `hunt-<runId>-<n>`, and `hunt.mjs cleanup --run <id>`
+deletes by that exact prefix and nothing else, printing every decision including what
+it left alone. `--run` is mandatory so there is no "delete everything that looks like
+a fixture" mode to reach by accident, and the run id is in the prefix so concurrent
+runs cannot delete each other's fixtures. Docker lifecycle is NOT reimplemented —
+`scripts/verify-integration-docker.mjs` owns it, and `preflight` reports what is missing.
 ### Phase 27: Panel Feedback Corpus
 
 **Principle:** Entropy Management — the panel has been tuned repeatedly with no
--- a/scripts/agent/charters/charters.json
+++ b/scripts/agent/charters/charters.json
@@ -76,5 +76,85 @@
     },
     "verifierMaxTurns": 20,
     "explorerMaxTurns": 55
+  },
+  {
+    "id": "round-trip",
+    "title": "Round-trip and metamorphic relations",
+    "model": "claude-opus-5",
+    "samples": 3,
+    "verifiers": 2,
+    "oracles": [
+      "round-trip"
+    ],
+    "surface": "cli",
+    "needsBackend": true,
+    "mutating": true,
+    "codeScope": [
+      "packages/cli/src/**",
+      "packages/docs/src/**",
+      "packages/sheets/src/**",
+      "packages/slides/src/**",
+      "packages/backend/src/**"
+    ],
+    "docsScope": [
+      "packages/cli/README.md",
+      "docs/design/cli.md",
+      "docs/design/rest-api.md",
+      "packages/cli/skills/**"
+    ],
+    "requiresDocCitation": false,
+    "minCitations": 1,
+    "explorerMaxTurns": 75,
+    "verifierMaxTurns": 20,
+    "reportableSeverities": [
+      "critical",
+      "major"
+    ],
+    "maxCandidates": 6,
+    "probeBudget": {
+      "maxProbes": 60,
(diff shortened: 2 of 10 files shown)


hwisoo-kim:
@claude review

This PR introduces the **first destructive capability** in the harness — charters that create, rename and delete documents — so please weight it accordingly. `scripts/agent/hunt-workspace.mjs` is the file that matters; everything else is plumbing.

1. **Try to find a path through the workspace refusal.** `WAFFLEBASE_HUNT_WORKSPACE` must be explicit, must not be an obviously-real name, must not equal `WAFFLEBASE_WORKSPACE`, must not appear in the resolved `config.yaml`/`session.json`, and any unreadable file refuses. What configuration passes all five and is still a real workspace?
2. **The collision check is a substring scan, not a parse.** Deliberate — a hand-rolled YAML parser's bugs are silent, and a missed collision is unrecoverable. Is that the right trade, and does honouring `WAFFLEBASE_CONFIG`/`WAFFLEBASE_SESSION` cover every path the CLI itself reads?
3. **`isSeeded` is the only filter on the delete path.** Exact-prefix, so `my-hunt-abc123-notes` is never matched. Is there any name that should match and doesn't, or worse, shouldn't and does?
4. **I replaced an existing test.** The blanket "no charter mutates" assertion became `mutating ⟹ needsBackend`, because `needsBackend` is what routes a charter through both gates. Is that the right invariant, or did I weaken something?

Two things stated rather than hidden: **no live backend run** (docker isn't up here, so `round-trip`/`state` have never executed), and the residual risk I could not close — a developer who sets a genuine workspace name that happens to appear nowhere in their config. Closing that would need the CLI to report its resolved workspace, and `status --format json` prints prose instead (ground-truth defect #9).

hwisoo-kim:
Rebased onto `main` now that #600 has merged, and verified all eight findings against the rebased code. **All eight were already fixed** — five by later work on this branch, one by #600's own review round, and two were never present as described. Details, with the checks I ran:

| Finding | Verdict |
|---|---|
| `HARD_DENIED_COMMANDS` bypassed under `mutating` | **Already fixed.** Verified: `api-keys revoke`, `login`, `logout`, `ctx switch` all refuse with `mutating: true`. The rationale in `hunt-probe.mjs` is worth reading — `ctx switch` would have moved probes *off* the isolated hunt workspace, defeating the refusal the whole tier depends on. |
| `seedPrefix` overlap on nested run ids | **Already fixed**, exactly as suggested — length-prefixed. `seedPrefix("abc")` → `hunt-3-abc-`, `seedPrefix("abc-123")` → `hunt-7-abc-123-`, so `isSeeded("hunt-7-abc-123-4", "abc")` is `false`. Regression test at `hunt-workspace.test.mjs:143-149`. |
| Shared `context.cfg` mutated per charter | **Already fixed** — `charterContext = { ...context, cfg: { ...context.cfg, workspace } }`, so a later non-mutating charter is unaffected. |
| Hard-fail on `needsBackend` in `cmdRun` | **Not present.** Backend charters reach the per-charter `checkStack` and are skipped loudly; nothing rejects them up front. |
| Stale "Not yet built" list | **Already fixed** — it now records that the `round-trip`/`state` charters and the `WAFFLEBASE_HUNT_WORKSPACE` rail shipped, pointing at the backend-tier section. |
| Cleanup accepting guessed wrappers | **Already fixed** — an unrecognised payload shape is a **failure**, not an empty list, with an explicit "refusing to assume it is empty" message. A seeded entry with no id also fails, and `process.exitCode = 1`. |
| Help exemption vs option values | **Fixed in #600.** Your exact example: `docs delete d1 --force --help --pages 1` → **refused**; `docs content d1 --force --help --pages 1` → allowed (read-only); `docs --help` → still allowed. The exemption only rescues an *unresolvable* annotation and never overrides a resolved one. |
| Nitpick: preflight re-loads charters | **Already fixed** — `allCharters` is reused, the backend note is keyed on `needsBackend`, and the workspace check is gated on `mutating` so read-only backend charters don't emit workspace warnings. |

## What the rebase involved, since it was not a fast-forward

This branch carried a **stale duplicate of the #600 commit** that is now in `main`, so I replayed only this PR's own commit with `--onto` and resolved five conflicts:

- **`hunt-tool.mjs` and its test → took `main`'s version.** This branch had independently performed the same lazy-import split (as `buildProbeHandler`), but without #600's merged review fixes. Taking `main`'s keeps the session-budget timeout clamp, redact-before-clip, and the narrowed help exemption. `hunt-tool.mjs` is consequently no longer part of this PR's diff at all, which is correct.
- **`charters.json`** — kept the two new charters, then re-applied #600's turns-must-exceed-probes rule to all four. `round-trip`/`state` are now 75 turns / 60 probes.
- **`hunt.mjs`** — kept the hoisted `const wafflebase = await createProbeServer(...)` (cleaner than `await` inside an object literal) and combined the fuller stats rationale with the `skipped` declaration.
- **`harness-engineering.md`** — kept **both** sections; `main` had added Phase 27 from #608. Ordering is 25 → 26 → 27.

## Verification

- `verify:self` — 11/11 lanes green
- `agent:tests` — **474 pass**
- Every finding above re-checked by running the code, not by reading it

Still true and worth restating: **no live backend run.** Docker is not up here, so `round-trip` and `state` have never executed. The gate before trusting them is that `cleanup` provably removes every seeded document, a stack-down run skips, and a run pointed at a real workspace is refused.

hackerwins (approved):
Thanks for your contribution.

PR #603, opened 2026-07-30
Attribute review-panel tokens per lens and detection/verifier

hwisoo-kim:
## Summary

The review panel is expensive enough that two runs can exhaust a session limit, but the logs couldn't say **where** the tokens went: `review-execution.json` was a flat, un-tagged list of SDK result messages (`ask.mjs` did a bare `sessionLog.push(message)`), and the effort summary collapsed it to one aggregate. So "which lens?" and "is it the verifier?" were unanswerable from the ledger.

This adds a per-lens / detection-vs-verifier token breakdown, surfaced in the effort summary you already get.

## How

- **`ask.mjs`** — `askStructured` takes an optional `logMeta` and stamps it onto the logged result message (**shallow copy** — the SDK message is not mutated; every field is preserved so the log stays `claude-execution-output`-shaped).
- **`review-panel.mjs`** — `runLens` tags `role: "detection"`; `verifyFinding` takes a `lensId` and tags `role: "verifier"`. Both verifier call sites (representative + fold re-verify, and the prior-round re-check) pass `lensId`, so verifier cost isn't undercounted.
- **`metrics.mjs`** — `attributionBreakdown(messages)` → `{ lens: { detection|verifier|other: {cost,weighted,tokens,turns,calls} } }`; `aggregateAttribution` sums it across rounds; `cmdRecord` attaches it to the review record; `renderSummary` renders a per-lens **cost · weighted-tokens** table plus a detection-vs-verifier headline.

Example (rendered in the `🤖 Agent effort` → Review panel section):

```
#### Token attribution — cost · weighted tokens, all rounds
| Lens | Detection | Verifier |
| --- | --- | --- |
| correctness | $0.30 · ~120K | $0.50 · ~400K |
| security | $0.20 · ~90K | — |
| **Total** | $0.50 · ~210K | $0.50 · ~400K |

- Detection vs verifier: $0.50 vs $0.50 (~210K / ~400K weighted).
```

## Safe by construction

- **No workflow change** — the attribution rides inside `review-execution.json`, which `metrics.mjs record --kind review` already reads on both the autonomous and on-demand paths.
- **Backward compatible** — a message with no attribution → `other` bucket; a round with none → no table at all, so pre-instrumentation PRs render exactly as before.
- **Measurement only** — gates nothing; `metrics.mjs` stays fail-safe (exits 0 on any error).

## Test plan

- New tests: `attributionBreakdown`, `aggregateAttribution`, the rendered table, and the no-attribution (no-table) case.
- Full agent suite: `cd scripts/agent && node --test *.test.mjs` → **386/386 pass**.

Follow-up (task file): once real rounds carry attribution, decide cost levers (samples 2→1 on cheap lenses, lower absence-verifier `maxTurns`, cap verifier calls) from the table rather than a guess.

## Summary by CodeRabbit

* **New Features**
  * Added token attribution to review summaries, broken down by lens and detection versus verifier activity.
  * Added cost, weighted-token, turn, and call metrics for attributed review activity.
  * Preserved compatibility with earlier records by handling unattributed activity separately.

* **Documentation**
  * Documented token attribution behavior, measurement scope, and backward-compatible handling.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -597,7 +597,15 @@ Components:
   `lensStats.lanes` reports `blocking`/`backlog`/`unknownOrigin` and
   `scripts/agent/metrics.mjs` renders it; read `unknownOrigin` first, since a zero `backlog`
   beside a high `unknownOrigin` means the gate ran blind rather than that nothing
-  was relocated. The
+  was relocated.
+  **Token attribution.** Every SDK call is stamped with `{ lens, role }`
+  (`role` ∈ `detection` | `verifier`) as `scripts/agent/ask.mjs` pushes its result
+  to the shared execution log, so `scripts/agent/metrics.mjs`'s
+  `attributionBreakdown` can split the panel's spend per lens and by
+  detection-vs-verifier instead of one anonymous total — the effort summary renders
+  it as a cost · weighted-tokens table. This is what makes "which
+  lens, and is it the verifier?" answerable from the ledger rather than by guessing;
+  it is measurement only and gates nothing. The
   **trusted orchestrator** (run from a `main` checkout, via the shared
   `scripts/agent/severity.mjs` rule) computes each lens's conclusion — the
   subagents only classify — and the job records one unforgeable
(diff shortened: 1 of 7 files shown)


harrykim8672 (dismissed):
Thanks for your contribution

harrykim8672 (approved):
.

PR #607, opened 2026-07-31
Run every review lens at a single detection sample

hwisoo-kim:
## Summary

The review panel exhausts a session limit on large PRs. On **#605** (L-size) the panel cost **~$14 / 428 turns / 73 min for a single round**, its verifier sessions hit session-limit **429s (6 of 10 errored → findings unfiltered)**, the docs lens blew its turn ceiling, and the loop **paged** (`agent:blocked`).

The per-lens **token attribution** (shipped in #603) pinpointed the sink — **detection, not the verifier**:

```
Detection vs verifier: $11.60 vs $2.74
correctness  $3.81 · ~690K   |  blast-radius $2.97 · ~609K  |  design-fit $2.29 · ~490K
```

Each lens ran **two opus detection samples**. This drops **every** blocking lens to one.

## Change

Data-only manifest edit (`scripts/agent/lenses/lenses.json`) + a design-doc note. All blocking lenses go `samples: 2 → 1` (`docs` was already 1): `correctness`, `security`, `design-fit`, `test-adequacy`, `blast-radius`.

## The recall tradeoff (accepted)

`samples: 2` + union raises recall against single-sample non-determinism (the #521 false negative). Dropping to 1 gives up that recovery — a defect a single sample misses is no longer caught by a second.

- **Code-quality lenses:** a miss is still backstopped by the **verifier**, the **tests**, and **CI**.
- **`security`:** has **no** such backstop — a planted instruction or pasted secret missed on the one sample is missed. This is the sharpest edge of the trade; it was initially kept at 2 for that reason and then **dropped to 1 on request**, a deliberate, documented acceptance of the risk.

`samples` is a per-lens field, so raising `security` (or any lens) back to 2 is a one-field change the moment the attribution table or a missed finding says it's worth the spend. Secondary benefit: fewer raw findings per round → less verifier load, which is what was 429ing.

## Not in this PR
- The docs lens `error_max_turns` page (the proximate #605 trigger) is a separate fix.

## Test plan
- `sampleCountFor` honors `samples: 1` and is unit-tested against synthetic lenses, so this is a data-only change with no code/test churn.
- Full agent suite: `cd scripts/agent && node --test *.test.mjs` → **388/388 pass**.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -634,6 +634,16 @@ Components:
        highest severity; distinct bugs never merge), and the verifier refute pass
        is the precision counterweight. No LLM/semantic merge — that could
        over-merge two distinct bugs into one, reintroducing the miss.
+       The code-quality lenses (`correctness`, `design-fit`, `test-adequacy`,
+       `blast-radius`) run at `samples: 1`. A second opus detection sample was the
+       panel's single largest cost — detection dominated one L-size PR's ~$14
+       panel (#605, surfaced by the per-lens token attribution above) — while a
+       missed code defect is already backstopped by the verifier, the tests, and
+       CI. `security` deliberately keeps `samples: 2`: a planted instruction or a
+       pasted secret it non-deterministically misses has no other backstop, and it
+       is one of the cheaper lenses, so the recall insurance is worth the marginal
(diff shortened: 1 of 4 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #525, opened 2026-07-23
Harden the review panel against false negatives (sampling + cross-round re-check)

harrykim8672:
## Summary

Harden the review panel against **false negatives** — a lens missing a real
finding on a given run. Motivated by a live case (#521): the correctness lens
flagged a real regression, then reviewed the **same unchanged code clean** on a
re-run. The verifier refute pass can't help here — it only *drops* findings
(fights false *positives*); it can't add back a missed one.

Two complementary measures:

### Part 1 — sample each lens, union findings
Each lens now runs `samples` times (per-lens field in
`scripts/agent/lenses/lenses.json`, default 2) and the findings are **unioned**.
Any sample's finding enters the gate. The union goes through the existing
conservative `coerceFindings`/`dedupeFindings` (identical file+summary collapse
to the highest severity; distinct bugs never merge), and the existing verifier
refute pass is the precision counterweight. If **all** samples of a lens error,
it still fails closed. No LLM/semantic merge — that could over-merge two distinct
bugs into one, reintroducing the miss.

### Part 2 — re-check prior findings across rounds
Each round persists its blocking findings in the per-lens check run's
`output.text`. The next round reads the latest prior `agent-review-<lens>`
findings (`--prior-findings`) and re-verifies each against the **current** diff
with the same **biased-to-keep** refute pass: a prior finding survives unless it
is confidently *resolved*, so it can't vanish just because a later fresh pass
missed it. Unresolved priors merge (deduped) into the round's findings.
Persistence rides on the check-run output — **no new PR comment, no new
permission** (the review job keeps `issues: read`).

## Design notes
- Reuses `dedupeFindings`/`coerceFindings`/`verifyFinding`/`applyVerifications`
  unchanged; `review-panel.mjs` still never touches GitHub (the workflow feeds it
  the prior findings as a file).
- New exported helpers `unionSamples` / `parsePriorFindings`, unit-tested.

## Validation
`node --check`; `lenses.json` + workflow YAML parse; **16/16** agent unit tests
(3 new: `unionSamples`, `parsePriorFindings`, and a composition test guarding the
#521 miss — an unresolved prior finding a fresh pass "missed" still blocks);
dead-code + doc-staleness clean. The SDK sampling/re-check paths run only in real
CI; the pure merge/parse logic is unit-tested.

## Caveat (documented in `harness-engineering.md`)
Both measures **lower** false-negative odds; neither makes the panel safe to
self-promote. The required human review remains the backstop — #521 is the proof.

[...]

PR #608, opened 2026-07-31
Record when the review panel was wrong

hwisoo-kim:
## Summary

Records **when the review panel was wrong**, so the panel's tuning decisions can be
scored instead of argued. Script-only; consumed by nothing yet.

## The gap

The panel has been tuned repeatedly — severity thresholds, `samples: 2`, file-class
routing (#582), the novelty gate (#583), verifier unanimity, restatement clustering
(#591, #601) — and **not one of those decisions can be scored.**

Everything the pipeline records measures effort (`metrics.mjs`), self-agreement
(`compareSampleAgreement`), or process (`rounds.mjs`). Nothing records **outcomes**.
So each change was argued from intuition and two or three remembered incidents — and
the audit's own bar ("0 proven false negatives", "`critical` demonstrably in use", "≤ 2
rounds to converge") is unmeasurable for exactly the same reason.

## What lands

**`scripts/agent/misses.jsonl`** — strict JSONL, stable field order:

```
schema, id, label, source, pr, handoffAt, evidence{commitSha,commentId,url},
files[], fileClasses[], lens, severity, origin, summary,
panelSaw{reviewedSha,conclusion,blockingFindings}, verifiedBy, notes
```

`label` is `miss` **or** `false-positive`, both in one corpus deliberately: a change
that cuts misses by raising more findings has to pay for it in false positives, and a
corpus holding only one of the two would score that change as pure progress.

`fileClasses` and `origin` are additions to the planned shape, now that #582 and #583
ship the classifiers. They turn "we missed four bugs" into a claim about **which**
population the panel is weak on.

**`scripts/agent/harvest.mjs`** — two signatures, both from data GitHub already keeps:

1. **Human commits after handoff.** A human editing reviewable code after the panel
   approved is the shape of a missed defect.
2. **CodeRabbit findings the panel did not raise**, at blocking severity only — this
   corpus measures the *gate*.

## `verifiedBy: ""` is the load-bearing field

The harvester **proposes**; a human confirms. Nothing may consume a record until
someone has put their name there, and the harvester cannot set it even by accident
(`toMissRecord` defaults it, neither harvest path passes it, and a test asserts it).

An auto-harvested, auto-trusted corpus is a corpus of noise — and harvester noise is
**systematic**, since it follows whatever the matcher over-fires on. It would move the
panel somewhere specific and wrong rather than nowhere.

## Corrected from the plan: the CodeRabbit vocabulary was wrong for this repo

The plan specified keeping `Potential issue` / `Refactor suggestion` and dropping
`Nitpick`. **Nothing in this repository has ever emitted those strings.** Every inline
CodeRabbit finding back to #525 uses a three-field italic header:

```
_🎯 Functional Correctness_ | _🟠 Major_ | _⚡ Quick win_
```

A classifier keyed on the upstream labels would have matched zero comments and
reported an empty corpus as a clean bill of health — **a failure that reads as good
news**, which is the worst possible failure mode for this particular component.

Only *Functional Correctness* → `correctness` and *Security & Privacy* → `security` get
an automatic lens. Stability, Performance, Maintainability and Data Integrity each
plausibly map to two of our lenses or to none, and a guessed mapping would corrupt the
per-lens miss count the corpus exists to produce.

## Reuse the classifier, not the glob list

`interestingFiles` routes through `classifyFile` (#582) rather than the planned
hand-written `docs/**` / `*.md` / `.github/**` drop-list — which would have been the
*third* copy of a path table in `scripts/agent/`, and the second copy is how the first
rots.

It also forced a distinction the glob list hid: **`policy` is excluded on purpose.** A
human editing `.github/**` or `lenses/*.md` after handoff is changing **the reviewer**,
a different event from the reviewer missing a bug. One list cannot hold both
populations without corrupting the number.

## Opposite fail directions

| path | on doubt | why |
|---|---|---|
| every read | fewer candidates, never throws | a GitHub hiccup costs this run's proposals, not the corpus |
| `--append` (the only write) | **refuses** | one unparseable line means we don't know every id in the file, so appending could duplicate a curated record — and a duplicate double-counts in every later tally, silently and forever |

`dedupeById` keeps the **first** occurrence for the same reason: existing records are
passed ahead of fresh candidates, so a re-harvest can never blank a human's
`verifiedBy`.

## Three bugs the tests found, all failing toward "smaller corpus, no complaint"

- **`candidateId` coerced with `typeof === "string"`.** GitHub returns comment ids as
  **numbers**, so every CodeRabbit candidate on a PR got the identical id
  (`coderabbit:548:`) — and `dedupeById` would then have kept one, silently discarding
  the rest.
- **`parseJsonl` accepted a bare `7` as a record.** Valid JSON, not a record. Counting
  it as *unreadable* makes it visible and makes `--append` refuse.
- **`--append` silently did nothing.** The shared `parseArgs` has no boolean flags, so
  it read the next token as the value, got `undefined`, took the print-only branch, and
  **exited 0 reporting success.** The module's only write path was unreachable and said
  so in the affirmative.

`parseArgs` now takes a **declared** `booleans` list. Inference ("boolean if the next
token starts with `--`") also fails at end-of-argv, and `review-panel.mjs` deliberately
passes possibly-**empty** `--since-sha` / `--review-mode`, which must stay values.
Callers that pass no `booleans` are bit-for-bit unaffected.

## The seed data already answers a live question

#548's `test-adequacy` miss is proven **by construction**, not by judgement:

| sha | conclusion | blocking |
|---|---|---|
| `ab952c9` | success | 0 major |
| `ffeb1d2` | failure | 1 major |

The only difference between those commits is **+16 lines in `docs/design/sharing.md`**.
No test file changed. No source file changed. The same lens, reading byte-identical
test content, found a real and specific defect (a vacuous paste test that passes
whether or not the read-only guard exists) on one run and missed it on the other — and
the `success` came **first**, so a one-sample panel would have approved and stopped.

That is the strongest evidence yet against dropping to `samples: 1`.

## The two signatures corroborated each other

The seeds were hand-built from evidence. Then the harvester ran over #548 from the
completely independent human-commit signature and proposed:

- *"Guard read-only mutations at the EditorAPI boundary"* → the fix for seed 1
- *"Add real-payload paste test + cut/select-all/table-menu coverage"* → the fix for seed 2

Two signatures sharing no input agreeing on the same two defects is stronger evidence
the harvester works than its own test suite. The third candidate ("restore test shims,
add copy coverage") is probably **not** a panel miss — that's the honest
signal-to-noise ratio, and it was left in rather than quietly curated out.

## Never feed this to a lens

Not as few-shot examples, not as "past misses to watch for". Three independent reasons,
any one sufficient: it biases lenses toward historical bug shapes when the next defect
is by definition new; it re-grows the prompt incremental review exists to shrink; and
it is a verbatim archive of **attacker-influenceable text** (CodeRabbit bodies,
contributor commit messages). Stated at the top of the module and in the design doc.

## Verification

`pnpm verify:self` green (**11/11**) · **425 agent tests** (388 on `main`).

- **End-to-end against the real API** on #548: 3 candidates, exit 0.
- **`--append` run, then re-run** — the second run appends nothing.
- **The append refusal mutation-tested**: a hand-truncated line produced
  `refusing to append — 1 unreadable line(s)`, named the line number, and wrote nothing.
- **The corpus file has its own test** — every line parses, every record carries the
  exact field order, a known label/source, a PR number, a summary, an evidence URL, and
  no duplicate ids. An eval set that silently fails to parse means every later
  measurement is quietly taken against a smaller corpus than people believe they have.
- **`mark-ready.mjs` and `spec-to-pr.mjs` still load** after `HANDOFF_MARKER` moved into
  `disclosure.mjs` (it is now a contract between two modules, and `mark-ready.mjs` runs
  its CLI at import time so nothing can import a constant from it).

**The corpus is five records, three of them unverified candidates and none of them
signed.** It measures nothing yet, and it will not until someone curates. That is the
design — but it means the value of this PR is entirely deferred, and a corpus nobody
curates is worse than none, because its emptiness will read as evidence.

**`samples: 2` is not settled by one record either.** The flip proves single-sample
non-determinism exists in the dangerous direction; it does not measure how often. That
needs `unionSamples` to tag sample provenance, which is not in this PR.

## Linked Issues

## Author checklist

## Verification checklist

## Risk Assessment

- **Data/security risk:** no permission or trust-model change. The one behavioral
  change outside new files is `parseArgs` gaining an opt-in `booleans` parameter;
  callers that omit it parse identically, which is asserted by a test, and it matters
  because that parser is on the path that narrows the reviewed diff.
- **Rollback:** revert. Nothing reads `misses.jsonl`.

## Notes for Reviewers

  and reviewed by me.
- **What needs a human next:** put a handle in `verifiedBy` on the records that hold
  up, and delete the ones that don't. Two are mechanically proven; one is likely noise.
- **`commitCheckRuns`** was split out of `gh-checks.mjs` so reading one commit's verdict
  no longer costs one API call per commit on the PR. `prCommitsWithCheckRuns` now
  delegates to it, and the two keep **opposite** catch behavior on purpose — pinned by a
  test, because the harvester needs "could not look" to be distinguishable from "the
  panel never ran".

## Summary by CodeRabbit

* **New Features**
  * Added tooling to collect, normalize, validate, and deduplicate review-panel misses and false positives in JSONL format.
  * Added feedback harvesting from pull requests, human follow-up changes, and automated review findings.
  * Added five historical feedback records for calibration.
  * Improved command-line flag handling and check-run retrieval.

* **Documentation**
  * Documented corpus design, harvesting workflows, validation rules, security boundaries, and remaining curation tasks.

* **Tests**
  * Added comprehensive coverage for harvesting, parsing, validation, deduplication, filtering, and command-line behavior.


hackerwins:
@coderabbitai review this.

harrykim8672 (approved):
Thanks for your contribution

PR #610, opened 2026-07-31
Drop the security review lens to one sample too

hwisoo-kim:
## Summary

Follow-up to **#607** (merged), which cut the code-quality lenses to `samples: 1` but **kept `security` at 2**. This drops `security` to 1 as well, on request — so every blocking review lens is now single-sample.

## Change

One field in `scripts/agent/lenses/lenses.json` (`security`: `2 → 1`) plus the design-doc note and task files updated to match. Final state:

| Lens | samples |
|---|---|
| correctness, security, design-fit, test-adequacy, blast-radius | **1** |
| docs | 1 (already) |

## The tradeoff (accepted, and reversible)

`security` was the one lens #607 deliberately left at 2, because `samples: 2` + union is the only recall guard against single-sample non-determinism there — a planted instruction or pasted secret it misses on one sample has **no** verifier/test/CI backstop the way a code defect does. Dropping to 1 accepts that edge for the cost of a panel that fits within one session limit.

`samples` is a per-lens field, so raising `security` back to 2 is a one-line change the moment the token-attribution table or a missed finding says it's worth the spend.

## Test plan

- Data-only manifest change; `sampleCountFor` honors `samples: 1` and is unit-tested against synthetic lenses, so no code/test churn.
- Full agent suite: `cd scripts/agent && node --test *.test.mjs` → **388/388 pass**.

## Summary by CodeRabbit

* **Documentation**
  * Updated review-lens guidance to use one sample for all lenses.
  * Documented the associated cost and detection-recall tradeoffs.
  * Recorded the accepted, reversible reduction for security reviews and criteria for future increases.

* **Configuration**
  * Reduced security review sampling from two samples to one.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -634,16 +634,19 @@ Components:
        highest severity; distinct bugs never merge), and the verifier refute pass
        is the precision counterweight. No LLM/semantic merge — that could
        over-merge two distinct bugs into one, reintroducing the miss.
-       The code-quality lenses (`correctness`, `design-fit`, `test-adequacy`,
-       `blast-radius`) run at `samples: 1`. A second opus detection sample was the
-       panel's single largest cost — detection dominated one L-size PR's ~$14
-       panel (#605, surfaced by the per-lens token attribution above) — while a
-       missed code defect is already backstopped by the verifier, the tests, and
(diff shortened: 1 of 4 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #615, opened 2026-07-31
Cap the review fix loop at three rounds

hwisoo-kim:
## Summary

`MAX_REVIEW_ROUNDS: "5"` → `"3"`, with the reasoning recorded in the workflow env
block and in `docs/design/harness-engineering.md`.

## Why

Nothing in the pipeline can page for non-convergence before round 3.
`detectStalledRounds` needs `minRepeats + 1` = 3 rounds of evidence before it will
report a stall, and `review-round-guard.mjs` checks it *before* the round cap so the
more specific reason wins. Rounds 4 and 5 were therefore a PR paying a full panel
round each — **~$12 at the current measured baseline** — on the way to a page that was
already inevitable. A PR the fixer has failed to converge three times is not going to
converge on the fourth without the human the guard is about to ask for.

This is not a hypothetical bound. `harness-engineering.md` records **PR #521 burning
all five rounds** re-litigating overlapping findings before anyone was paged — which
is what motivated the convergence detector in the first place. The detector shortened
the tail; this trims what is left behind it.

## What this does NOT do

Worth stating plainly, because the constant's name invites over-reading it, and it is
now called out in both the workflow comment and the design doc:

**This bounds the fixer, not panel spend.** The guard runs in the `fix` job,
downstream of `review-panel`, so a PR that has already been paged still re-runs the
whole panel on every CI-green push. #605 is the worked example:

| time | event |
|---|---|
| 07-30 15:57:26 | **PAGED** — "a review lens did not produce a valid verdict, a human should review" |
| 07-30 15:57:27 | round 1 — $14.33 |
| 07-30 16:02:09 | round 2 — $10.05 |
| 07-31 00:06:19 | round 3 — $18.25 |
| 07-31 07:16:17 | round 4 — $11.71 |

Four rounds, **$54.35**, of which **$40.02 landed after the pipeline had already paged
a human at round 1**. None of it would have been prevented by this change: the fixer
was latched off by the `agent-review-paged` check at the top of the guard long before
the round cap could be reached. That is a separate hole and I'll raise it separately
rather than widen this PR into a gate-behaviour change.

## Linked Issues

## Author checklist

## Verification

Locally: the workflow parses (`yaml.safe_load`, `env.MAX_REVIEW_ROUNDS == '3'` — still
a string, as `review-round-guard.mjs` expects), and `cd scripts/agent && node --test
*.test.mjs` is **0 fail** (1 skip, pre-existing on `main`).

and covers neither `.github/workflows/**` nor `docs/**`. `verify:entropy` (which does
check that backticked paths in `docs/design/*.md` resolve) runs in CI; the doc edit
adds no new file references.

## Risk Assessment

- **User-facing risk:** none. Affects only the autonomous pipeline's fix loop.
- **Rollback plan:** revert the single commit; it is one env value plus its rationale.

**The trade:** a PR that genuinely would have converged on round 4 now pages a human
instead. I think that is the right side to err on — a human at round 3 costs less than
two more panel rounds, and the same guard already pages at round 3 for non-convergence
anyway, so this only changes the outcome for PRs that are *still reducing* their
blocking-finding count at round 3 and would have finished at 4. If that turns out to
be common, the constant is one character back the other way.

## Notes for Reviewers

**AI assistance:** this PR was written with Claude Code — the cost analysis, the diff,
and the doc edit.

I deliberately did **not** also set `STALL_REPEATS`. It is env-driven and unset
(defaulting to 2), and dropping it to 1 would page a round earlier still — but I could
not validate it. Replaying `detectStalledRounds` offline across every agent-managed PR
in the repo found only two with any lens check runs at all (#605 with 2 rounds, both
`partial`, and #575 with 1), and `detectStalledRounds` breaks its run on any partial
round, so the entire available corpus returns `progressing` / `too-few-rounds` at both
`minRepeats: 1` and `2`. There is nothing to validate against, so changing it would be
guessing at when to page a human. Worth revisiting once a few PRs have accumulated
non-partial multi-round histories.

- Follow-up work: this is PR 2 of a series. #613 (fixer no longer receives minor/nit
  findings) and #614 (verifier presence turn ceiling) have merged. Next is threading a
  per-lens `effort` onto the review sessions, which is the largest remaining
  per-round lever.


hackerwins:
@coderabbitai review this PR.

hackerwins (approved):
Thanks for your contribution.

PR #622, opened 2026-08-01
Board SP3: import a Miro board into a new board document

hackerwins:
## Summary

SP3, the final board sub-project ([design](docs/design/board/board-miro-import.md)) — import a Miro board into a new wafflebase `"board"` document as native elements. Three layers:

- **Backend proxy** (`packages/backend/src/miro/`) — workspace-scoped `POST /workspaces/:wid/miro/import`. Parses the board id from a pasted URL, fetches `/items` **and** the separate `/connectors` feed with cursor pagination, and re-hosts image bytes into the workspace image bucket. The user's Miro token is used for that one request and is **never stored, logged, or returned** — no Prisma model, no encryption-at-rest, no connection CRUD.
- **Pure mapper** (`packages/board/src/import/miro/`) — `mapMiroItems({items, connectors, resolveImageUrl}) → { inits, skipped, approximated }`. Two-pass with an id map so connectors resolve regardless of item order. Maps sticky/shape/text/image + best-effort frame/card/app_card; unsupported types are counted, never dropped silently.
- **Frontend** — an "Import from Miro…" dialog on the documents list, and a `board` branch in `applyImportedContent` that persists the elements headlessly in one `YorkieBoardStore.batch`.

Every degradation — unsupported types, downgraded shape kinds, failed images, a truncated or stalled read, a dropped connector — reaches the user; an incomplete import can never show a plain success.

## Test plan

- Suites: backend **373**, board **33**, frontend **940**. `pnpm verify:self` green across all 11 lanes. Chunk gate 144 → 145 (the applier importing `YorkieBoardStore` makes it a second importer, so Vite hoists it into its own shared chunk).
- Built via subagent-driven development: 8 tasks, each independently reviewed (5 needed a fix round), then an Opus whole-branch review.

### The whole-branch review earned its keep

Every layer's tests were green and the layers still didn't compose. Two Criticals, both at a seam, both now fixed **and covered by a cross-layer composition test that fails 4/4 against the pre-fix code**:

1. **Every connector was broken.** The mapper assigned its own element ids and pointed endpoints at them, but `YorkieBoardStore.addElement` mints its *own* id and spreads it last — so every arrow referenced an id absent from the document and collapsed to a ~1×1 frame at the world origin, under a success toast. The applier now remaps endpoints onto the ids the store actually returns.
2. **Every imported image 404'd.** The backend returns a root-relative `/api/v1/...` URL; the native upload path runs it through `resolveImageUrl` because frontend and backend are cross-origin in dev *and* prod. The Miro path skipped that and persisted the relative URL into the CRDT. Now normalized via the same exported helper, injected as a **required** field so a future missing call site is a compile error.

Also hardened in the same wave: an allowlist + streaming size cap on the image re-host (it was the one hop whose URL comes from upstream JSON, and it carried the user's bearer token), orphan-document cleanup when an apply fails, and six smaller correctness fixes.

### Manual smoke (recommended before merge)
`docker compose up -d` + `pnpm dev`, with a real Miro token (`boards:read`) and a board containing a sticky, shape, text, connector, and image: Documents → New/Import → "Import from Miro…" → verify placement matches Miro, the connector joins the same two shapes, the image renders from an `/api/v1/workspaces/...` URL, a bad token keeps the dialog open with values intact, and an unsupported item is reported in the toast.

Known follow-ups (none blocking) are listed in the lessons file.

## Summary by CodeRabbit

- **New Features**
  - Import Miro boards into native Wafflebase board documents.
  - Supports elements, connectors, text, images, shapes, frames, and cards.
  - Added an “Import from Miro…” option with progress, success, and warning feedback.
  - Temporary access tokens are protected, and imported images are securely rehosted.
  - Boards now fit imported content to the viewport automatically.

- **Bug Fixes**
  - Improved board rendering performance and snapshot consistency.
  - Preserved connector relationships and handled unsupported content gracefully.

- **Documentation**
  - Added Miro import design specifications, API guidance, implementation plans, and lessons learned.

- **Tests**
  - Expanded coverage for imports, mapping, security, pagination, images, rendering, and warnings.


hackerwins:
Addressed all 12 findings across 4 commits (`63f77c6f5`, `e31691062`, `a75d363b8`, `103282e66`). Every one was verified against the code first — all were real.

| Finding | Fix |
|---|---|
| No timeout on any outbound fetch (Major) | `AbortSignal.timeout` on both call sites — `API_TIMEOUT_MS=15s` per page, `IMAGE_TIMEOUT_MS=30s`; the image path combines it with the size-cap controller via `AbortSignal.any` so both aborts still work. A timeout now reads "did not respond in time" rather than "could not reach". |
| Total image work unbounded (Major) | Aggregate ceilings `MAX_REHOSTED_IMAGES=100` / `MAX_TOTAL_IMAGE_BYTES=64MB`; the remainder is skipped **un-fetched** and reported as a new `image-budget` note, worded distinctly from `image-failed`. |
| Dangling `idMap` entry for an image without `imageUrl` | Validated in pass 1, before `generateId()`/`idMap.set`, so an unusable image can never be a connector target. It previously degraded safely (the applier's remap guard dropped the connector) but was attributed to the wrong counter; the mapper now reports it as `skipped.connector`. |
| Miro entry offered without a workspace (Major) | Hidden when there's no `workspaceId`, at both dropdown call sites — no more dead-end dialog on `/documents`. |
| Failure rendered during auth expiry | Early return on `isAuthExpiredError`, matching the ~10 other call sites. |
| Stale error on reopen | Error cleared with the dialog. |
| Delimiters accepted after decoding the board id | Re-validated post-decode (`ab%2Fcd` is now rejected). |
| Formatting + 3 doc nits | Prettier over the added files; the split `shape` table row rejoined; completed plan steps checked and the superseded connector-fallback step annotated rather than rewritten. |

Tests: backend **382**, board **34**, frontend **941**; lint clean; `pnpm verify:self` all 11 lanes green. Chunk count unchanged, so no gate bump. New tests were confirmed RED first for the image-budget and dangling-idMap fixes. The token invariant is intact — no new path logs, stores, or returns it, and no new message carries the token or a URL.

One item stays open by necessity: the manual smoke needs a live Miro token, so that step is still unchecked.

PR #630, opened 2026-08-03
Stop running the review panel on a paged PR

hwisoo-kim:
## Summary

The `gate` job now reads the `<!-- agent-review-paged -->` latch and reports the PR as
unmanaged, so a PR the pipeline has already handed to a human stops being re-reviewed.

## Why

`review-round-guard.mjs` reads that latch to stop dispatching the fixer — but the
guard runs in the `fix` job, **downstream of `review-panel`**, and neither `gate`
(`AGENT_PIPELINE_ENABLED` + CI conclusion + same-repo) nor `review-panel`
(`needs.gate.outputs.managed == 'true'`) checked it. So a paged PR kept re-reviewing
on every CI-green push: six lenses plus per-finding verification, ~$12 a round, for
verdicts nothing consumes.

Surveying PRs 540–630 against the metric ledger, counting a round as post-page only if
it posted **more than 120 s** after the latch (the round that *causes* a page posts
about a second after it, from the same job):

| | |
|---|---|
| PRs with review activity | 9 |
| PRs paged | 4 |
| **PRs that kept billing after the page** | **2** |
| **Post-page rounds / cost** | **6 / $58.53** |
| Share of all review spend in range | **44.4%** |

Low frequency, unbounded tail: **$53.98 of that is #605 alone** — paged at round 1,
then five more full rounds over two days ($10.05, $18.25, $11.71, $12.76, $1.21).

### Those rounds are not the price of keeping the promotion path open

They could have been. `promote` is gated only on `blocked == 'false'` and does **not**
check the latch, so in principle a human who fixes everything gets a clean round and
an automatic promotion. In practice nothing has ever used it:

| PR | outcome | labels | post-page |
|---|---|---|---|
| #548 | merged by `hackerwins` | `agent:blocked` | 1 round, $4.55 |
| #564 | merged by `harrykim8672` | `agent:ready` | **0 rounds** — promoted without any |
| #575 | closed | `agent:blocked` | 0 rounds |
| #605 | merged by `hackerwins`, **no label** | — | 5 rounds, **$53.98** |

#605 is the decisive one: it never got `agent:ready`, so those five rounds bought no
promotion. The maintainer asked **CodeRabbit** for a review and merged manually 28
minutes later. Four for four, a paged PR ends with a human.

If a paged PR ever does need to re-enter the gate, an `agent:resume` label read by
`gate` (which already reads labels for `agent:managed`) is a small follow-up. Building
it now would be designing for a path the evidence says nobody walks.

## Design notes

**Folded into `managed` rather than exposed as its own output — load-bearing, not
tidiness.** `stalled` fires on `needs.gate.outputs.managed == 'true' &&
needs.review-panel.result == 'skipped'`, so a separate `paged` flag would make every
paged PR post a **second** page. A deliberate skip belongs in the same category as
unmanaged/fork/disabled — which is exactly what the comment above `stalled` already
says that check is for: *"we do NOT page when the panel was skipped for being an
unmanaged/fork/disabled PR — only when a managed PR's pipeline actually broke."*

**Fails open**, matching the label lookup beside it. Not knowing whether a PR was paged
must cost a review round, never skip one — a missed review is worse than a wasted one.

**`issues: read`** added to `gate`: the latch is an issue comment on the PR.

**The page comments now say the panel has stopped.** The latch freezes the
`agent-review-*` checks, so a page that doesn't say so leaves a human waiting for
verdicts that will never arrive. All three write sites get the line, and a test
asserts each one carries it so a fourth page path can't be added silently.

**`PAGED_LATCH` moved to `rounds.mjs`** so the guard imports it instead of declaring
its own copy. `gate` does no checkout and cannot import, so it carries a literal copy —
and a drifted copy wouldn't error, it would just stop latching, so `rounds.test.mjs`
asserts the two are byte-identical.

## Linked Issues

## Author checklist

## Verification

`cd scripts/agent && node --test *.test.mjs` → **480 pass, 0 fail** (1 skip,
pre-existing on `main`). Workflow YAML parses; the `gate` step's script passes
`node --check`; so does `review-round-guard.mjs`.

The gate step was also driven end-to-end against a stubbed GitHub API — extracted from
the YAML so the *shipped* script runs, not a copy:

| case | `managed` |
|---|---|
| `agent/` branch, not paged | `true` |
| `agent/` branch, **paged** | `false` + notice |
| labelled human PR, **paged** | `false` + notice |
| unmanaged branch | `false` |
| **comment read throws 403** | **`true`** + warning (fails open) |

`pnpm verify:fast` was not run locally: it needs a full install in a fresh worktree and
covers neither `.github/workflows/**` nor `scripts/agent` (outside the pnpm workspace).

## Risk Assessment

- **User-facing risk:** none. Affects only the autonomous pipeline.
- **Data/security risk:** none. Adds one read-only permission (`issues: read`) and
  strictly *reduces* what runs.
- **Rollback plan:** revert the single commit.

**The trade:** a paged PR can no longer be auto-promoted by fixing it and pushing.
On the evidence that path has never been used, and the page comment now states it
plainly. Fail direction is "a human merges it themselves", which is where all four
paged PRs already ended up.

## Notes for Reviewers

**AI assistance:** this PR was written with Claude Code — the survey, the diff, and the
tests.

The one judgement call worth challenging is folding into `managed` instead of adding a
`paged` output. It reads as conflating two things, and I'd normally agree — but a
separate flag makes `stalled` fire on every paged PR, and the alternative (also
teaching `stalled` about the new flag) means two conditions that must stay in sync
for the pipeline not to double-page. One concept, one output, and `core.notice()`
records *why* a PR was skipped in the run summary.

- Follow-up work: this is PR 10 of a series; #613, #614 and #615 have merged. Still
  open in the plan: per-lens reasoning `effort` (the largest remaining per-round
  lever), verifier failure-kind instrumentation, and dropping the duplicate
  verification of carried-forward findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented automated review panels from restarting after a pull request has been handed off for human review.
  * Improved handling of stalled pipelines and unavailable comment checks so review status is clearer and safer.

* **User Experience**
  * Added explicit notices when automated review and merge readiness are paused, indicating that manual review and merging are required.

* **Tests**
  * Added coverage to verify handoff markers and pause messages remain consistent.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -54,6 +54,7 @@ jobs:
     timeout-minutes: 5
     permissions:
       pull-requests: read
+      issues: read # the paged latch is an ISSUE comment on the PR
     outputs:
       managed: ${{ steps.gate.outputs.managed }}
       pr: ${{ steps.gate.outputs.pr }}
@@ -81,6 +82,42 @@ jobs:
               }
             } catch (e) {
               core.warning(`Could not resolve PR labels (${e.message}); using branch-prefix only.`);
+            }
+            // PAGED LATCH — the pipeline has already handed this PR to a human.
+            // review-round-guard.mjs reads the same marker to stop dispatching
+            // the fixer, but it runs in `fix`, DOWNSTREAM of `review-panel`, so
+            // until now a paged PR still re-reviewed on every CI-green push:
+            // six lenses plus verification, ~$12 a round, for verdicts nothing
+            // consumes. On #605 that was five rounds and $53.98 after the page,
+            // and it merged manually without ever using them.
+            //
+            // Folded into `managed` rather than exposed as its own output, and
+            // that is load-bearing: `stalled` fires on
+            // `managed == 'true' && review-panel.result == 'skipped'`, so a
+            // separate flag would make every paged PR post a SECOND page. This
+            // belongs in the same category as unmanaged/fork/disabled — reasons
+            // the panel legitimately does not run — which is exactly what the
+            // comment above `stalled` already says that check is for.
+            //
+            // Literal copy of PAGED_LATCH in scripts/agent/rounds.mjs; this job
+            // does no checkout, so it cannot import. rounds.test.mjs asserts the
+            // two match.
+            const PAGED_LATCH = '<!-- agent-review-paged -->';
+            if (managed && prNumber) {
+              try {
+                const comments = await github.paginate(github.rest.issues.listComments, {
+                  owner, repo, issue_number: Number(prNumber), per_page: 100,
+                });
+                if (comments.some((c) => (c.body ?? '').includes(PAGED_LATCH))) {
+                  core.notice(`PR #${prNumber} was already paged to a human — skipping the review panel.`);
+                  managed = false;
+                }
+              } catch (e) {
+                // Fails OPEN, matching the label lookup above. Not knowing
+                // whether a PR was paged must cost a review round, never skip
+                // one: a missed review is worse than a wasted one.
+                core.warning(`Could not read comments (${e.message}); reviewing as if not paged.`);
+              }
             }
             core.setOutput('pr', prNumber);
             core.setOutput('managed', String(managed));
@@ -1059,6 +1096,8 @@ jobs:
             printf '%s\n' \
               '<!-- agent-review-paged -->' \
               '🛑 The fix agent did not advance the branch head, so the requested changes were not applied and the loop cannot re-trigger itself (no push → no new CI → no new panel). Likely causes: it produced no new commit, did not converge within its turn budget, hit tool-permission denials, could not push, or the branch head could not be determined — see the `claude-fix-execution-output` artifact and the run log. **A human should take over** on this PR.' \
+              '' \
+              'The review panel will not run again on this PR, so the `agent-review-*` checks are now frozen at their current state and the ready gate will not promote it. Review and merge it manually.' \
               | gh pr comment "$PR" --body-file -
             node ./scripts/agent/set-state.mjs "$PR" blocked || true
           fi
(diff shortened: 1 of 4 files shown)


hwisoo-kim:
Verified against current code — the finding is valid, and worse than the comment implies. Fixed in e6fd981cf.

**Confirmed.** `wafflebase/wafflebase` is `visibility=public`, so anyone with a GitHub account can comment on an agent PR. The check matched on the marker alone, so a stranger posting `<!-- agent-review-paged -->` would have permanently stopped the review panel. Denial of review from an unauthenticated position.

**Two things the finding didn't cover, both handled:**

1. **The hole predates this branch.** `review-round-guard.mjs` already read the latch the same way, so a stranger could already stop the *fixer*; this PR widened the blast radius to the whole panel. Fixing only the new gate check would have left that. The rule now lives once in `rounds.mjs` as `isPagedLatchComment`, and both readers use it.

2. **`author_association` alone would not have worked** — the tempting fix. Both bots that actually write the latch report `CONTRIBUTOR`, the same value an arbitrary outside contributor gets:

   | PR | latch author | type | association |
   |---|---|---|---|
   | #548 / #564 / #575 | `yorkie-agent[bot]` | Bot | CONTRIBUTOR |
   | #605 | `github-actions[bot]` | Bot | CONTRIBUTOR |

   So trust is **an allow-listed bot login** (GitHub reserves the `[bot]` suffix for Apps, so neither name can be registered — and `user.type === 'Bot'` is checked too, not the name alone) **or a human with write access** (`OWNER`/`MEMBER`/`COLLABORATOR`), which keeps a maintainer halting the pipeline by hand working. Both bot identities are required: the guard and `stalled` comment as `github-actions[bot]`, while the fix job's branch-head page uses the App token.

**Validation.** Agent lane **483 pass / 0 fail**. New tests cover the untrusted-author case you asked for — outside `NONE`/`CONTRIBUTOR`/`FIRST_TIME_CONTRIBUTOR`, a `User` wearing a `[bot]`-shaped login, and an unrelated App (`coderabbitai[bot]`, a `Bot` but not ours). Plus the shipped gate script extracted from the YAML and driven against a stubbed API:

| case | `managed` |
|---|---|
| real bot `github-actions[bot]` / `yorkie-agent[bot]` | `false` — latched |
| maintainer by hand (`MEMBER`) | `false` — latched |
| **stranger posts marker (`NONE`)** | **`true` — panel still runs** |
| **outside `CONTRIBUTOR`** | **`true` — panel still runs** |
| **unrelated App (`Bot`)** | **`true` — panel still runs** |
| comment read throws 403 | `true` — fails open |

One test of mine also had to be fixed: it counted latch *write sites* by grepping for the marker, and my new rationale comment quotes the marker, so prose lines are now excluded from that count.

harrykim8672 (approved):
Thanks for your contribution

PR #633, opened 2026-08-03
Let the author dispute a finding, and the panel decide

hwisoo-kim:
## Summary

Closes the loop the fixer prompt has been apologising for since #564: the author
can now **dispute** a finding, and the trusted path **decides**.

## The gap

The fixer prompt has always said *"if you believe a finding is wrong, reply in the
PR thread with your reasoning"* — and **nothing consumed that reply**.
`review-panel.mjs` receives the diff, the changed files, the issue spec and the
prior findings. It has never received a word the author wrote.

On **#564** the fixer posted a correct, evidenced rebuttal — the App cannot push
`.github/workflows/**`, with the literal push error — at 06:46. The 07:05 panel
never saw it and re-raised the same finding, twice more.

The prompt is currently honest about the hole:

> A reply does NOT resolve the finding — nothing consumes it, the next round will
> raise it again, and only **an independent adjudication** or a human can clear it.

This is that adjudication.

## The trust rule

| step | who | trusted? |
|---|---|---|
| writes the rebuttal | fixer, via `rebuttal.mjs post` → hidden PR comment | **no** — it's a claim |
| reads it | panel job, `ref: main`, `rebuttal.mjs read` | yes |
| judges it | adjudicator subagent, fresh context, read-only tools | n/a — advisory |
| **decides** | `isOverturningVerdict` in the trusted script | **yes** |

`isOverturningVerdict` is the deliberate mirror of `isDroppingVerdict`: an explicit
`overturned`, `high` confidence, an enumerated ground that isn't `none`, and at
least one `file.ext:line` the adjudicator actually read.

## Persuasion must never be a bypass

Structural in four places rather than argued once:

| where | on doubt | |
|---|---|---|
| `parseRebuttalComment` | ignored | a half-understood rebuttal is worse than none |
| `matchRebuttal` | **refuses a tie** | text fitting two findings equally well names neither |
| errored session | upholds | an error argues nothing |
| `isOverturningVerdict` | upholds | needs a ground *and* a location |

**A rebuttal can only ever lose.**

The ambiguity rule is the least obvious and the most important. Without it, an
author who writes text matching several findings in one file gets them adjudicated
together — one argument buying several removals. Refusing ties costs a legitimate
rebuttal nothing (re-word it) and closes the whole shape.

## Undeliverable is not wrong

**There is no overturn ground for "I cannot make this change"** — and #564 is
exactly that case. The claim was *true*; the finding was still correct and still
needed doing, by a human. A ground for inability would let a PR merge with the work
declared impossible, which is worse than the deadlock it replaces. Such a rebuttal
is upheld, re-raised, disputed again, and pages at two.

`out-of-scope` is absent too — scope is an argument, not a fact about the code, and
it's the most persuadable ground a model could be handed. So is `pre-existing`:
provenance is the novelty gate's job since #583, answered with git.

## Bounded

`MAX_REBUTTAL_ROUNDS = 2`, checked in `review-round-guard.mjs` **before**
convergence — the more specific reason wins, same ordering rationale as the infra
branch above it.

The count rides on the finding through the check run's `output.text`, the
**unforgeable** channel — never the author's own comments. Paging is the safe
direction, so a forged count only summons an unnecessary human; but a bound the
disputing party can move is not a bound.

It also fires on a shape the plan didn't name: overturned in round 1, re-raised in
round 2 (lenses are stateless), disputed again. Both mean the loop can't settle it
alone, so the counter doesn't distinguish them.

## The bug only the end-to-end run found

`groupReviewRounds` projects every carried finding down to
`{lens, severity, file, summary}`. The count was written by the panel and **dropped
before the guard read it** — and a dropped count is indistinguishable from "nobody
ever disputed anything", so the page would simply never have fired.

Both halves' unit tests passed. It surfaced only by assembling the real shape:
serialize what the panel writes → run the workflow's own `.map()` extracted from
the YAML → feed it through `groupReviewRounds` → call the guard's predicate. Now a
regression test.

## Inert by default

`--rebuttals` absent → empty list → short-circuits before any session opens. A
panel invoked without it is the panel that existed before this.

## Verification

`pnpm verify:self` green (**11/11**) · **507 agent tests** (478 on `main`) · both
touched workflows parse.

- **The adjudicator is injectable**, so the *wiring* is pinned and not just the
  judgement: overturn drops, every weaker verdict upholds and counts, an errored
  session upholds, an ambiguous match **opens no session at all**.
- **The prompt's fence is asserted by construction** — an injection string provably
  sits inside `<author-rebuttal>`, and `UPHOLD unless` provably precedes it, so the
  model meets the argument already knowing that merely-plausible loses.
- **The carry-forward map was extracted from the YAML and executed**: the integer
  survives, the adjudicator's prose is stripped (it would spend the 60k budget on
  text no consumer reads), folded `mergedFrom` counts survive, an undisputed
  finding carries no key.
- `rebuttal.mjs post` round-trips its own output before posting — a record this
  module can't read back is one the panel will ignore, and the author would never
  learn the dispute went nowhere.

**This adds a model call to the argument path.** A disputed finding now costs a
session on top of detection and verification, and the thing being adjudicated is
the finding *most likely* to be re-disputed next round. The bound caps it at two,
but the honest read is that arguing is now billable and nobody has measured how
often the author is right.

**Nothing yet measures whether an overturn was correct.** An overturn that
shouldn't have happened is a false negative like any other and belongs in
`misses.jsonl` (#608) — but that link isn't built, so the precision of this
mechanism is currently unobserved.

## Linked Issues

## Author checklist

## Verification checklist

## Risk Assessment

- **Data/security risk:** this adds an **author-controlled string to a model
  prompt**, which is the change worth the most scrutiny here. It is fenced as DATA
  with the uphold-default stated before it, the adjudicator has read-only tools,
  and no verdict it returns can do anything except *remove* a finding the trusted
  script has already agreed is removable. The verifier's path — the one prompt in
  the panel that has never carried author text — is untouched.
- **Rollback:** revert. Without `--rebuttals` the panel is unchanged.

## Notes for Reviewers

  and reviewed by me.
- **Why `adjudicateFinding` is a sibling of `verifyFinding` and not a mode of it:**
  they ask different questions of different evidence, and only one is handed
  author-written text. Folding them together would put an untrusted string on the
  verifier's path.
- **`ADJUDICATOR_MAX_TURNS = 20`**, matching the presence verifier's post-#614
  ceiling rather than the 8 it replaced — every one of #605's eight failures came in
  at exactly 9 turns.

## Summary by CodeRabbit

- **New Features**
  - Added structured rebuttals for disputing review findings with supporting evidence.
  - Added independent adjudication to overturn only well-supported findings.
  - Preserved disputed findings when rebuttals are invalid, ambiguous, or cannot be verified.
  - Added tracking for repeated upheld disputes and escalation to human review.
  - Updated review results and summaries to reflect adjudicated findings.

- **Documentation**
  - Added guidance and implementation notes for structured rebuttals, adjudication, safeguards, and operational lessons.

- **Tests**
  - Added coverage for rebuttal validation, matching, adjudication, persistence, safeguards, and failure handling.

### DIFF
--- a/.github/workflows/agent-review-on-demand.yml
+++ b/.github/workflows/agent-review-on-demand.yml
@@ -300,6 +300,17 @@ jobs:
           --lenses .trusted/scripts/agent/lenses/lenses.json
           --out /tmp/prior-findings.json
 
+      # Same untrusted side-channel the gating panel reads, for the same reason:
+      # an on-demand review that re-raised a finding an adjudicator has already
+      # overturned would contradict the gating panel on the same PR, and a human
+      # reading the advisory comment has no way to tell which one is current.
+      - name: Read author rebuttals (align with the gating panel)
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR: ${{ needs.authorize.outputs.pr }}
+        run: node .trusted/scripts/agent/rebuttal.mjs read "$PR" --out /tmp/rebuttals.json
+
       - name: Run review panel
         id: panel
         # A crash must not fail the job — the aggregate step below then posts a
@@ -310,14 +321,16 @@ jobs:
           # Via env, never interpolated into the `run:` string — this step holds
           # the Claude token. Same rule as the twin step in agent-review-panel.yml.
           BASE_SHA: ${{ steps.diff.outputs.base }}
-        # --prior-findings is tolerant of a missing file (existsSync guard in
-        # review-panel.mjs → treated as []), so a skipped/failed read step is safe.
+        # --prior-findings and --rebuttals are both tolerant of a missing file
+        # (existsSync guard in review-panel.mjs → treated as []), so a
+        # skipped/failed read step above is safe.
         run: >-
           node .trusted/scripts/agent/review-panel.mjs
           --diff-file /tmp/pr.diff
           --issue-file /tmp/issue.txt
           --changed-files /tmp/changed.txt
           --prior-findings /tmp/prior-findings.json
+          --rebuttals /tmp/rebuttals.json
           --base-sha "$BASE_SHA"
           --repo "$GITHUB_WORKSPACE"
           --lenses-dir .trusted/scripts/agent/lenses
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -410,6 +410,23 @@ jobs:
           --lenses .trusted/scripts/agent/lenses/lenses.json
           --out /tmp/prior-findings.json
 
+      # The author's structured rebuttals of earlier findings, as hidden PR
+      # comments. UNTRUSTED DATA: the fixer writes these, and everything
+      # downstream treats them as a claim to be grounded rather than as evidence.
+      # The TRUSTED copy of the reader runs here (.trusted, not the branch), so a
+      # branch cannot change what counts as a rebuttal — only what one says.
+      #
+      # continue-on-error plus a script that returns [] on any failure: an
+      # unreadable side-channel must degrade to "nobody disputed anything", which
+      # is exactly the pre-#608 behaviour, never to a failed panel.
+      - name: Read author rebuttals
+        if: steps.pr.outputs.number != ''
+        continue-on-error: true
+        env:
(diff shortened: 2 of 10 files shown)


harrykim8672:
@coderabbitai full review

PR #636, opened 2026-08-03
Set per-lens reasoning effort on review panel sessions

hwisoo-kim:
## Summary

Adds an optional per-lens `effort` to the manifest and threads it onto the review
session. Set to `medium` on five lenses; `security` and the verifier keep the SDK
default.

## Why

`buildSessionOptions` never set `effort`, so **every review session has been running
at the SDK default `high`**. Effort governs how much the model thinks *and* how many
tool calls it makes getting there — and panel spend tracks agentic turns almost
exactly. The measured baseline is **$11.71 over 409 turns, ~$0.032 a turn**, so turns
are the bill and effort is the untouched dial that moves them.

Both Anthropic's `/code-review` docs and the Opus 5 guidance say the same thing for
this workload specifically: at `low`/`medium` the review reports the findings it is
most confident in, so it cuts cost *and* the minor/nit churn that drives extra rounds.

## What each session now gets

| lens | model | effort |
|---|---|---|
| correctness | opus-5 | `medium` |
| **security** | opus-5 | **(absent → SDK default)** |
| design-fit | opus-5 | `medium` |
| test-adequacy | opus-5 | `medium` |
| blast-radius | opus-5 | `medium` |
| docs | sonnet-5 | `medium` |
| **verifier** | per-lens | **(absent → SDK default)** |

**Why security is excluded.** Every other lens has a backstop for a missed finding —
the verifier re-checks blocking findings, a test-adequacy gap is re-derivable from a
failing suite, docs is prose. At `samples: 1` a security miss has none. It is the one
lens where buying turns back isn't worth the recall edge, and it costs ~1/6 of the
saving to leave it alone.

**Why the verifier is excluded.** Lowering effort there looks like the obvious safe
place to save and is backwards: a weaker verifier refutes *less* — it must still reach
`refuted` + `high` confidence + a named `refutationGround` + a real `file:line` in
`groundedIn` to drop anything — so more findings survive to the fixer and the round
gets *more* expensive. A test pins this so a later "set effort everywhere" sweep has
to argue with it.

## Design notes

**Threaded exactly like `maxTurns`.** Optional manifest field → `loadLenses` →
`runLens` → `askStructured` → conditional spread. An unset lens keeps taking the SDK
default rather than being pinned to `high` here, so the field stays inert until set.

**It cannot fragment the cache.** `effort` is a session option and never becomes
prompt text, so it can't split the cross-lens warm-up prefix. #611 made that
structural by dropping the lens argument from `lensCacheKey`; this pins the property
one level up at `buildLensSystemPrompt`, which is where a future "tell the model its
effort level" edit would land.

**`assertEffort` is strict** — `null`, `"MEDIUM"`, `" medium"`, `"hgih"` and wrong
types all throw rather than being dropped. A dropped value leaves the session at
`high` with no error, no changed output and nothing in the logs: the same
silent-cost-regression shape `assertBoundaryMatchesSdk` exists to catch.

**Validated before a token is spent.** `main()` checks the whole manifest between
`loadLenses` and the first session. Inside the per-lens loop instead, a typo would
land in the fail-closed catch and be reported as a blocking *"reviewer did not produce
a valid verdict"* finding — after the lenses ahead of it had already been paid for.

## Linked Issues

## Author checklist

## Verification

`cd scripts/agent && node --test *.test.mjs` → **491 pass, 0 fail** (1 skip,
pre-existing on `main`). SDK checked: pinned `0.3.217`, `effort?: EffortLevel` at
`sdk.d.ts:1637`, levels at `:539` — matching the literal `EFFORT_LEVELS`.

The table above is not hand-written — it's printed by driving the real manifest
through the same `buildSessionOptions` call `runLens` makes.

and `scripts/agent` sits outside the pnpm workspace, so it is covered only by the
`agent:tests` lane inside `verify:self`.

## Risk Assessment

- **User-facing risk:** none. Affects only the autonomous review panel.
- **Rollback plan:** delete the `"effort"` line from any lens in `lenses.json` — one
  field, per lens, same idiom as `samples`. The plumbing is inert without it.

**This is the largest recall risk in the cost series, and the mitigation is the
granularity.** Shallower detection can miss findings. If the ledger shows
`Findings raised` falling while `Rounds` stays flat, revert the affected lens with a
one-line edit rather than the whole change.

## Notes for Reviewers

**AI assistance:** this PR was written with Claude Code — the cost analysis, the diff,
and the tests.

Two things worth a second look:

1. **`medium` is a starting point, not a tuned value.** There is no per-lens effort
   data to fit — nothing has ever run at anything but `high`. `medium` is one step
   down, on five lenses, with the riskiest excluded; the next move should come from
   the ledger, not from another guess.

2. **This PR cannot dogfood itself.** It goes out from a fork, and the panel's `gate`
   job requires `head_repository == github.repository`, so the review panel will not
   run on it — the one change in the series where watching it review itself would have
   been the most direct signal. If you want that, it needs to be re-opened from a
   base-repo branch.

- Follow-up: #613, #614, #615 and #630 have merged. Still open in the plan: verifier
  failure-kind instrumentation, dropping the duplicate verification of carried-forward
  findings, and the CI-coverage note.

## Summary by CodeRabbit

* **New Features**
  * Added configurable AI review effort levels: low, medium, high, xhigh, and max.
  * Review configurations can now select an effort level, while unspecified settings retain the default behavior.
  * Invalid effort values are reported before a review begins.

* **Bug Fixes**
  * Improved validation and error handling for review configuration settings.

* **Tests**
  * Added coverage for supported, omitted, and invalid effort values across review workflows.

### DIFF
--- a/scripts/agent/ask.mjs
+++ b/scripts/agent/ask.mjs
@@ -84,6 +84,42 @@ export const PERMITTED_TOOLS = Object.freeze(["Read", "Grep", "Glob"]);
 export const PERMITTED_MCP_TOOLS = Object.freeze(["mcp__wafflebase__run"]);
 
 const PERMITTED_SET = new Set(PERMITTED_TOOLS);
+
+/**
+ * Reasoning-effort levels the SDK accepts (`EffortLevel`, sdk.d.ts:539 in the
+ * pinned 0.3.217). Effort controls how much the model thinks AND how many tool
+ * calls it makes to get there, so on this pipeline it is a COST dial: panel
+ * spend tracks agentic turns almost exactly, and turns are what effort moves.
+ *
+ * Frozen and listed literally rather than derived, for the same reason
+ * `PERMITTED_TOOLS` is: if the SDK ever drops a level, a literal list turns that
+ * into a failing test instead of a session that silently runs at the default.
+ */
+export const EFFORT_LEVELS = Object.freeze(["low", "medium", "high", "xhigh", "max"]);
+
+const EFFORT_SET = new Set(EFFORT_LEVELS);
+
+/**
+ * Validate an effort value, returning it unchanged. `undefined` means "unset" and
+ * passes through — the caller then omits the key entirely and takes the SDK
+ * default, exactly as `maxTurns` does.
+ *
+ * Everything else throws, including `null` and a mis-cased or misspelled level.
+ * That strictness is the point: an unrecognised value would otherwise be dropped
+ * and the session would run at the default `high`, which is a silent COST
+ * regression — no error, no changed output, nothing in the logs. The same class
+ * of failure `assertBoundaryMatchesSdk` exists to prevent, one option over.
+ */
+export function assertEffort(effort) {
+  if (effort === undefined) return undefined;
+  if (typeof effort !== "string" || !EFFORT_SET.has(effort)) {
+    throw new Error(
+      `askStructured: \`effort\` must be one of ${EFFORT_LEVELS.join(", ")} ` +
+        `(got: ${JSON.stringify(effort)}). Omit it to take the SDK default.`,
+    );
+  }
+  return effort;
+}
 const PERMITTED_MCP_SET = new Set(PERMITTED_MCP_TOOLS);
 
 /** Is this grant one of the in-process MCP tools rather than a built-in? */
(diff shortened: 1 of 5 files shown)


harrykim8672 (approved):
Thanks for your contribution

Issue #635, opened 2026-08-03
CLI: `status` and `ctx list` print prose and ignore `--format` entirely


hwisoo-kim:
## Summary

`docs/design/cli.md:687-689` (§8.1 Structured Output) promises:

> All output is JSON by default. Errors are also JSON so agents can parse success and failure uniformly.

`status` and `ctx list` emit **bare English prose** on stdout, ignore the documented global `--format` flag entirely, and do not reject an invalid value for it.

## Reproduction

```console
$ wafflebase status
Not logged in. Run `wafflebase login`.

$ wafflebase status --format json          # byte-identical
Not logged in. Run `wafflebase login`.

$ wafflebase status --format bogus          # not rejected either
Not logged in. Run `wafflebase login`.

$ wafflebase ctx list --format json         # same
Not logged in. Run `wafflebase login`.
```

All exit 0. No backend or auth required.

## Cause

`packages/cli/src/commands/status.ts` never reads the format option — its action takes no `opts` at all, and every branch is a bare `console.log` of a human sentence:

```ts
.action(() => {
  const session = loadSession();
  if (!session) {
    console.log('Not logged in. Run `wafflebase login`.');
    return;
  }
  console.log(`Logged in as ${session.user.username} (${session.user.email})`);
```

It never calls the `output()` formatter that every structured command routes through. `ctx list` behaves the same way.

## Why it matters

`status` is precisely the command an agent calls **first**, to decide whether to prompt for login. It is listed in `wafflebase schema` as a first-class command (`{"name": "status", "safety": "read-only"}`) and in the safety table at `docs/design/cli.md:838`. An agent following §8.1 and calling `JSON.parse` on its stdout throws instead of branching — and the failure is at the very start of any session, before anything else can be attempted.

The silently-accepted `--format bogus` compounds it: there is no signal at all that the flag is unsupported here.

## Suggested direction

Route both commands through `output()` so `--format` is honoured, and reject unknown format values rather than ignoring them. If prose is deliberately wanted for humans, that is what a non-JSON `--format` is for — the default should still satisfy the documented contract.

## Environment

`main` at `dc538024a`, built from source (`pnpm cli build`).

## Provenance

Found by the CLI issue-hunting harness (#600/#602) on its first live run — it noticed `status` and `status --format json` returning byte-identical output — then verified by hand. This is ground-truth defect #9 from the original hunting design, now confirmed empirically. The `--format bogus` and `ctx list` observations are mine, from checking the blast radius.

[...]

Issue #634, opened 2026-08-03
CLI: `schema --format table` reports "(no results)" with exit 0 while 39 commands exist


hwisoo-kim:
## Summary

`wafflebase schema` claims **39 commands exist** under the default JSON format, and reports **`(no results)`** with **exit 0** under the documented `--format table`. `--format csv` is broken differently: the whole command list is crammed into a single escaped cell.

An agent or operator doing capability discovery with a documented format gets a successful exit code and the claim that the CLI has no commands.

## Reproduction

```console
$ wafflebase schema --format json | head -c 40
{ "commands": [ { "name": "login", …          # exit 0, 39 commands

$ wafflebase schema --format table
(no results)                                   # exit 0

$ wafflebase schema --format csv | head -c 40
commands "[{""name"":""login"",""descrip…      # exit 0, one unusable cell
```

No backend or auth required — this reproduces on a clean checkout.

## Cause

`packages/cli/src/output/table.ts:5` only formats arrays:

```ts
export function formatTable(data: unknown): string {
  if (!Array.isArray(data) || data.length === 0) return '(no results)';
```

…and `packages/cli/src/commands/schema.ts:33` hands it an **object**:

```ts
output({ commands }, opts.format, opts.quiet);
```

So `formatTable` takes the `!Array.isArray` branch and reports an empty result set. The CSV formatter serialises the same object as one row with one column, which is why the array arrives as an escaped JSON string.

The payload shape is the problem, not the formatters: `{ commands: [...] }` is right for JSON and wrong for every row-oriented format.

## Why this is worth fixing rather than documenting

`(no results)` is not a degraded rendering — it is a **false statement with a success exit code**. A script that checks `exit == 0` and parses the table concludes the CLI exposes nothing. Failing loudly on an unsupported format would be strictly better than this; rendering the array is better still.

## Suggested direction

Pass the array (`output(commands, …)`) so row-oriented formats get rows, keeping the `{ commands }` wrapper for JSON only — or have `formatTable`/`formatCsv` unwrap a single-key object whose value is an array. Either way, `(no results)` should be reachable only when there genuinely are none.

Related but distinct: #585 is `--format yaml` printing `undefined` (a missing switch case). This is a payload-shape mismatch in two *implemented* formats.

## Environment

`main` at `dc538024a`, built from source (`pnpm cli build`).

## Provenance

[...]

PR #637, opened 2026-08-03
Count probes that ran, and require resolvable paths in finding prose

hwisoo-kim:
## Summary

Two observability defects that the **first live run of the executing explorer** exposed. Neither changes what the hunter finds; both change whether you can tell what it did.

## `probesRun` counted probes charged, not probes run

The run reported 26 probes in one sample and 30 in the other. The journals held 23 and 28.

A probe blocked by `assertSafeArgv` is *charged* — the charge deliberately precedes validation, so a stream of malformed calls cannot keep an exhausted session alive — but it never executes, and it was not recorded as a budget refusal either. So it vanished: absent from the journal, absent from `probeRefusals`. A session that burned half its budget on denied commands read exactly like one that probed freely.

- `probesRun` now counts journal entries — what actually ran.
- The tool records its own refusals through a new `budget.noteRefusal`, so **charged minus journalled is always accounted for**.

## Findings cited files the reader cannot open

The run produced `schema.ts:33` and `status.ts:8` in finding prose. Neither resolves.

**My first diagnosis was wrong** and the correction is the interesting part. I assumed the report renderer was truncating paths. The raw journal shows the `citations` arrays were clean, full, repo-root-relative paths throughout — the gate validates those against `codeScope`, so they cannot be bare. The bare refs were in the model's **prose** (`expected` / `observed`), which nothing validates.

So this is a rubric fix, not a code fix. All four charters now require repo-root-relative paths in prose as well as in `citations`, with the reason stated: a bare filename passes every gate and then lands in a filed issue as something a reader cannot click.

## Reviewer's guide

- **`budget.noteRefusal` is the load-bearing change.** The charge-before-validate ordering is deliberate and unchanged; this only makes its consequence visible.
- **Nothing about which findings pass the gate changes.** These are counters and prose guidance.

## Linked Issues

Fixes #

No issue — follow-up from the first live run of #600/#602.

## Author checklist

## Verification

- [x] `agent:tests` — 485 pass
- [x] The refusal accounting is **mutation-tested**: dropping the `noteRefusal` call fails its test
- [x] Evidence is from a real run, not a hypothetical: `8 proposed → 5 unique → 5 novel → 2 reproduced → 2 corroborated → 2 reported`, $4.93, both findings hand-verified and filed as #634 and #635

## Risk Assessment

- **Behaviour risk:** none to the gate. `probesRun` will read *lower* than before for the same run — that is the fix, not a regression.
- **Rollback:** revert. Two modules, four rubric files.

## Notes for Reviewers

- **AI assistance:** written with Claude Code (Opus 5).

- **The live run worked, and is worth recording.** The explorer genuinely iterates — it ran `--help`, then `schema`, drilled into individual commands, then tried `status` and `status --format json`, noticed byte-identical output, and reported it. Three candidates were dropped for citing journal indices that did not exist, which is the honesty check doing precisely what it was built for on its first outing.

- **`soloReproduced` was 0**, so on this run cross-sample agreement cost nothing — it dropped only what replay would have dropped anyway. One data point, but it is the measurement #599 exists to produce.

- **Unrelated, and it has now cost me time twice:** `extractFailureSummary` in `scripts/verify-self.mjs` reported a `WARN` line as a failing lane's summary. Its `/\b(FAIL|ERROR|error|…)\b/` alternation matches warnings and the substring `error` inside unrelated words. Worth tightening separately; I left it out to keep this diff on one subject.

## Summary by CodeRabbit

- **Improvements**
  - File references in generated prose and citations now consistently use repository-relative paths, making reported issues easier to locate.
  - Activity statistics now count only probes that actually execute, excluding safety-refused attempts.
  - Safety-refused probe attempts are now recorded for clearer session tracking while remaining excluded from execution and activity results.

- **Tests**
  - Added coverage to verify safety refusals are tracked accurately without being executed or journaled.

### DIFF
--- a/scripts/agent/charters/contract.md
+++ b/scripts/agent/charters/contract.md
@@ -56,6 +56,13 @@ cannot be cited, and a candidate whose citations do not resolve is dropped.
 exit code and the output. `expected` is what the documentation says should have
 happened. A vague pair cannot be contradicted and therefore cannot be verified.
 
+Write every file reference as a **repo-root-relative path**, in prose as well as in
+`citations` — `packages/cli/src/commands/schema.ts:33`, never `schema.ts:33`. The
+gate only validates the `citations` array, so a bare filename in `expected` or
+`observed` passes review and then lands in a filed issue as something a reader
+cannot open. The first live run produced exactly that: correct paths in
+`citations`, `schema.ts:33` and `status.ts:8` in the prose beside them.
+
 ## In your lane
 
 - Documented exit codes vs what the code actually sets.
--- a/scripts/agent/charters/crash.md
+++ b/scripts/agent/charters/crash.md
@@ -43,6 +43,13 @@ Cite the runs that demonstrate the crash: `probeRefs` is the 0-based indices of
 your own runs this session, and `failingRef` is the one that crashed. You cannot
 cite a run you did not perform.
 
+Write every file reference as a **repo-root-relative path**, in prose as well as in
+`citations` — `packages/cli/src/commands/schema.ts:33`, never `schema.ts:33`. The
+gate only validates the `citations` array, so a bare filename in `expected` or
+`observed` passes review and then lands in a filed issue as something a reader
+cannot open. The first live run produced exactly that: correct paths in
+`citations`, `schema.ts:33` and `status.ts:8` in the prose beside them.
+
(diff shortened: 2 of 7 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #638, opened 2026-08-03
Show review lenses as in-progress while the panel runs

hwisoo-kim:
## Summary

The `agent-review-<lens>` check runs are only written at the **very end** of the ~35-min review-panel job. So while the panel is actually running, the PR keeps showing the **previous** round's conclusion — a red ✗ that looks like a fresh failure when the review is really still in progress. On #632 this made it impossible to tell a running panel apart from a failed one.

This makes each lens check show a live **in-progress** spinner for the duration of the run.

## How

- **New step `Mark lens checks in progress`** (review-panel job): right before the orchestrator starts, create each `agent-review-<lens>` check with `status: in_progress`. It runs *after* `Read prior findings` / `Resolve review scope` on purpose — those read the previous round's checks for finding carry-forward and the incremental-review pointer, and creating empty checks any earlier would make them the "latest" per lens and clobber both. The created run ids are handed to the post step via `$RUNNER_TEMP`.
- **Post step now updates instead of always-creating:** `Post per-lens check runs` updates the same in-progress run to its final `completed` + conclusion. One check run per lens transitions `in_progress → completed` — a live spinner during the run, and **no duplicate or orphaned check**. Falls back to `create` if the start step didn't run (e.g. errored, or a panel already mid-flight on a re-run).
- **Crash safety:** if the panel job is *killed* mid-run (timeout/cancel), the post step's `always()` finaliser can't run, so a spinner could get stuck. The `stalled` job — which already fires on `review-panel` failure/skip and pages a human — now closes any lens check left `in_progress` (needs `checks: write`, was `checks: read`) so the PR shows a resolved state matching the paged comment.

## Test plan

- YAML validated; new step ordered `Read prior findings → Set state reviewing → **Mark lens checks in progress** → Run review panel → Post per-lens check runs`.
- All three embedded github-script bodies syntax-checked with `node --check`.
- No new external permissions beyond `stalled`'s `checks: read → write`; the review-panel job already had `checks: write`.
- Behavior on a normal run: checks spin during the panel, then resolve to their verdicts (same final state as before). On orchestrator error: post step's `always()` still finalizes. On job kill: stalled job closes them.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -476,6 +476,48 @@ jobs:
           PR: ${{ steps.pr.outputs.number }}
         run: node .trusted/scripts/agent/set-state.mjs "$PR" reviewing
 
+      # Flip each lens check to in_progress NOW, before the ~35-min orchestrator
+      # run, so the PR shows the review is running instead of the PREVIOUS round's
+      # stale conclusion (a red X that looks like a fresh failure). Runs AFTER
+      # "Read prior findings" / "Resolve review scope" on purpose: those read the
+      # previous round's checks (findings carry-forward + incremental pointer),
+      # and creating these empty in_progress checks any earlier would make them
+      # the "latest" per lens and clobber both. The ids are handed to the post
+      # step below, which UPDATES these same runs to their final conclusion — one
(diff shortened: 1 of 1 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #639, opened 2026-08-03
Stop feeding infra/quota errors into the next review round

hwisoo-kim:
## The bug (why #632 kept re-surfacing the 429)

When a review lens hits an **API/quota outage** — the `429 You've hit your session limit · resets 3:40am (UTC)` you saw on #632 — the reviewer never actually runs. To fail closed, the panel synthesises a `major` finding whose text **is the error message** (`review-panel.mjs`, the `catch` around the lens run).

That synthetic record then leaked into the cross-round **carry-forward**:

1. The per-lens check's `output.text` is the channel the next round reads to re-check prior findings. Its filter only kept `critical`/`major`, non-backlog findings — the synthetic error is `major`, so it **passed** and was persisted.
2. Next round, `prior-findings.mjs` carried it in via `--prior-findings`, so the panel re-checked *"the review could not run"* as if it were a real blocker.
3. The **verifier is biased to keep** — it drops only on a high-confidence *grounded* refutation. "The review could not run" is not a code claim it can disprove against the diff/repo, so it can never refute it. The bogus finding survives every round.

So neither layer "caught" it because neither was ever designed to: the panel treats it as infra at the *top level* (it pages honestly and sets `infraError`), but that knowledge wasn't propagated to the carry-forward channel; and the verifier has no concept of a finding that is actually an infrastructure error rather than a code claim.

## The fix (layered)

- **`review-panel.mjs`** — tag the synthetic record `{ infra: true }`: honest metadata that this is an infrastructure error, not a code finding. (A genuine no-verdict — model ran, produced nothing — is *not* infra and stays a normal fail-closed blocker.)
- **`agent-review-panel.yml`** *(primary fix)* — persist an **empty** carry-forward (`output.text = '[]'`) for any lens whose `panelEntry` carries `infraError`. An infra lens now carries **nothing** into the next round. It still fails closed via its check `conclusion`, so nothing promotes.
- **`prior-findings.mjs`** *(backstop)* — drop any carried record flagged `infra` on read, so the guarantee holds on both panel paths (scheduled + on-demand) regardless of which producer wrote the check output.

## Test plan

- New unit test in `prior-findings.test.mjs`: an `{ infra: true }` record is dropped while a real blocker in the same lens is still carried. **493 agent tests pass** (was 492 + 1 new), 0 fail.
- `node --check` on both changed scripts; `agent-review-panel.yml` YAML validated.
- `verify:entropy` (dead-code + doc-staleness + dep-freshness): **0 issues**.
- Design note added to `harness-engineering.md` (cross-round re-check section).

## Note

This is a workflow + script change, so it only affects panels that start **after** merge — it won't retroactively unstick #632 (that still needs a manual clear + re-fire, and its underlying issue is the shared org credential's 5h limit).

## Summary by CodeRabbit

* **Bug Fixes**
  * Infrastructure and quota outages are no longer carried forward as code findings in subsequent review rounds.
  * Genuine review findings continue to be preserved and rechecked as expected.
  * Review failures now more clearly distinguish service availability issues from actionable code issues.

* **Tests**
  * Added coverage to verify correct handling of current and legacy infrastructure failure records.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -612,7 +612,18 @@ jobs:
               // output.text so the NEXT round's "Read prior findings" step can
               // re-check them (Part 2). output.summary stays the human-readable body.
               let findingsText = '[]';
-              try {
+              // A lens that hit an API/quota outage (e.g. a 429 session limit)
+              // never actually reviewed — its only "finding" is the synthesised
+              // "Review could not run …" record written to fail closed. That is
+              // NOT a code finding: carrying it forward makes the NEXT round
+              // re-check "the review could not run" as a blocker, and the verifier
+              // (biased to keep) cannot refute it on grounded evidence. So persist
+              // an empty carry-forward for an infra lens; its check still fails via
+              // `conclusion` above, so nothing promotes. `p.infraError` is the same
+              // signal the honest-paging path uses, so the two cannot disagree.
+              if (p && p.infraError) {
+                findingsText = '[]';
+              } else try {
                 const v = JSON.parse(fs.readFileSync(`.agent-review/${lens.id}/verdict.json`, 'utf8'));
                 const norm = (s) => { const x = String(s ?? '').toLowerCase().trim(); return ['critical', 'major', 'minor', 'nit'].includes(x) ? x : 'major'; };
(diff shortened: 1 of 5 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #640, opened 2026-08-03
Record why a verifier session failed, and retry the transient ones

hwisoo-kim:
## Summary

Splits the verifier's single `errored` count into causes, and wraps `verifyFinding`
in `withRetry`.

## Why

Both verifier catch sites discarded the error — `.catch(() => null)` and a bare
`catch {}` — so `classifyResult`'s already-computed `kind` was thrown away and every
distinct cause arrived as the same anonymous `errored++`.

That is exactly how the failures behind #614 read. Eight of eighteen verifications
died at precisely the presence turn ceiling, and the summary could only say they
errored — indistinguishable from a quota outage. Here is that same round rendered
with this change:

```
- Sent to verifier: 18
- ⚠️ Verifier ERRORED on 8 of 18 — those findings are UNFILTERED, not confirmed
- Verifier failures: 1 api-error, 6 turn-limit, 1 no-output
- Refuted: 0 (0 high-confidence, 0 dropped)
```

`turn-limit` is the line that would have pointed straight at the ceiling. It is also
how anyone can now confirm #614 actually fixed it, rather than inferring it.

## `errored` is unchanged, and that is the point

`failures` sits *beside* `errored`, not instead of it, because their **difference** is
the number nobody could see:

- `errored` — blocking findings that ended with **no usable verdict**. That includes
  `resolveClusterVerdict` handing back a folded wording's missing verdict, which never
  threw at all.
- `failures.total` — the subset where a session actually threw, split by cause.

So `errored − failures.total` is the fold-null count, and the renderer names it
(`(+2 folded wording(s) with no verdict)`) rather than leaving `3` vs `1` looking like
a contradiction. Same report-both-and-read-the-gap idiom the file already uses for
`refutedHighConfidence` vs `dropped`.

## The retry

`verifyFinding` had no `withRetry` while detection samples always have, so a single
transient 429 killed a verification outright and its finding rode to the gate
unfiltered. Wrapped **inside** `verifyFinding` rather than at the call sites, so all
three paths — fresh, folded wording, carried-forward prior — get it from one edit
instead of three that can drift.

It cannot burn budget on a ceiling: `withRetry` only retries `err.retryable === true`,
and `classifyResult` marks every `limit` subtype and any session/usage limit
non-retryable. **Which is also why the retry is not what fixed the measured
failures** — those were all `limit`, and #614's ceiling raise is what addressed them.
This covers the genuine transient case (#592's 429 outage) that simply did not occur
in the sampled round.

## Linked Issues

## Author checklist

## Verification

`cd scripts/agent && node --test *.test.mjs` → **497 pass, 0 fail** (1 skip,
pre-existing on `main`). Both renderings above are printed by `renderSummary`, not
hand-written.

`pnpm verify:fast` was not run locally: it needs a full install in a fresh worktree and
`scripts/agent` sits outside the pnpm workspace, so it is covered only by the
`agent:tests` lane inside `verify:self`.

## Risk Assessment

- **User-facing risk:** none. Observability plus a retry, both inside the panel.
- **Rollback plan:** revert the single commit.

**Gate behaviour is unchanged in the fail direction.** A failed verification still
keeps its finding — both catch sites still `return null` — so this cannot let a
finding through. The retry can make the gate *slightly less* blocking, since a
verification that previously died on a transient error may now complete and refute;
that is the intended direction and the reason the retry exists.

**Ledger compatibility:** the new line is omitted when nothing threw *and* when the
field is absent, so healthy rounds and rounds recorded before this shipped both render
byte-identically to before. There is a test asserting exactly that, because an
all-zero line would read as data rather than as absence.

## Notes for Reviewers

**AI assistance:** this PR was written with Claude Code — the measurement, the diff,
and the tests.

Two details worth a look:

1. **`recordVerifierFailure` is total, deliberately.** It runs inside a `.catch`, so
   throwing there would turn "we lost a count" into "we lost a verdict". Junk, a plain
   `Error`, even a non-array accumulator all record rather than blow up — the
   unrecognised ones as `unknown`.

2. **`verifierFailureCounts` uses allowlist membership, not `in`.** `kind` comes from
   an append-only ledger, and `kind: "constructor"` under `in` would walk the
   prototype chain and leave a `NaN` own key on the returned counts — the same
   untrusted-string hazard `confidenceCounts` documents. There is a test pinning the
   returned key set.

While adding these I hit the same trap twice: a source-scan test that greps for a
pattern also matches the docblock *quoting* that pattern. The catch-site guard now
filters comment lines out before asserting.

- Follow-up: #613, #614, #615, #630 and #636 have merged. Next in the plan is
  dropping the duplicate verification of carried-forward findings — on every round ≥ 2
  a finding the fresh pass re-detects is verified twice, in two wordings, in the same
  round.

## Summary by CodeRabbit

* **New Features**
  * Added categorized verifier failure reporting for API errors, time limits, missing output, unknown causes, and total failures.
  * Summary views now display verifier failure counts when failures occur.
  * Retryable verification failures are retried automatically.
* **Bug Fixes**
  * Verifier failures are no longer silently discarded or shown only as missing verdicts.
  * Findings remain correctly blocked when verification fails.
  * Existing records without failure details continue to work without extra summary output.

### DIFF
--- a/scripts/agent/metrics.mjs
+++ b/scripts/agent/metrics.mjs
@@ -253,6 +253,13 @@ export function aggregatePanelStats(entries) {
   // they coerce to 0 — but a non-zero value means those findings were never
   // filtered at all, which is the difference between a review and a raw dump.
   let errored = 0;
+  // WHY those sessions threw, split by `classifyResult`'s kind. Absent from
+  // pre-instrumentation entries, which contribute nothing — so an all-zero
+  // `failures` means "no data yet", and the renderer omits the line rather than
+  // implying zero failures. `errored` above stays the wider count: it also
+  // includes a folded wording whose verdict was never produced, which never
+  // threw and so is deliberately not represented here.
+  const failures = { apiError: 0, limit: 0, noOutput: 0, unknown: 0, total: 0 };
   // Novelty gate. Absent from entries written before it existed, which coerce to
   // 0 — an all-zero row reads as "the gate never fired", the honest reading for
   // both a pre-instrumentation round and a round where it ran inert.
@@ -283,6 +290,9 @@ export function aggregatePanelStats(entries) {
     absenceRefuted += Number(e.verifier && e.verifier.absenceRefuted) || 0;
     unresolved += Number(e.verifier && e.verifier.unresolved) || 0;
     errored += Number(e.verifier && e.verifier.errored) || 0;
+    for (const k of Object.keys(failures)) {
+      failures[k] += Number(e.verifier && e.verifier.failures && e.verifier.failures[k]) || 0;
+    }
     laneBlocking += Number(e.lanes && e.lanes.blocking) || 0;
     laneBacklog += Number(e.lanes && e.lanes.backlog) || 0;
     laneUnknownOrigin += Number(e.lanes && e.lanes.unknownOrigin) || 0;
@@ -291,7 +301,7 @@ export function aggregatePanelStats(entries) {
   }
   return {
     agreementCounts, raised, raisedConfidence, kept,
-    verifier: { sentToVerifier, refuted, refutedHighConfidence, dropped, errored, absenceRaised, absenceRefuted, unresolved },
+    verifier: { sentToVerifier, refuted, refutedHighConfidence, dropped, errored, absenceRaised, absenceRefuted, unresolved, failures },
     lanes: { blocking: laneBlocking, backlog: laneBacklog, unknownOrigin: laneUnknownOrigin },
     clusters: { clustered, collapsed },
   };
@@ -503,6 +513,9 @@ export function renderSummary({ agg, panelAgg, panelStats, panelAttribution, fli
     const rc = panelStats?.raisedConfidence || {};
     const k = panelStats?.kept || {};
     const v = panelStats?.verifier || {};
+  // Absent on every round recorded before this shipped, so `?? {}` is what keeps
+  // those rendering byte-identically rather than gaining an all-zero line.
+  const vf = v.failures || {};
     const l = panelStats?.lanes || {};
     const c = panelStats?.clusters || {};
     const sampledRounds = (ac.identical || 0) + (ac.partial || 0) + (ac.disjoint || 0);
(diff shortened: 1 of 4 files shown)


hwisoo-kim:
Verified against current code — the finding is valid. Fixed in 7e11a76ca, though not the way suggested; reasoning below.

**Confirmed, and here is the exact path.** `verifyBlocking` runs once for the representative and once per folded wording, and the caller only re-verifies folds when the representative's verdict *drops*. So for one clustered finding with two folds:

```
rep verdict drops                     → folds get re-verified
both fold sessions throw              → failures.total = 2
resolveClusterVerdict: drops(null) is false for fold 0
  → returns foldVerdicts[0] ?? null   → finding has no verdict → errored = 1
```

`errored - failures.total` is **-1** there. The old guard `v.errored > vf.total` only suppressed the clause when it went negative, which hid the wrong number rather than fixing it — and left a reader looking at 1 errored against 2 failures with no explanation. The reverse direction is real too: a finding can end with no verdict having thrown nothing at all.

**The fix: state the unit, derive nothing.** `errored` counts FINDINGS, `failures` counts SESSIONS, so the line now reads:

```
- ⚠️ Verifier ERRORED on 1 of 5 — those findings are UNFILTERED, not confirmed
- Verifier failures: 2 session(s) threw — 0 api-error, 2 turn-limit, 0 no-output
```

Both counts stay, because each is independently useful: `errored` says how much of the gate went unfiltered, `failures` says what went wrong and therefore what to do about it. Neither derives from the other. The docblock on `verifierFailureCounts` now spells out the units and the cluster case rather than claiming a fold-null identity.

**One thing I did not do: the separate finding-level fold-null counter.** Once the units are labelled, "several sessions served one clustered finding" reads correctly on its own — a third counter would be new state reporting a number with no action attached to it. If you'd rather have it explicitly, I'll add it, but I'd want a use for the number first.

**Validation.** Agent lane **498 pass / 0 fail**. Tests:

- `review-panel.test.mjs` — the assertion that pinned the bad derivation is replaced by the cluster case that disproves it, driving `resolveClusterVerdict` directly with two null fold verdicts and asserting `failures.total > errored` is legitimate; plus the opposite direction (no verdict, no throw).
- `metrics.test.mjs` — all three shapes render, asserting no remainder clause and no negative in the failures line: 8 findings/8 sessions, **1 finding/2 sessions** (the case that broke), and 3 findings/1 session.

One of my own new assertions was wrong on the way in: a repo-wide `doesNotMatch(/-\d/)` for "no negative" matches `claude-opus-5`. It is scoped to the failures line now.

harrykim8672 (approved):
Thanks for your contribution

PR #641, opened 2026-08-03
Capture per-lens stage detail as a review-panel artifact

dlgpdmsly2:
## Summary

Each review lens now writes `stage-detail.json` into its out dir — what it
detected per sample, and how the verifier ruled on each blocking finding — and a
new upload step publishes those at 30-day retention.

```json
{ "lensDiff": "…", "scopeNote": "", "samples": [[…], […]], "verifications": [{ "population": "fresh", "finding": {…}, "verdict": {…}, "dropped": false }] }
```

**Producer only.** Nothing reads the file. No new job permissions. Zero model
cost — the panel already computed all of this and then discarded it.

## Why

#608 opened the feedback corpus because *"nothing records outcomes, so every
tuning decision was argued from intuition."* That is still true one stage
earlier, and worse: the panel's own decisions are not merely unscored, they are
**unrecoverable** once the job ends.

The two channels that survive a round were built for other jobs:

- **`output.text`** keeps `critical`/`major` only, drops the whole `backlog` lane
  the novelty gate (#583) exists to produce, and drops **trailing findings** until
  the array fits 60k. `review-state.mjs` already refuses to trust it for exactly
  this reason, in as many words: *"the panel workflow trims it to fit a 60k limit
  by dropping findings, i.e. it is **DESIGNED to lose data**."*
- **`output.summary`** is prose whose shape has drifted across #573–#610 — a
  renderer, not a record.

Neither has ever carried the two things a scoring pass most needs, at any
severity:

- **which sample raised what** — `unionSamples` collapses the samples, and
  `lensStats.agreement` keeps a *score*, not the findings it scored;
- **how the verifier ruled on each finding** — `lensStats.verifier` keeps
  tallies; every verdict, refutation ground and citation is discarded.

So `misses.jsonl` can record that the panel was wrong about a PR, but nothing can
say *which stage* was wrong, on which sample, or on what ground. This is the
missing input to the corpus #608 already justified, not a new direction.

### What it records, and why each field

- **`samples`** — raw per-sample findings, *before* `unionSamples` and *before*
  `clusterFindings`. This is the only irreplaceable part: per-sample detection
  cannot be reconstructed from anything that survives today.
- **`verifications`** — every finding that reached the verifier, tagged
  `population: "fresh" | "prior-round"`, with its verdict and whether it dropped.
  `dropped` is **recomputed through the real `isDroppingVerdict`** rather than
  restated, so the capture cannot drift from the gate. `verdict: null` means the
  verifier errored and the finding was kept.
- **`lensDiff`** — the *routed* slice this lens actually reviewed (#582), not the
  whole PR diff. It is what makes a capture replayable.
- **`scopeNote`** — the incremental-scope prompt addendum (`""` in full mode).
  Part of the prompt, so a round is not reproducible without it.

### Two fail directions, both deliberate

**The write can only degrade to "no capture."** It runs inside
`Promise.all(allLenses.map(...))`, where a throw would abort every *other* lens —
turning an instrumentation bug into a review outage. So `writeStageDetail`
wraps everything, logs, and returns `false`. Losing one round's diagnostic costs
a row in a dataset; failing the round blocks a PR. There is no symmetry.

**An unset gate resolves to ON.** A GitHub repo variable that was never set
arrives as the **empty string**, so the ordinary `if (env.X)` flag shape would
resolve *"nobody configured it"* to OFF and the capture would silently never run.
A diagnostic that is quietly absent is worse than one that is loudly broken — the
corpus would look thin rather than unwired, and nothing would page. Unset, empty
and whitespace all mean ON; only an explicit `0`/`false`/`off` disables it.

`AGENT_STAGE_DETAIL_CAPTURE` is passed **raw** into the step's `env:`, with no
`&& … || …` default in YAML: that would put the inverted logic in the one place no
test can reach it. `stageDetailCaptureEnabled` is tested in both directions. This
is the inverse of `AGENT_PIPELINE_ENABLED`'s `== 'true'` opt-in, which is why it
is a named function rather than a truthiness check at the call site.

### Size 

Built with the real `lensReviewPlan`/`sliceDiffByFile` routing over four actual
`main` commits, two findings per sample and three verdicts per lens:

| PR diff | lenses applicable | per run, raw | largest lens |
|---|---|---|---|
| 9.6 KB (#638) | 2 | 25 KB | 12.7 KB |
| 17.4 KB (#636) | 5 | 105 KB | 21.0 KB |
| 204.6 KB | 6 | 1.0 MB | 214.7 KB |
| 470.7 KB | 6 | 2.0 MB | 486.4 KB |

`lensDiff` is 76–97% of every file, so this tracks diff size and nothing else.
`upload-artifact@v4` deflates and these diffs compress 3.1–3.6×, so the worst case
above stores as ~0.6 MB and a typical round under 10 KB. At a few PRs a day and
30-day retention that is single-digit MB standing.

Two honest notes from measuring it: the estimate this was planned against
(30–80 KB per lens) **holds only for small PRs** — a 470 KB diff costs ~25× that.
And routing narrowed the slice by only 0–30% on all four samples, since most
lenses read source files; #582 buys focus, not size.

### Explicit non-goals

- **No verdict may change.** This is instrumentation placed inside the decision
  path, so the design makes that checkable rather than argued: `buildStageDetail`
  is pure and asserted not to mutate its inputs, `writeStageDetail` touches only
  the filesystem, and the call site reads `lensDiff`, `ok`, `detected`,
  `verdicts`, `priorForLens`, `priorVerdicts` and writes to none of them. Capture
  ON and OFF differ by exactly one file on disk.
- **No consumer and no collector.** Deliberately a separate PR — see permissions
  above.
- **`stage-detail.json` must never enter a lens or verifier prompt** — the same
  prohibition `misses.jsonl` carries, for a stronger reason: `lensDiff` is a
  verbatim copy of contributor-authored diff text. If it ever reaches a model it
  goes in fenced as DATA, like the diff it contains.

## Linked Issues

No issue — this completes the motivation stated in #608. Related: #582 (routing,
which defines `lensDiff`), #583 (the `backlog` lane `output.text` drops),
#591/#601 (the clustering order that dictated where the writer goes).

Fixes #

## Author checklist

      (No open or merged PR touches stage capture. Three open PRs do touch
      `review-panel.mjs` — see *Follow-up work*.)
      *Author to tick after review — every line is explained in* Why *above and in
      `docs/tasks/active/20260803-panel-stage-detail-capture-todo.md`.*
- [x] Changes follow the conventions in the touched packages; any deviations are
      explained in *Why* above.
- [x] If AI tools assisted with this PR, I noted where in *Notes for Reviewers*
      below.

## Risk Assessment

- **User-facing risk:** none. No behaviour a PR author sees changes — same check
  runs, same conclusions, same comment bodies. The only observable difference is
  an extra artifact on the workflow run.
- **Data/security risk:** the artifact contains verbatim diff text (already
  visible on the PR) and model-authored finding prose, at 30-day retention on a
  public repo's workflow runs. No secrets pass through the panel's out dir. **No
  new permissions**, and no new network egress or tool access. The one forward
  constraint is recorded in the task doc: this file must never be fed back into a
  model prompt except fenced as DATA.
- **Rollback plan:** set the repo variable `AGENT_STAGE_DETAIL_CAPTURE=false`.
  No revert, no deploy, effective on the next round — capture stops and the
  upload step no-ops via `if-no-files-found: ignore`. That is most of why the gate
  exists. Reverting the commit is also safe: nothing depends on the file.

## Notes for Reviewers

- **Where to look first:** the call site in `review-panel.mjs` (immediately after
  `gatingFindings(merged)`). The claim worth checking is that `fresh: detected`
  and `freshVerdicts: verdicts` are index-aligned — that is the one mistake here
  that tests would not catch, and it is why the writer was re-placed rather than
  copied.
- **Follow-up work:**
  - The collector job that pulls these artifacts into a store, keyed by run and
    commit. Separate PR, separate permissions.
  - `agent-review-on-demand.yml` is deliberately untouched. It runs the same
    panel, so capture is ON there and the files are written, but it uploads no
    artifact (it does not upload the execution log either) and its rounds record
    no check runs and drive no promote/fix loop. Advisory `@claude review`
    invocations are not the population a tuning corpus should be scored on. One
    step to add if you disagree.

## Summary by CodeRabbit

- **New Features**
  - Review results now include detailed per-lens stage information, including routed changes, sample findings, and verifier outcomes.
  - Stage-detail artifacts are retained for 30 days and remain available even when execution logs expire.
  - Capture is enabled by default and can be explicitly disabled through configuration.

- **Bug Fixes**
  - Missing detail files no longer cause artifact uploads to fail.
  - Capture or write failures do not affect review verdicts.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -549,6 +549,17 @@ jobs:
           # SHA the check runs below are attested against, so the pointer and the
           # attestation cannot drift apart.
           HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+          # Per-lens stage-detail capture (uploaded as an artifact below).
+          # DEFAULT ON — an opt-OUT, deliberately the inverse of
+          # AGENT_PIPELINE_ENABLED's `== 'true'` opt-in. Passed RAW, with no
+          # `&& ... || ...` default applied here, because an unset repo variable
+          # interpolates to the EMPTY STRING (the same semantics the BASE_SHA /
+          # SINCE_SHA pair documents above) and the script resolves empty,
+          # whitespace and absent alike to ON. Only an explicit `0`/`false`/`off`
+          # in the repo variable disables it. Deciding the default here instead
+          # would put the inverted logic in a YAML expression that no test can
+          # reach; `stageDetailCaptureEnabled` is tested.
+          STAGE_DETAIL_CAPTURE: ${{ vars.AGENT_STAGE_DETAIL_CAPTURE }}
         run: >-
           node .trusted/scripts/agent/review-panel.mjs
           --diff-file /tmp/pr.diff
@@ -715,6 +726,44 @@ jobs:
             .agent-review/review-lens-stats.json
           if-no-files-found: ignore
           retention-days: 1
+
+      # What each lens actually did this round: per-sample findings before the
+      # union, and the verifier's verdict on every blocking finding. Nothing else
+      # keeps this. The check runs persist only what GATES (`output.text` is
+      # critical/major, drops the `backlog` lane, and drops trailing findings to
+      # fit 60k — review-state.mjs calls that channel "DESIGNED to lose data"),
+      # and `review-lens-stats.json` above keeps counts, not the findings counted.
+      # So panel tuning decisions cannot currently be scored after the fact, which
+      # is the gap #608's feedback corpus exists to close — this records its input.
+      #
+      # A SEPARATE step from the execution log above, on purpose. Those two files
+      # are an intra-run hand-off to promote/fix and their 1-day retention is
+      # chosen to match that lifetime; this one has to outlive the run to be worth
+      # capturing. Extending that step would silently change the hand-off's
+      # semantics, and upload-artifact@v4 rejects a duplicate name within a run,
+      # so the name differs too.
+      #
+      # An artifact rather than a comment or a commit for the same reason the
+      # execution log is one: this job's SDK cwd is the untrusted branch checkout,
+      # so it is not granted the write scopes that would let it push this data
+      # anywhere itself. Anything that collects these belongs in its own job.
+      #
+      # Size tracks diff size and nothing else — `lensDiff` is 76-97% of each
+      # file. Measured over real PRs: ~25 KB per run for a 10 KB diff, ~105 KB for
+      # 17 KB, and ~2 MB for a 470 KB diff (the largest in the last 40 commits).
+      # upload-artifact deflates, and diffs compress ~3x, so the worst case above
+      # stores as ~0.6 MB.
+      - name: Upload per-lens stage detail
+        if: always() && steps.pr.outputs.number != ''
+        continue-on-error: true
+        uses: actions/upload-artifact@v4
+        with:
+          name: review-panel-stage-detail
+          path: .agent-review/*/stage-detail.json
+          # Absent whenever capture is disabled, or a lens was skipped, or the
+          # orchestrator crashed before writing — none of which is an error here.
+          if-no-files-found: ignore
+          retention-days: 30
 
   promote:
     needs: review-panel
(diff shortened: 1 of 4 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #644, opened 2026-08-03
Tier the routed diff out of the stage-detail capture

dlgpdmsly2:
## Summary

Follow-up to #641. That PR added the per-lens `stage-detail.json` capture and measured its own size honestly: `lensDiff` is **76–97% of every file**, and a 470 KB diff costs ~2 MB per run — 25× the ceiling it was planned against. This tiers that field out of the default path.

`lensDiff` is also the only part of the payload that is not ours. Every other field is data this pipeline generated about itself — findings it raised, verdicts it reached, a scope note it rendered. `lensDiff` is a verbatim copy of contributor-authored diff text from a possibly-forked branch. #641 flags that risk and then carries the content on every round anyway, for a consumer that does not exist yet.

It is carried for one job: **replay** — feeding a lens exactly the bytes it saw. So:

- **`lensDiff` content is now opt-in**, behind `STAGE_DETAIL_DIFF_CONTENT`, **default OFF**.
- **`lensDiffSha256`, `lensDiffBytes` and `lensFiles` are always emitted**, in both modes.

### Why a hash rather than re-deriving the slice later

Because the slice is **not a function of the diff alone**. It is the output of the file-class router (#582) applied to the diff, and that router has changed once already and can change again. Re-deriving with a newer router silently yields different bytes than production used — however perfectly the commit was preserved. In `--review-mode incremental` it is a function of two SHAs *plus* the router.

So the hash makes drift **detectable** even when the exact input cannot be reproduced. `lensFiles` is kept for the same reason it cannot be back-filled: `pulls/{n}/files` returns the PR's *current* diff, not the slice reviewed at that round, and it is what keeps a scope-discipline metric computable once the body is gone.

### The default is deliberately the inverse of `STAGE_DETAIL_CAPTURE`

`stageDetailCaptureEnabled` defaults **ON**; `stageDetailDiffContentEnabled` defaults **OFF**. That asymmetry is the design, not an oversight, and it is stated in the code comment, the workflow comment and the task doc so a later reader does not "fix" the two into agreement: capture is a few KB of first-party data useful on every round, while the diff body is bulky, third-party, and useful only when a lens is actually being replayed.

Both gates apply the same empty-string discipline in opposite directions — an unset GitHub repo variable arrives as `""`, not as absent — and both fall to their **own** default on a value they cannot parse. That is why `"banana"` is ON for capture and OFF here: one rule, two defaults.

### `lensDiff` is omitted, not emptied

This is the subtle part. #641's builder degrades a missing `lensDiff` to `""`, so the obvious implementation leaves `lensDiff: ""` on the off path — and that is wrong in a way no test would have caught. A consumer keys on `typeof detail.lensDiff === "string"` and, finding no entry, falls back to the full PR diff — the path built for captures written before `lensDiff` existed. `""` is a real value the router produces for a lens whose slice is empty, so an emptied field would be taken literally and replay the lens against **nothing**, instead of landing on the fallback that is already there.

## Size, measured

Same method and the same four commits as #641's table — real diffs through the real `lensReviewPlan`/`sliceDiffByFile` routing, two findings per sample, three verifier verdicts per lens. The "body ON" column reproduces that table, which is what makes this like-for-like:

| PR diff | lenses | per run, body ON | per run, **default** | reduction | largest lens, default |
|---|---|---|---|---|---|
| 9.6 KB (#638) | 2 | 24.9 KB | **5.2 KB** | 4.7× | 2.6 KB |
| 17.4 KB (#636) | 5 | 104.2 KB | **13.7 KB** | 7.6× | 2.7 KB |
| 204.6 KB (#588) | 6 | 1002.2 KB | **18.4 KB** | 54.5× | 3.3 KB |
| 470.7 KB (#622) | 6 | 2029.5 KB | **29.5 KB** | 68.7× | 5.5 KB |

The default payload no longer tracks diff size at all — it tracks finding count. The 2 MB worst case becomes 29.5 KB.

## Non-goals

- **No verdict can change.** This is serialisation only: which byte of the input is recorded, not which findings are. Both modes derive from the same inputs, neither mutates them, and a test asserts the two payloads are identical **except for the body**.
- **No consumer changes.** Nothing upstream reads `stage-detail.json`; this is writer-only.
- **No change to which findings are captured.** `samples`, `verifications`, `scopeNote` and the population split are untouched.

## A note on failure direction

`buildStageDetail` is evaluated as an *argument* to `writeStageDetail`, so it runs **before** that function's `try/catch`, inside `Promise.all(allLenses.map(...))` — where a throw aborts every other lens. The `createHash` call therefore carries its own guard: if hashing fails, the hash is omitted and the round is unaffected. `Buffer.byteLength` and `sliceDiffByFile` are left unguarded deliberately — both are total over a string, and the router already runs the latter over the whole diff on every round.

## Test plan

- `node --test "scripts/agent/*.test.mjs"` — **512 tests, 511 pass, 0 fail, 1 skipped** (pre-existing). No `node_modules` needed for this lane.
- Flag off → **no `lensDiff` key** (asserted with `in`, not `=== ""`); hash, bytes and file list all present.
- Flag on → body present **and** byte-identical to the router's output, with the same hash as the off path.
- The hash is a **known-answer constant**, not a recomputation — a literal digest for the empty string and for a two-file fixture. Recomputing the expectation with the same `createHash` call would assert only determinism.
- The hash is of the **raw** slice: six byte-distinct variants of one diff (leading space, trailing space, extra newline, CRLF, trimmed) must produce six distinct digests, so a trim or normalisation cannot creep in unnoticed.
- `lensFiles` from a multi-file slice; `lensDiffBytes` in UTF-8 bytes, asserted against a multi-byte string where it exceeds `String.length`.
- The default payload round-trips through `JSON.stringify`/`parse` unchanged.
- Both gates in both directions, including the unrecognized-value case that shows each falling to its own default.

One existing test was **replaced rather than patched**: #641's *"a missing lensDiff or scopeNote degrades to an empty string"* justified itself on the grounds that "a consumer reading `detail.lensDiff.length` never has to guard the field's absence" — reasoning this PR inverts. Its *Fail directions* row in the task doc is superseded by a new one that says so.

## Summary by CodeRabbit

- **New Features**
  - Stage-detail artifacts now include compact diff metadata, including file paths, byte counts, and integrity hashes.
  - Raw diff content is omitted by default and can be enabled when detailed inspection is needed.
  - Artifact capture remains configurable independently from raw diff inclusion.

- **Documentation**
  - Updated artifact-size guidance to clarify the difference between default metadata and optional raw diff content.

- **Tests**
  - Added coverage for flag behavior, hashing, file tracking, sizing, and preservation of review outcomes.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -767,8 +767,9 @@ jobs:
       # upload-artifact deflates, and diffs compress ~3x, so that worst case
       # stores as ~0.6 MB. On the DEFAULT path the body is replaced by a hash, a
       # byte count and the routed file list, which collapses those same four runs
-      # to 5.2 / 13.7 / 18.4 / 29.5 KB — 4.7x to 69x smaller, and bounded by the
-      # finding count rather than by diff size.
+      # to 5.2 / 13.7 / 18.4 / 29.5 KB — 4.7x to 69x smaller. It still grows with
+      # the number of files routed to each lens (`lensFiles` is 48% of that last
+      # figure), but no longer with how much was written inside them.
       - name: Upload per-lens stage detail
         if: always() && steps.pr.outputs.number != ''
         continue-on-error: true
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -560,6 +560,18 @@ jobs:
           # would put the inverted logic in a YAML expression that no test can
           # reach; `stageDetailCaptureEnabled` is tested.
           STAGE_DETAIL_CAPTURE: ${{ vars.AGENT_STAGE_DETAIL_CAPTURE }}
+          # Whether that capture carries the raw routed DIFF BODY as well.
+          # DEFAULT OFF — an opt-IN, and deliberately the OPPOSITE default to
+          # STAGE_DETAIL_CAPTURE directly above. Not an inconsistency: capture is
+          # a few KB of first-party data useful on every round, while the diff
+          # body is 76-97% of the payload and is verbatim contributor text from a
+          # possibly-forked branch, useful only when a lens is actually being
+          # replayed. The default path keeps `lensDiffSha256` + `lensFiles`
+          # instead, which is enough to detect that a re-derived slice drifted.
+          # Passed RAW for the same empty-string reason as above; the script
+          # resolves empty, whitespace and absent alike to OFF, and only an
+          # explicit `1`/`true`/`on` turns it on.
+          STAGE_DETAIL_DIFF_CONTENT: ${{ vars.AGENT_STAGE_DETAIL_DIFF_CONTENT }}
         run: >-
           node .trusted/scripts/agent/review-panel.mjs
           --diff-file /tmp/pr.diff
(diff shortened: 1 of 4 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #559, opened 2026-07-27
Add a local Spec → PR front door (Claude Code CLI)

harrykim8672:
## Summary

A second **front door** to the autonomous pipeline: a developer authors a spec
locally and the Claude Code CLI drives spec → task → implement → local review →
`verify:self` → **draft PR**. The pipeline's back half (CI-fix loop + review
panel + ready-gate) is triggered by CI on `agent/*` branches, **not** by issues,
so a locally-produced `agent/<slug>` draft PR is picked up identically to an
issue-originated one. Only the local front half is new; the back half is reused
unchanged.

### What's here
- **`.claude/commands/spec-to-pr.md`** — the `/spec-to-pr` slash-command playbook
  (spec → **required human gate** → task files → branch → implement → local review
  → local `verify:self` → deterministic handoff → end session).
- **`scripts/agent/spec-to-pr.mjs`** — fail-closed `handoff` (slug + branch +
  commit-trailer + slug-uniqueness + body self-check → push → draft PR →
  `set-state awaiting-ci` → loud "branch is now cloud-owned" notice) + a
  token-gated local `review`.
- **`scripts/agent/disclosure.mjs`** — shared `disclosesAiAuthorship`, the single
  source of truth imported by **both** `mark-ready.mjs` (the ready gate) and
  `spec-to-pr.mjs` (the local self-check), so the local body can never drift from
  the gate it must pass.
- Tests, a Phase 25 design-doc subsection, and task files.

### Notes
- **Base-repo push tier only** — the back half rejects fork-originated runs, so
  this front door works only for someone with push access to the base repo (same
  tier that arms the pipeline). Documented in the playbook + Phase 25.
- **Single-writer** — the local session owns the branch only until the draft PR
  exists; the rebase-on-main happens *before* handoff, the helper prints a loud
  cloud-ownership notice, and the playbook ends the session. Pushing again races
  the cloud fixer.
- Rebased onto current `main` (post #538). Related overlap: `main` now also
  accepts human PRs via the `agent:managed` label / "@claude loop" gate — this PR
  doesn't touch that path (it produces `agent/<slug>` branches, which the gate
  already accepts), but a maintainer may want to reconcile the two front doors.

## Test plan
- [x] `cd scripts/agent && node --test *.test.mjs` — 71/71 (5 new: slug validation,
  `renderPrBody`, disclosure predicate + trailer helpers)
- [x] `node --check`; `pnpm verify:entropy` (dead-code 0, doc-staleness 0);
  pre-push `verify:self` green
[...]

PR #646, opened 2026-08-03
Attribute CodeRabbit findings to the panel's own findings

dlgpdmsly2:
## Summary

`harvest.mjs`'s CodeRabbit signature filed **every** blocking finding as a candidate
panel miss, including the ones the panel caught and merely worded differently. It
could not do otherwise: the only thing it knew about the panel's output was a
**count**. `panelVerdictAt` computed `blockingFindings` and dropped the findings on
the floor, so the harvester could say the panel raised three blockers and never
whether one of them was the comment in hand.

That is the failure the module's own header calls out — harvester noise is
*systematic*, it follows whatever the matcher over-fires on, so it moves the panel
somewhere specific and wrong rather than nowhere.

Two commits, separable:

1. **`finding-match.mjs` + the wiring.** A location signal — symbols, line ranges and
   paths mined from the `summary`/`evidence` prose the finding schema has no field for
   — layered on top of the existing text similarity. `match` suppresses (counted and
   logged), `maybe` emits flagged, `no` emits.
2. **Harvest the PRs the panel actually reviewed.** Signature 2 had inherited
   signature 1's preconditions, and that is where all the data was going.

### The similarity metric is not re-derived

`findingSimilarity`, `summaryTokens`, `DEFAULT_SIMILARITY` and `MIN_SHARED_TOKENS` are
all imported from `rounds.mjs`, which owns them. The only change there is one `export`
keyword on `MIN_SHARED_TOKENS` — `tokenOverlap` needs the same floor for the same
reason, and copying `3` into a second file is the drift `CITATION` and
`REFUTATION_GROUNDS` both exist to avoid.

Anchors only ever **demote** a pair to `maybe`. Measured over 1,249 real pairs from
153 captured findings, **55% of all pairs share at least one symbol** — in a focused
PR nearly every finding names the same identifiers — so location agreement means
"same topic", not "same defect".

### `clusterFindings` is untouched, and here is the evidence for why

This was measured rather than assumed, and the result is a free piece of validation
for your calibration. Replicating `clusterFindings` exactly (single-linkage,
lens-neutral key, `findingSimilarity >= 0.3`) over 153 real captured findings — 10
`(item, lens)` groups, 1,249 candidate pairs — then re-running it with an anchor
condition added:

An anchor guard blocks **0 of the 39 merges** clustering performs.

It is not vacuous. The guard has real discriminating power across the population:

| anchor outcome | all 1,249 pairs |
|---|---:|
| line-range hit | 92 (7.4%) |
| shares ≥1 symbol | 692 (55.4%) |
| **disagrees — would block a merge** | **525 (42.0%)** |

But of the 45 pairs that clear the text threshold, **none** disagree on anchors:

```
{ lineHit: 4, nullOverlap: 0, sharedSymbols: 41, disagree: 0 }
```

**Your clustering is already anchor-consistent.** The `(lens, file)` gate plus the 0.3
overlap threshold never merges anything an independent location signal would object
to. A filter that discriminates on 42% of the population and 0% of the operating
region is pure surface area — on a production gate path — so there is no hook into
`clusterFindings` here, in either direction.

## What changed in `harvest.mjs`

Measuring the matcher found it suppressing nothing on any PR in the repo's history.
Four causes, none of them the matcher:

1. **`isAgentPr` gated the whole function.** Signature 1 needs it — a hand-off only
   exists on a promoted agent PR. Signature 2 does not: *"CodeRabbit flagged something
   our panel reviewed and did not raise"* is exactly as true on a human PR someone ran
   `@claude review` on.
2. **A missing hand-off marker skipped the whole PR.** 18 of 33 agent PRs were opened
   ready rather than promoted from draft, so they carry neither a marker nor a
   `ready_for_review` event. For signature 1 that is correct. For signature 2
   `handoffAt` is not merely unknown but *meaningless* — nothing was handed off.
3. **`panelVerdictAt` reads check runs, and the on-demand review writes none.**
   `agent-review-on-demand.yml` states this four times over (*"purely advisory: it
   records NO check runs"*, `checks: read` and never `checks: write`). Its findings
   exist only as markdown in the comment it posts, so that whole population was
   invisible. `parsePanelComment` now reads it, as a **fallback** — check runs win
   where they exist, because one is the structured record and the other a rendering.
4. **Agent authorship is not evidence that the panel reviewed anything.** Of 11 agent
   PRs carrying CodeRabbit blockers, **9 have no `agent-review-*` check run on any
   commit** — only `verify-*` and codecov. They match `isAgentPr` by branch prefix but
   never went through the panel.

Widening the population opens one hazard, closed here: on a PR the panel never
reviewed, every CodeRabbit blocker would file as a miss. Signature 2 now requires
**empirical** evidence of a review — readable lens runs or a trusted panel comment.

The panel-comment parser is author-gated to the App. That is a security gate, not
tidiness: findings parsed from a comment can **suppress** a candidate, so anyone able
to forge one could silence real misses with no row left behind. The marker alone is
insufficient — any user can type an HTML comment.

## Before / after on real data

Full run over the harvestable population, `--since 2026-07-01` (100 PRs):

| | count |
|---|---:|
| CodeRabbit candidates emitted | 2 |
| **suppressed as restatements of a panel finding** | **3** |
| human-fix candidates | 21 |
| withheld — no evidence the panel reviewed the PR | ~29 |

The three suppressions, each inspected by hand and each a genuine restatement:

- **#582** — CR *"Security lens scopeClasses is missing `design-spec`"* ↔ panel *"The
  security lens's `scopeClasses` omits `design-spec`, so a `docs/design/**`-only PR is
  reviewed by design-fit alone"*. Score 1.00.
- **#582** — CR *"Docs lens `appliesWhen` is missing `docs/…`"* ↔ panel *"…does not
  cover `docs/**/*.txt`, which the `prose` class does"*. Score 1.00.
- **#578** — CR *"Retryability is inferred from message text, not status"*, whose body
  states *"`resets?\b` matches substrings like 'connection reset by peer'"* ↔ panel
  *"the quota-detection regex's `resets?\b` alternative matches ordinary transient
  network errors"*. Same regex, same reasoning. Score 0.75.

Both PRs are human-authored and carry no lens check runs, so **all three were
unreachable before this change** — and all three would have been filed as panel
misses. Both emitted candidates are backed by a real panel review (#548 via check
runs, #559 via a *"Review panel: looks good"* comment), and both carry
`matchVerdict: "no"` — compared and unmatched, not "we could not check".

Nothing was written to `misses.jsonl`: every measurement called `harvestPr()` directly
and `--append` was never passed. The corpus still holds its original 5 records.

## Two bugs found in the ported layer

**It promoted on location alone.** The cross-source match condition scored
`max(tokenOverlap, symbolOverlap)` against the threshold, and `symbolOverlap` is 1.0
whenever both sides name exactly one symbol and it is the same one — so two distinct
defects both touching `parseARef`, sharing no summary vocabulary at all, matched at
1.00. The gate now keys on token overlap; anchors keep their veto and their
contribution to the reported score, and get no vote on clearing the bar.

**A backticked path was mined as a bag of symbols.** `IDENT_RE` ran over every
backtick span including `` `@packages/docs/src/view/text-editor.ts` ``, minting
`packages`, `docs`, `src`, `view` as *symbols*, so any two findings under one
directory shared symbols on tree vocabulary alone. `PATH_RE` already records that
string as a file. CodeRabbit backticks the path on every finding, so this was
reachable rather than theoretical.

## Known limitations

- **No negative control in the real sample.** 3 of 3 matched, so the `no` and `maybe`
  paths are evidenced only by tests.
- **L2's cross-source weights are inherited uncalibrated.** `DEFAULT_SIMILARITY` was
  calibrated for same-run pairs; reusing it as the cross-source bar is a reasoned
  guess. `MIN_SHARED_TOKENS` and the title+prose input (below) are what keep it
  honest, not a fresh calibration.
- **The panel-comment parser requires the exact `<!-- agent-review:<40-hex> -->`
  anchor.** An older or differently-formatted panel comment is invisible to it.
- **Without a hand-off, lens check runs are never read** — `headAtHandoff` anchors
  that read, so only the comment path reaches signature 2 there. Measured cost today:
  none; that population has zero lens runs on any commit, so an all-commits read finds
  nothing either. Asserted in a test so it cannot change silently.
- **A commits-list outage withholds CodeRabbit candidates too**, since `headAtHandoff`
  derives from it. The comment path is unaffected, so the two sources fail
  independently.
- **No LLM adjudication.** The designed L3 stays unbuilt until the `maybe` queue is
  shown to be worth paying to shrink; in the measured window it is empty.

## A note on CodeRabbit's comment format

The token comparison gets **title + prose**; the anchor layer gets the **whole body**.
Both halves matter. A title alone reduces to six significant tokens
(`guard, public, mutation, apis, read, boundary`), and containment is scored over the
smaller set, so two incidentally shared words clear the bar. The whole body fails the
other way and in the dangerous direction: every comment ends with a `🤖 Prompt for AI
Agents` block opening with the same boilerplate sentence, contributing `code`, `fix`,
`issues`, `changes`, `validate` to *every* comparison, plus suggestion blocks that
dump literal source — inflating containment on exactly the vocabulary every finding in
the file already shares. Verified against all 10 inline findings on #548, #594 and
#639. The full body still reaches the anchor layer, where the `around lines N - M`
range in that same block is the most precise location signal a review comment carries.

## Test plan

- `node --test "scripts/agent/*.test.mjs"` — **559 pass, 0 fail** (48 new). This lane
  needs no `node_modules`.
- Both regressions the ported code was built around: a `0` from `findingSimilarity` is
  not read as "different file" (it also means "same file, under the token floor"), and
  a `null` `symbolOverlap` means *no evidence*, never disagreement.
- A third regression for the promotion bug: shared anchors alone reach `maybe`, never
  `match`.
- Harvest-level: a restatement is suppressed and logged, an unrelated finding emits, an
  ambiguous one emits flagged, an unreadable verdict withholds, a matcher throw
  resolves to `maybe`, a forged panel comment is ignored, check runs beat the comment,
  authorship is not evidence of a review, and a search outage degrades to agent PRs.
- *EVERY harvested candidate is unverified* still passes — `verifiedBy` stays `""` and
  no harvest path can set it.
- End-to-end against the live API on #582, #578, #548, #559, #628, #594.

### DIFF
--- a/docs/tasks/active/20260803-finding-level-matching-todo.md
+++ b/docs/tasks/active/20260803-finding-level-matching-todo.md
@@ -1,0 +1,174 @@
+# Finding-level matching: tell a restatement from a miss
+
+Script-only, offline harvesting tooling. Changes nothing about what the panel
+decides or what gets reviewed.
+
+## The problem
+
+`harvest.mjs`'s second signature files every CodeRabbit blocking finding as "the
+panel missed this". It cannot do otherwise: the only thing it knows about the panel's
+output is a **count**. `panelVerdictAt` computed `blockingFindings` via
+`tagPriorFindings` and dropped the findings on the floor, so the harvester could say
+the panel raised three blockers and never whether one of them was the comment in
+hand.
+
+So signature 2 over-fires by construction, and the noise lands on a human curator via
+the `verifiedBy` gate. The module's own header names this as the central risk —
+harvester noise is *systematic*, it follows whatever the matcher over-fires on, so it
+moves the panel somewhere specific and wrong rather than nowhere.
+
+## The change
+
+- [x] `scripts/agent/finding-match.mjs` — the anchor layer and the layered matcher,
+      ported from the eval-harness archive branch. `extractAnchor`, `anchorIsEmpty`,
+      `linesOverlap`, `compareAnchors`, `tokenOverlap`, `matchFindings`, `bestMatch`.
+      The two eval-only shape adapters (`labelToFinding`, `artifactToFinding`) were
+      left behind — they describe artifacts that do not exist upstream.
+- [x] **The similarity metric is not re-derived.** `findingSimilarity`,
+      `summaryTokens`, `DEFAULT_SIMILARITY` and `MIN_SHARED_TOKENS` are imported from
+      `rounds.mjs`, which owns them. What this module adds is a *location* signal —
+      a different axis from text — mined from the `summary`/`evidence` prose the
+      finding schema has no field for.
+- [x] `MIN_SHARED_TOKENS` gained an `export` in `rounds.mjs`. One keyword, no
+      behaviour change, no new caller of the old kind: `tokenOverlap` needs the same
+      floor for the same reason and copying `3` into a second file is the drift
+      `CITATION` and `REFUTATION_GROUNDS` both exist to avoid.
+- [x] `panelVerdictAt` returns `blockers` — the blocking findings themselves —
+      alongside the existing three fields, which callers depend on. The **blocking
+      subset only**: a CodeRabbit blocker "already raised" by a nit the panel filed is
+      not a gate hit, and matching against non-blocking findings would suppress a real
+      gate miss on the strength of a passing remark.
+- [x] `attributeToPanel` in `harvest.mjs` — `match` suppresses (counted and logged),
+      `maybe` emits flagged, `no` emits, and the verdict lands in `panelSaw`
+      alongside its score and the panel finding it matched. No new top-level field;
+      `schema` stays `wafflebase/miss@1`.
+- [x] Both sides compared **lens-neutral**, via the `{ ...f, lens: "" }` copy
+      `clusterFindings` already uses. Not a second approach to the same problem.
+      CodeRabbit's lens is a category→lens guess and is often `""`, and a defect the
+      panel raised under a *different* lens is still a defect the panel raised.
+- [x] `codeRabbitDetail` — title plus prose, cut at the first `<details>`, fence or
+      HTML comment. See "Corrected from the plan".
+- [x] 540 agent tests (503 on `main`), all green, no `node_modules` needed.
+
+## Corrected from the plan
+
+**The plan said to feed the matcher CodeRabbit's `summary`. That is six tokens.**
+`summaryTokens("Guard public mutation APIs at the read-only boundary.")` reduces to
+`guard, public, mutation, apis, read, boundary`, and `findingSimilarity` scores
+containment over the *smaller* token set — so two incidentally shared words clear the
+0.3 bar. The title alone is not a summary, it is a headline.
+
+Feeding the **whole body** fails the other way, and in the dangerous direction. Every
+CodeRabbit comment ends with a `🤖 Prompt for AI Agents` block opening with the same
+boilerplate sentence — *"Verify each finding against current code. Fix only
+still-valid issues…"* — contributing `code`, `fix`, `issues`, `changes`, `validate`
+to **every** comparison, plus committable-suggestion blocks that dump literal source.
+Against a ~15-token panel summary that inflates containment on exactly the vocabulary
+every finding in the file already shares, producing false matches where the panel had
+the *most* findings — eating real misses where they are densest.
+
+So the two channels get different text: **title + prose** for the token comparison,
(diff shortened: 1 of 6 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #649, opened 2026-08-03
One review panel per branch, and 200 turns for the fixer

hwisoo-kim:
## Summary

Two fixes from investigating why **#648** came out `agent:blocked`. Neither the
label nor the page described what actually happened.

## What broke on #648

`agent-review-panel.yml` had **no `concurrency` guard**, so two CI completions
close together started two full pipelines against `agent/586-cli-system-exit-code`:

| run | panel | fix | turns | outcome |
|---|---|---|---|---|
| A `30803409664` | 09:55→10:05 | 10:05→**10:35:52** | **81** | ❌ `Reached maximum number of turns (80)` |
| B `30803718950` | 10:00→10:11 | 10:11→10:32:48 | 76 | ✅ pushed `102e0fa73` |

**Two fixer agents, 21 overlapping minutes, the same nine findings.** B converged
and pushed. A crossed its ceiling and its job failed, which fired `stalled`:

> 🛑 … the fixer agent failed, **so the requested changes were not applied**.

That sentence was false. `102e0fa73` was on the branch and CI-green three minutes
earlier. The paged latch then correctly froze the PR — so the commit that *did*
land was never reviewed. **$22.30 spent, roughly a third of it on a duplicate that
could only lose.**

The same absent guard had already broken **#605** with a different symptom: two
panels completing in the *same second* with identical `external_id` and
contradictory verdicts — correctness 1 major vs 2 major, docs success vs failure,
twelve check runs for six lenses, last writer wins.

## The changes

**1. `concurrency` on the panel, `cancel-in-progress: true`.**

Cancel rather than queue: a superseded panel is reviewing a sha that is no longer
the head, so its verdict is stale before it is written, and queueing pays for it in
full and then discards it. Keyed on the head **branch** — the PR number isn't
resolved until the `gate` job. No `run_id` in the key; that's the mistake that makes
a group match only itself and guard nothing.

**2. Fixer `--max-turns` 80 → 200.**

Above `agent-implement`'s 150 on purpose. The implementer writes a change it planned
itself; the fixer must first **read** code it did not write, at locations the panel
chose, then fix every finding in one pass because the prompt tells it to converge in
one round. At 80 it died at exactly 81 while the duplicate landed the same work in
76 — a coin flip, not a margin. The ceiling is a runaway backstop, not a budget:
`MAX_REVIEW_ROUNDS` and the 45-minute job timeout both bind well before 200 turns of
useful work does.

**3. The page no longer guesses.** On a fixer failure it distinguishes three states:
the branch advanced (applied, unreviewed, naming the sha), it did not (nothing
pushed), or the head couldn't be read (says so).

## The trap this change introduced

Caught before pushing, and it's the part most worth reviewing.

`stalled` ran under **`always()`**. With `cancel-in-progress` now cancelling
superseded runs, a panel cancelled *before it starts* reports `skipped` — which is
one of `stalled`'s own trigger conditions. It would have paged and latched the PR to
`agent:blocked` **at the exact moment a fresher round was starting**, and the latch
would then have stopped that round from running.

Now `!cancelled()`, which still fires on every genuine failure — a failed dependency
is not a cancellation. A spurious page is this job's worst outcome because the page
is **sticky**: every other part of the pipeline reads it as "a human owns this now".

I also checked the remaining `always()` uses in the file. All step-level, and the one
that writes state (`Post per-lens check runs`) is self-healing — a superseded run's
lens checks get overwritten, since check runs resolve latest-per-name.

## Verification

`pnpm verify:self` green (**11/11**) · workflow parses.

- **The three-state reason logic was extracted from the YAML and executed** — inline
  `github-script` JS can hold neither a test nor a linter, so it was run against
  #648's real shape (head moved → "applied and NOT reviewed"), a genuine no-push, and
  an unreadable head. The unknown case says it's unknown rather than falling back to
  "pushed nothing", which would be the same false certainty in a new place.

**200 is a judgement, not a measurement.** The only data points are 76 (converged)
and 81 (died); nothing says where the real ceiling should be. It's deliberately far
above both so the ceiling stops being the binding constraint — the round cap and job
timeout are the intended bounds — but if fixer runs start costing noticeably more per
round, this is the number to revisit.

**`cancel-in-progress` can kill a fixer mid-work.** That's intended (its base is
stale), and safe because the fixer only ever appends a commit. But a cancelled fixer
that had already pushed leaves a commit whose round never completed — the push
re-triggers CI and a fresh panel, so it converges, just not for free.

## Also in here

One stale line in `harness-engineering.md` claiming a paged PR still re-reviews on
every CI-green push. The gate's paged latch ended that, and the sentence sat directly
above what I was editing.

## Deliberately NOT in this PR

The third finding from the investigation: **don't latch on a page whose premise the
branch contradicts.** If the head advanced and CI is green after a page, that's worth
one more panel round. It needs a decision about who may clear a latch — a trust
question, not a bug fix.

**#648 itself still needs unblocking by hand** so one round can review `102e0fa73`.
Nothing here does that.

## Linked Issues

None — from the #648 post-mortem.

## Author checklist

## Verification checklist

## Risk Assessment

- **Data/security risk:** no permission or trust-model change. The concurrency guard
  only reduces what runs; the turn ceiling and the page wording change no scope.
- **Rollback:** revert. Both changes are single-value; removing the `concurrency`
  block restores the previous (racy) behaviour exactly.

## Notes for Reviewers

- **This is the second time the missing guard caused a real failure** (#605, #648).
  The failures looked unrelated — contradictory verdicts vs a mislabelled PR — which
  is why it survived the first one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented overlapping automated review runs for the same branch by cancelling superseded runs.
  * Improved stalled and failure notifications to ignore cancelled runs and accurately reflect branch status.
  * Reduced stale outcomes and duplicate work during automated review and fix processes.

* **Improvements**
  * Increased the automated fixer’s maximum work limit, enabling it to handle more complex changes.

* **Documentation**
  * Documented concurrency issues, completed safeguards, verification results, and remaining manual follow-up.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -21,6 +21,39 @@ on:
     workflows: ["CI"]
     types: [completed]
 
+# ONE panel per PR at a time. Without this, two CI completions close together
+# (a re-run, or a push landing while an earlier round is still in flight) start
+# two full pipelines against the same branch, and it has now broken two PRs in
+# distinguishable ways:
+#
+#   #605 — two panels finished in the SAME SECOND with identical `external_id`
+#          (same reviewed sha, same base, mode full) and CONTRADICTORY verdicts:
+#          correctness 1 major vs 2 major, docs success vs failure. Twelve check
+#          runs for six lenses; whichever wrote last became the truth.
+#   #648 — worse, because the panel dispatches a fixer. Two fixers ran on the
+#          same branch for 21 overlapping minutes on the same nine findings. The
+#          later one converged in 76 turns and pushed; the earlier one crossed
+#          its 80-turn ceiling and its job failed, which made `stalled` page
+#          "the requested changes were not applied" — false, they had just been
+#          applied by the other run. The paged latch then froze the PR, so the
+#          fix that DID land was never reviewed. ~$22 spent, a third of it on a
+#          duplicate that could only lose.
+#
+# CANCEL, not queue: a superseded panel is reviewing a sha that is no longer the
+# head, so its verdict is stale before it is written, and queueing would pay for
+# it in full and then discard it. Cancelling mid-fixer is safe — the fixer only
+# ever appends a commit, and the round the cancellation belongs to is re-run by
+# the push that superseded it.
+#
+# Keyed on the PR's head BRANCH, which is what the `workflow_run` payload
+# carries; the PR number is not resolved until the `gate` job. Two different PRs
+# never share a branch, so this is per-PR in practice. `run_id` is deliberately
+# NOT in the key — that is the mistake that makes a concurrency group match only
+# itself and guard nothing.
+concurrency:
+  group: agent-review-panel-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
 permissions: {}
 
 env:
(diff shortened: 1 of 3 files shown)


harrykim8672 (approved):
Thanks for your contribution

Issue #655, opened 2026-08-04
CLI: upstream error bodies are forwarded verbatim as if they were the error envelope


hwisoo-kim:
## Summary

Six command paths test `if (body?.error)` for *truthiness* before forwarding an upstream error body verbatim. Any JSON error body with a top-level `error` field passes — including every Express/Nest 404 and 500, where `error` is a **string** (`"Not Found"`), not the documented `{code, message}` object. The CLI then prints that body as though it were the documented envelope.

## Documented contract

`packages/cli/README.md:109-110` — a single JSON line on stderr, `{"error":{"code":"…","message":"…"}}`, where `code` is what agents branch on.

## The code

`packages/cli/src/commands/docs.ts:189-198`:

```ts
const res = await getClient(opts).getDocContent(docId);
if (!res.ok) {
  const body = res.data as { error?: { code?: string; message?: string } } | null;
  if (body?.error) {
    // Surface backend-shaped errors (e.g., TYPE_MISMATCH) verbatim
    // so agents reading stderr can act on the `code` field.
    console.error(JSON.stringify(body, null, 2));
    process.exitCode = 1;
    return;
  }
  throw new Error(`HTTP ${res.status}`);
}
```

The cast asserts the shape; the guard never checks it. The intent in the comment — surface *backend-shaped* errors so agents can read `code` — is exactly right, but truthiness does not establish "backend-shaped".

Same pattern at:
- `packages/cli/src/commands/docs.ts:192`, `:254`
- `packages/cli/src/commands/slides.ts:151`, `:204`
- `packages/cli/src/commands/notes.ts:145`, `:202`

## Reproduction

Fully local — a stub returning an Express-shaped 404:

```js
// stub-404.mjs
import http from "node:http";
const BODY = { message: "Cannot GET /api/v1/...", error: "Not Found", statusCode: 404 };
http.createServer((_q, s) => {
  s.writeHead(404, { "content-type": "application/json" });
  s.end(JSON.stringify(BODY));
}).listen(59997, "127.0.0.1");
```

```
$ node stub-404.mjs &
$ WAFFLEBASE_SERVER=http://127.0.0.1:59997 WAFFLEBASE_API_KEY=wfb_dummy \
  WAFFLEBASE_WORKSPACE=ws wafflebase docs content doc-1
{
  "message": "Cannot GET /api/v1/...",
  "error": "Not Found",
  "statusCode": 404
}
```

[...]

Issue #654, opened 2026-08-04
CLI: argument-parsing errors bypass the documented JSON error envelope


hwisoo-kim:
## Summary

Commander's own argument-parsing failures are printed to stderr as bare prose, not as the documented JSON error envelope. Any agent parsing stderr as JSON — the CLI's stated audience — fails to parse them.

## Documented contract

`packages/cli/README.md:109-110`:

> **Errors**: a single JSON line on stderr — `{"error":{"code":"…","message":"…"}}`.

`docs/design/cli.md:928-929` repeats it with a `command` field, and `docs/design/cli.md:23` gives the rationale — "JSON-by-default output, JSON error envelopes" for agent consumers (Gemini CLI, Cursor).

## Reproduction

Against the built CLI (`pnpm cli build`), no backend or auth needed:

```
$ wafflebase docs content
error: missing required argument 'doc-id'          # exit 1

$ wafflebase docs content --bogus-flag
error: unknown option '--bogus-flag'               # exit 1

$ wafflebase sheets get
error: unknown command 'get'                       # exit 1

$ wafflebase docs export
error: missing required argument 'doc-id'          # exit 1
```

Expected in each case: a single JSON line on stderr of the form `{"error":{"code":"…","message":"…"}}`.

## This is a gap, not a missing feature

The envelope machinery works. `outputError` (`packages/cli/src/output/formatter.ts:37-44`) emits exactly the documented shape, and it is reached for runtime failures:

```
$ WAFFLEBASE_SERVER=http://127.0.0.1:59999 wafflebase docs content doc-1
{
  "error": {
    "code": "ERROR",
    "message": "fetch failed"
  }
}
```

Commander exits during parsing, before any command action runs, so `outputError` is never reached on this path.

## Why it matters

A mistyped flag or a missing argument is the single most likely error an agent driver hits, and it is the one case that returns unparseable output. An agent that branches on `error.code` sees a JSON parse failure instead and cannot distinguish "you called me wrong" from "the server is down".

## Suggested direction

Commander exposes `.exitOverride()` and `.configureOutput({ writeErr })`, either of which lets the parse error be routed through `outputError` with a stable code (e.g. `USAGE`). Worth applying at the root program so every subcommand inherits it.

---
Found by the CLI issue hunter and hand-verified against `b252617d`. Not auto-filed.

harrykim8672:
@claude fix this issue

PR #656, opened 2026-08-04
Stop hunt probes reaching production, and fix two silent recall losses

hwisoo-kim:
## Summary

A CLI-hunter measurement run reported **0 findings for $4.24**. The funnel explained why, and the explanation was three defects in the harness itself. All three are fixed here.

Two product defects from the same run are filed separately as **#654** and **#655**.

## 1. The clean room was not isolating the network 🔴

`buildProbeEnv` set `WAFFLEBASE_SERVER` only when a hunt server was configured. Omitting it does **not** mean "no server" — the CLI falls back to its own default, `https://api.wafflebase.io` (`packages/cli/src/config/config.ts:7`).

So with no hunt server set, every probe in the `contract` and `crash` charters was talking to **production**. The 404 bodies in the run's journal came back from the live host; I confirmed it with a direct request.

Nothing was harmed — both charters are non-mutating and unauthenticated, so it was read-only 404s. But the isolation argument rested on `assertSafeArgv` and charter non-mutation rather than on the clean room, and reaching a real deployment has to be something an operator opts into.

The variable is now **always** set, defaulting to loopback port 1 where nothing listens and no packet leaves the machine. The CLI turns that into its documented envelope, so a backend-free charter still observes well-formed behaviour:

```
$ # probe with no configured server, through the real clean room
{ "error": { "code": "ERROR", "message": "fetch failed" } }
```

Preflight said `no server configured` and stopped, which reads as *nothing will be contacted*. It now names the destination on every run — the one line that would have caught this immediately.

## 2. `replay()` compared the wrong probe

`replay()` keyed `observations.at(-1)` while the caller keyed the claim on `failingIndex`. Any candidate whose failing probe was **not last** compared two different probes and could never reproduce. This run demonstrated it live:

```
claim  key (probe 3, the real defect): exit:1|to:0|err:none|kind:text
replay key (probe 4, merely last):     exit:1|to:0|err:none|kind:json
```

That candidate was dropped as `not-reproduced`. Pure recall loss, and invisible — the drop table just said the replay failed. An out-of-range index is now **refused rather than clamped**, since clamping would silently restore the same bug.

## 3. The probe budget bounded thinking, not probing

The clock started at construction, before the session opened — and the explorer also holds `Read`/`Grep`/`Glob`. A `contract` sample was refused on its very **first** `run` call:

> Session probe time exhausted (939s of 600s)

…having executed **zero** probes, because it spent fifteen minutes reading the CLI source — which the charter explicitly asks it to do. A `crash` sample lost its last two probes the same way; 3 of the run's 7 refusals were this, and the starved charter reported nothing at all.

The clock now starts at the first `charge()`, so the opening probe can never be refused for time it did not spend, and the bound still applies once probing begins.

## Note on #655

That defect (upstream error bodies forwarded verbatim) is only reachable **with** a backend — it was found by way of the egress bug above. Now that probes are pinned, it belongs to the backend tier rather than the backend-free charters.

## Test plan

- `cd scripts/agent && node --test *.test.mjs` — 646 pass
- `pnpm verify:self` — 11/11 lanes
- **All four guards mutation-tested** — each fails with its fix reverted, including the out-of-range clamp
- End-to-end: a probe with no configured server returns `{"error":{"code":"ERROR","message":"fetch failed"}}` from loopback instead of a 404 from the live API
- `hunt.mjs preflight` now prints the probe destination on every run

## Summary by CodeRabbit

* **Bug Fixes**
  * Probes without a configured destination now avoid unintended connections to production services.
  * Replay checks now validate the specific probe that reported a failure, improving result accuracy.
  * Initial probing is no longer rejected solely because time elapsed before the first probe.
* **Improvements**
  * Preflight information now clearly identifies the destination used for each run and warns when probes are isolated.
  * Probe time limits are applied more consistently after probing begins.

### DIFF
--- a/scripts/agent/hunt-probe.mjs
+++ b/scripts/agent/hunt-probe.mjs
@@ -93,6 +93,20 @@ export function renderReproSh(probes, { bin = "wafflebase", extra = [] } = {}) {
 // --- the clean room ---------------------------------------------------------
 
 /**
+ * Where a probe's network traffic goes when no hunt server has been named.
+ *
+ * Loopback port 1: nothing listens there, `connect` fails immediately with
+ * ECONNREFUSED, and no packet leaves the machine. Deliberately not a public
+ * black-hole address — a probe should fail fast and locally, not hang on a
+ * route that silently drops.
+ *
+ * The CLI turns this into its documented error envelope
+ * (`{"error":{"code":"ERROR","message":"fetch failed"}}`), so a backend-free
+ * charter still observes well-formed behaviour rather than a crash.
+ */
+export const NO_SERVER = "http://127.0.0.1:1";
+
+/**
  * Build the probe environment. REPLACES the ambient environment; never extends it.
  *
  * This is the load-bearing line in the module. Spreading `process.env` would let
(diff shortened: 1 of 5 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #657, opened 2026-08-04
Count fix attempts, and restore the budget on rerun

hwisoo-kim:
## Summary

`MAX_REVIEW_ROUNDS = 3` meant **one** actual fix attempt, not three. This makes it
mean three, and makes `@claude rerun` restore the budget so a hand-back is worth
something.

## The problem

The cap is documented as *"Failed **fix** rounds before review-round-guard.mjs pages
a human."* What it counted was:

> commits that are single-parent **and** carry a failing lens verdict

The predicate was named `isFixerCommit`, which is what made the miscount look
correct. It's literally `parents.length === 1` — "not a merge commit" — and its own
docstring says it's *"deliberately identity-independent"*. It cannot tell who pushed
a commit.

**The implement workflow pushes twice**: its work, then a self-review fix. On #648
the agent said so at 09:47 — *"Two blocking findings, both fixed in `1d9e19d`."*
Each commit drew its own panel round, and both counted as failed fix rounds before
the fix loop had run once.

Measured on both PRs whose data still exists:

| PR | counted as rounds | real fix attempts | pre-consumed |
|---|---|---|---|
| #648 | 3 | **1** | 2 |
| #605 | 5 | 3 | 2 |

So when #615 lowered the cap 5 → 3 — sound reasoning *for panel rounds* — the real
effect was to cut the fix loop from three attempts to **one**. #648 then paged
*"requested changes 3 times without converging"* after a single attempt. That
attempt had fixed the original nine findings and introduced four new ones in the
SSRF guard it added, which is an ordinary second round. There was no second round.

## The fix

**A commit committed before the panel first spoke cannot be a response to it.** The
discriminator is already in the data, needs no identity, and gives `failedRounds = 1`
on #648 where the old code gave 3.

`isFixerCommit` → `isSingleParentCommit`, because that is all it ever tested.

## `@claude rerun` now restores the budget

#650 landed this command while the above was in progress. It clears the paged latch,
drops `agent:blocked` and re-runs CI — but its own summary said the PR was *"still
bounded by the pipeline's round/attempt caps"*, which on a PR at the cap means one
panel round and an immediate re-page. **That is exactly what #648 did when I
un-stuck it by hand.**

It now writes a hidden `RERUN_MARKER` into its result comment, and the round guard
counts fix attempts only from the newest one.

**Author-checked**, for a sharper reason than the paged latch: the latch only ever
stops work, while this *grants* budget. On a public repo a body test alone would let
any account hand the fixer unlimited attempts. Only the workflow's own bot, or a
human with write access, may move the floor.

## Reworked mid-flight

I had built a separate `@claude retry` for the same job before #650 landed. **I
dropped it** — two verbs for un-sticking a PR would be worse than the gap. That also
removed an inline copy in the gate and the cross-implementation drift test that copy
required: `@claude rerun` deletes the paged comments outright, so there's no latch
left to out-date and no duplication to pin. `agent-review-panel.yml` is byte-identical
to `main` in this PR.

## Fail direction

**The count fails toward counting.** Over-counting pages a round early, which a rerun
undoes; under-counting means the cap never trips and the loop is unbounded,
recoverable only if someone notices. So a commit whose position cannot be established
counts, and with no verdict timestamps anywhere the floor is abandoned and every
failing commit counts — exactly the old behaviour. That is also why the PR #521
fixture still returns 3 unchanged.

## Verification

`pnpm verify:self` green (**11/11**) · **651 agent tests**.

- The counter tested against **#648's real shape** — three commits, real timestamps.
- A test asserts `agent-rerun.yml` emits `RERUN_MARKER`: it's a contract between a
  workflow and a module that cannot import it, and a drifted copy wouldn't error —
  the budget would silently never reset and the PR would re-page after one round,
  exactly as before.
- A latent crash fixed on the way: `(commits ?? []).filter` threw on a non-array, and
  a thrown guard is a dead fix job rather than a conservative one.

**Is 3 still the right number?** It was chosen when it meant one real attempt. Three
genuine fix rounds at ~$12 of panel each is a materially bigger spend than what #615
measured, and nobody has data on how often a third attempt converges.

**The floor is a heuristic about time, not provenance.** A human who pushes to an
agent PR after the first verdict has their commit counted as a fix attempt. That's
arguably right — it did draw a round — but it isn't what the name says.

## Linked Issues

None — from the #648 post-mortem. Companion to #649 (concurrency race), which was the
other, independent half.

## Author checklist

## Verification checklist

## Risk Assessment

- **Data/security risk:** one new trust surface — the rerun marker resets a safety
  cap, so it is author-checked to the same standard as the paged latch. The count
  itself only ever moves in the conservative direction when data is missing.
- **Rollback:** revert. The cap returns to counting panel rounds.

## Notes for Reviewers

- **The rename is the interesting part of the diff.** `isFixerCommit` reading as
  "the fixer pushed this" is why a counter that never counted fix attempts survived
  review for as long as it did.

## Summary by CodeRabbit

- **Improvements**
  - Review-round limits now reflect actual fixer attempts, excluding unrelated changes and merge activity.
  - `@claude rerun` restarts review-round tracking from the rerun point.
  - Rerun requests require trusted maintainer authentication, reducing spoofing risk.
  - Stall and rebuttal-limit alerts are delayed for one attempt after a rerun.
  - CI-fix attempt limits remain cumulative across previous failed runs.
  - Human approval requirements remain unchanged.

- **Documentation**
  - Added guidance and verification details for review-round counting, reruns, fallback behavior, and validation results.

### DIFF
--- a/.github/workflows/agent-rerun.yml
+++ b/.github/workflows/agent-rerun.yml
@@ -161,12 +161,21 @@ jobs:
               core.warning(`Could not re-run CI (${e.message}); the loop will engage on the next CI run.`);
             }
 
+            // The <!-- agent-rerun --> marker is the RESUME POINT, not decoration:
+            // review-round-guard.mjs counts fix attempts only from the newest one, so
+            // un-sticking a PR gives the loop its full MAX_REVIEW_ROUNDS back instead
+            // of resuming a count that was already at the cap. Without it, clearing
+            // the latch buys exactly one panel round and an immediate re-page — which
+            // is what #648 did when it was un-stuck by hand. Kept in the RESULT
+            // comment (rather than a second one) so un-sticking stays one comment, and
+            // hidden rather than prose because the prose will be reworded.
             core.setOutput('outcome',
+              `<!-- agent-rerun -->\n` +
               `🔁 Rerun engaged — cleared ${clearedLatch} paged marker(s) and dropped \`agent:blocked\`. ` +
               (reran
                 ? "Re-running CI now; the review panel will run again and can promote or fix this PR."
                 : "No CI run to re-run — the panel will engage on the next CI run.") +
-              " Still bounded by the pipeline's round/attempt caps, and a human approval is required to merge.");
+              " The fixer's round budget restarts from here; a human approval is still required to merge.");
 
       # Edit the placeholder into the final result. always() so a failure in the
       # clear step still replaces the "working on it" note instead of stranding it.
--- a/.github/workflows/agent-rerun.yml
+++ b/.github/workflows/agent-rerun.yml
@@ -161,21 +161,14 @@ jobs:
               core.warning(`Could not re-run CI (${e.message}); the loop will engage on the next CI run.`);
             }
 
-            // The <!-- agent-rerun --> marker is the RESUME POINT, not decoration:
-            // review-round-guard.mjs counts fix attempts only from the newest one, so
-            // un-sticking a PR gives the loop its full MAX_REVIEW_ROUNDS back instead
-            // of resuming a count that was already at the cap. Without it, clearing
-            // the latch buys exactly one panel round and an immediate re-page — which
-            // is what #648 did when it was un-stuck by hand. Kept in the RESULT
-            // comment (rather than a second one) so un-sticking stays one comment, and
-            // hidden rather than prose because the prose will be reworded.
             core.setOutput('outcome',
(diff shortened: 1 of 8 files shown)


harrykim8672:
@claude review

harrykim8672:
@claude review

harrykim8672:
@coderabbitai review

harrykim8672 (approved):
Thanks for your contribution

PR #658, opened 2026-08-04
Lint scripts/, which had no static analysis at all

hwisoo-kim:
## Summary

`scripts/**` — the ~30 modules that decide whether a PR may merge — had **no static
analysis at all**. This turns on `eslint` for them and fixes the ten pre-existing
violations.

## The gap

`verify:fast` lints `packages/frontend` and nothing else. For the agent control
plane, the only check between a typo and `main` was `node --test`, over whatever
paths a test happens to reach.

**#657 shipped `retryAt`** — an undeclared identifier on the round-cap page path —
straight through **green CI**. No test exercises that page, so the guard would have
thrown a `ReferenceError` at precisely the moment it was meant to latch a PR. The
review panel caught it, which is the expensive way to catch a typo.

Verified on this branch rather than assumed:

```
215:79  error  'retryAt' is not defined  no-undef
```

## Three shape decisions

**`js.configs.recommended`, not a hand-picked rule list.** The whole directory
produced ten violations against the full baseline, all fixed here. A curated subset
has to justify each omission, and the omissions are where the next silent bug lives.

**Rooted at the top level, not in `scripts/agent/`.** That directory is an
npm-managed island: the `deps` job `npm ci`s it and uploads the tree as an artifact
on *every* panel run. A linter there would bloat that artifact for a check that
never runs in that job.

**Declared, not borrowed.** ESLint 9 was already resolvable from `packages/frontend`
via pnpm hoisting — but `globals` was not, and hoisting isn't a contract. The
failure mode of relying on it is this lint silently not running, which is the exact
shape of the gap being closed. Versions match what frontend already pins.

Wired **first** in `verify:fast`, so a typo fails in ~1.9s rather than after the
build chain.

## The ten violations

Eight dead bindings and two redundant regex escapes, all pre-existing:

| file | what |
|---|---|
| `hunt.mjs` | unused `readdirSync`, `HUNT_WORKSPACE_VAR`, `probeCount` |
| `hunt.test.mjs`, `hunt-corpus.test.mjs` | unused `readFileSync` imports |
| `prior-findings.test.mjs` | unused `isCheckRuns` |
| `review-panel.test.mjs` | literal 4-space run in a regex → `{4}` |
| `hunt-probe.mjs` | `\/` inside a character class |
| `scripts/verify-entropy.mjs` | `\-` at the end of a character class |

Both regex edits were checked for **behavioural identity** before changing them —
one is a shell-argument allowlist, the other the doc-staleness matcher, so "looks
equivalent" wasn't good enough.

## I made the exact mistake this PR is about

My first config spread `js.configs.recommended` and then set `rules` for one
override — which **replaced** the recommended set and left `no-undef` **disabled**.
`eslint scripts` still exited 0. I only found it by re-introducing #657's bug and
noticing the lint stayed green.

So the config is under test. `scripts/agent/lint-config.test.mjs` asserts on the
**resolved** rule set: `no-undef` is `error`, and every upstream recommended rule
survives the override — which also means a future ESLint release that *adds* a
recommended rule is inherited rather than quietly missed.

**A config that lints nothing reports success.** That's why this isn't
belt-and-braces. Mutation-tested: removing the spread fails two assertions.

Its imports are dynamic and skip when absent, because `agent:tests` runs with
`scripts/agent/node_modules` **absent** and `ask.test.mjs` enforces that no module
there statically imports a third-party package. **That guard caught my first draft**
— I followed the SDK-dependent probe test's precedent rather than arguing that
resolution would have worked anyway, which is what the guard exists to stop.

## Verification

`pnpm verify:self` green (**11/11**, `verify:fast` 79.6s) · **655 agent tests**
(654 pass, 1 pre-existing skip) · `eslint scripts` clean.

**This is a gate on a directory nobody was gating, so it will bite.** The next
person to add a `scripts/**` module gets a lint failure for something the codebase
tolerated for months. That's the point, but it's friction they didn't ask for, and
`--fix` handles only one of the ten classes found here.

**`no-unused-vars` is configured with `caughtErrors: "none"`.** The codebase uses
bare `catch {}` and `catch (e) { /* ignore */ }` deliberately in fail-soft paths;
flagging those would have meant touching a lot of intentional code in a PR about
turning a linter on. Worth revisiting separately.

## Linked Issues

None — follow-up to the #657 review, which found the defect this closes the class of.

## Author checklist

## Verification checklist

## Risk Assessment

- **User-facing risk:** none. No product code; the ten fixes are dead-binding
  removals and two provably-equivalent regex edits.
- **Data/security risk:** none. Three devDependencies added, matching versions the
  repo already resolves.
- **Rollback:** revert. `verify:fast` returns to its previous chain and the lint
  simply stops running.

## Notes for Reviewers

  and reviewed by me.
- **The most useful thing to check** is `eslint.config.mjs`'s `rules` block. The
  merge-vs-replace distinction there is the whole difference between this config
  working and silently doing nothing, and it is one spread operator wide.

## Summary by CodeRabbit

- **New Features**
  - Added automated linting for agent and project scripts.
  - Integrated script linting into the fast verification workflow.
  - Added safeguards to verify linting rules and configuration remain active.

- **Bug Fixes**
  - Resolved existing script lint warnings and removed unused code.
  - Corrected source checks and pattern handling without changing runtime behavior.

- **Documentation**
  - Documented script linting coverage, verification results, and future type-checking work.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1484,6 +1484,44 @@ visible only in the check body today), and any measurement of how often a
 rebuttal is *right* — which is a `misses.jsonl` question, since an overturn that
 should not have happened is a false negative like any other.
 
+### Phase 29: Lint the agent control plane
+
+**Principle:** Mechanical Enforcement — the cheapest check that could have caught it.
+
+`verify:fast` lints `packages/frontend` and nothing else, so `scripts/**` — the
+~30 modules that decide whether a PR may merge — had **no static analysis at
+all**. The only thing between a typo and `main` was `node --test`, and only over
+paths a test happens to reach.
+
+#657 shipped `retryAt`, an undeclared identifier on the round-cap page path,
+straight through green CI: no test exercises that page, so the guard would have
+thrown a `ReferenceError` exactly when it was supposed to latch a PR. The review
+panel caught it, which is the expensive way to catch a typo.
+
+`eslint.config.mjs` at the repo root, scoped to `scripts/**/*.mjs`, wired into
+`verify:fast` so it fails in about a second rather than after the build chain.
+Three notes on the shape:
+
+- **`js.configs.recommended`, not a hand-picked rule list.** The whole directory
+  produced ten violations against the full baseline — eight dead bindings, two
+  redundant regex escapes — all fixed in the same commit. A curated subset has to
+  justify each omission, and the omissions are where the next silent bug lives.
+- **Rooted at the top level, not in `scripts/agent/`.** That directory is an
+  npm-managed island whose tree is `npm ci`'d and uploaded as an artifact on every
+  panel run; adding a linter to it would bloat that artifact for a check that never
+  runs there.
+- **`eslint`, `@eslint/js` and `globals` are root devDependencies**, not borrowed
+  from `packages/frontend` by pnpm hoisting. Hoisting is not a contract, and the
+  failure mode of losing it is this lint silently not running — the exact shape of
+  the gap it closes.
+
+The config is itself under test (`scripts/agent/lint-config.test.mjs`). That is not
+belt-and-braces: the first version spread `js.configs.recommended` and then set
+`rules`, which **replaced** the recommended set and left `no-undef` disabled.
+`eslint scripts` still exited 0. A config that lints nothing reports success, so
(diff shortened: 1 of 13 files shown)


harrykim8672 (approved):
Thanks for your contribution

Issue #661, opened 2026-08-04
CLI: the error envelope is multi-line pretty JSON and omits the documented `command` field


hwisoo-kim:
## Summary

The error envelope is emitted as pretty-printed JSON spread over several lines and never carries the documented `command` field. The docs specify a single line, and the field is what tells an agent *which* call failed.

Distinct from #654 and #655, which are about the envelope being **bypassed** (bare prose from the arg parser; an upstream body forwarded verbatim). This one is about the shape the CLI emits when the envelope path *is* taken correctly.

## Documented contract

`packages/cli/README.md:109-110`:

> **Errors**: a single JSON line on stderr — `{"error":{"code":"…","message":"…"}}`.

`docs/design/cli.md:930-931` adds the field:

> `{"error":{"code":"…","message":"…","command":"docs.content"}}`

and `docs/design/cli.md:691-698` shows the same three-key shape.

## Reproduction

```
$ export WAFFLEBASE_SERVER=http://127.0.0.1:1 WAFFLEBASE_API_KEY=wfb_dummy WAFFLEBASE_WORKSPACE=ws
$ wafflebase docs content d1 --pages garbage 2>&1
{
  "error": {
    "code": "ERROR",
    "message": "fetch failed"
  }
}
```

Five lines, 71 bytes, no `command`. Expected one line carrying `command: "docs.content"`.

## The code

`packages/cli/src/output/formatter.ts:43-44`:

```ts
console.error(
  JSON.stringify({ error: { code: errorCode(error), message } }, null, 2),
```

`null, 2` produces the multi-line form, and the object literal has no `command` key — there is nowhere for the command name to come from, since `outputError` does not receive it.

## Why it matters

Line-orientation is not cosmetic here. A single JSON line per error is what lets a caller read stderr with a line-delimited parser, interleave it with other output, and match one error to one command. Multi-line output forces a consumer to buffer to EOF and guess at boundaries — and if two errors are ever emitted, they cannot be separated at all.

The missing `command` is the bigger gap for the documented audience: an agent driving several calls sees `{"code":"ERROR","message":"fetch failed"}` with no way to attribute it.

## Suggested direction

Drop the `null, 2` argument, and thread the command name into `outputError` so it can populate `command`. Commander exposes it via the action handler's command object, so the call sites already have it in scope.

---
Found by the CLI issue hunter and hand-verified against `b70302dd1`. Not auto-filed.

harrykim8672:
@claude fix this issue

Issue #660, opened 2026-08-04
CLI: --quiet discards the result body and the error envelope, leaving failures with no diagnostic


hwisoo-kim:
## Summary

`--quiet` suppresses everything, not just progress notices. A successful command produces zero bytes instead of its result, and a *failing* command exits non-zero with **zero bytes on both stdout and stderr** — no diagnostic at all.

## Documented contract

`docs/design/cli.md:926`:

> Text results (json/md/text): stdout by default; `--out` redirects to a file. `--quiet` suppresses progress notices but **preserves the body**.

## Reproduction

No backend or auth needed:

```
$ wafflebase schema           # exit 0, 4797 bytes of JSON
$ wafflebase schema --quiet   # exit 0, 0 bytes          <- body discarded
```

And the worse case — a failure under `--quiet`:

```
$ export WAFFLEBASE_SERVER=http://127.0.0.1:1 WAFFLEBASE_API_KEY=wfb_dummy WAFFLEBASE_WORKSPACE=ws
$ wafflebase docs content d1 --quiet
$ echo "exit=$?"
exit=1
# stdout: 0 bytes
# stderr: 0 bytes
```

Without `--quiet` the same command prints `{"error":{"code":"ERROR","message":"fetch failed"}}`.

## Why it matters

Two separate problems, the second more serious than the first.

**The body is data.** `--quiet` is the natural flag for scripting, and it currently makes every read command return nothing. A caller doing `wafflebase schema --quiet > out.json` silently writes an empty file.

**A non-zero exit with no diagnostic is unactionable.** The caller knows only that something failed — not what, not whether it is retryable. Suppressing *progress* output is a display concern; suppressing the error envelope removes the one machine-readable signal the CLI documents. Errors go to stderr precisely so they survive output redirection and quiet modes.

## Suggested direction

`--quiet` should gate progress notices only, and never the result body or `outputError`. The relevant sink is `packages/cli/src/output/formatter.ts` — `outputError` (`:37-44`) takes `quiet` and honours it; the documented behaviour is that it should not.

---
Found by the CLI issue hunter and hand-verified against `b70302dd1`. Not auto-filed.

harrykim8672:
@claude fix this issue

Issue #659, opened 2026-08-04
CLI: --dry-run is silently ignored by docs list/get and sheets tabs list/cells get


hwisoo-kim:
## Summary

Four commands accept `--dry-run` and then perform the request anyway. Every sibling read command honours it, so this reads as four missed wirings rather than a deliberate scope.

**No data-loss risk** — every write and destructive command honours the flag correctly (verified below). The cost is that `--dry-run` cannot be trusted as a uniform "show me what you would do" for inspection or for agent pre-flight.

## Documented contract

`docs/design/cli.md:707-708` (§8.2 Dry-Run):

> `--dry-run` validates inputs, resolves the target API endpoint, and prints the request that would be sent — without executing it.

## Affected

| command | `--dry-run` |
|---|---|
| `docs list` | **ignored** |
| `docs get <id>` | **ignored** |
| `sheets tabs list <id>` | **ignored** |
| `sheets cells get <id> <tab>` | **ignored** |
| `docs content`, `slides content`, `notes content` | honoured |
| `docs create` / `rename` / `delete` | honoured |
| `sheets cells set` / `batch`, `sheets import` | honoured |

`printDryRun` has 19 call sites (`packages/cli/src/commands/{docs,slides,notes,cells,sheets-import}.ts`), including read-only `GET .../content` paths — the four above simply never call it.

## Reproduction

Point the CLI at a port where nothing listens, so a sent request is provable and nothing can be affected:

```
$ export WAFFLEBASE_SERVER=http://127.0.0.1:1 WAFFLEBASE_API_KEY=wfb_dummy WAFFLEBASE_WORKSPACE=ws

$ wafflebase docs list --dry-run
{ "error": { "code": "ERROR", "message": "fetch failed" } }      # exit 1
```

`fetch failed` only happens if the request was actually attempted — a dry-run must never reach the network. Same for `docs get d1 --dry-run`, `sheets tabs list d1 --dry-run`, `sheets cells get d1 t1 --dry-run`.

Contrast a command that honours it:

```
$ wafflebase docs rename d1 newname --dry-run
{
  "dry_run": true,
  "method": "PATCH",
  "url": "http://127.0.0.1:1/api/v1/workspaces/ws/documents/d1",
  ...
}
```

## Why it matters

[...]

PR #662, opened 2026-08-04
Remove cross-sample agreement from the hunt gate

hwisoo-kim:
## Summary

Cross-sample agreement kept only candidates that ≥2 independent explorer samples found, matched by overlapping `file:line` citations inside `codeScope`. It was the cheapest precision lever here for exactly as long as its own measurement said so — and that measurement has now said the opposite.

## What changed the answer

`soloReproduced` existed to answer *"is 2-of-3 too strict?"*. It read **0** for several runs — agreement discarded nothing real. But those runs were measuring a **starved** harness: the probe budget's clock started before the session opened, so samples burned their allowance on read-only exploration and probed shallowly. #656 fixed that.

The next run probed **137 times instead of 27**, and measured `soloReproduced: 6` **out of 6**. Agreement threw away every reproducing finding.

```
BEFORE #656   2 proposed →  1 unique → 1 novel → 0 reproduced → 0 reported   $4.24
AFTER  #656  11 proposed →  6 unique → 6 novel → 6 reproduced → 0 reported   $9.07
                                                  ^^^^^^^^^^^^ all 6 lost to agreement
```

## Why it failed

Deeper exploration makes samples diverge, and two samples that found the **same** defect cited it a line or two apart:

| defect | sample A | sample B |
|---|---|---|
| envelope shape | `formatter.ts:43` | `formatter.ts:44` |
| `--dry-run` ignored | `docs.ts:75,116` | `docs.ts:70,113` |

Both pairs *also* shared a doc citation (`cli.md:707`, `README.md:109`) — but `codeLocations` filters to `codeScope`, which excludes docs paths, so the one thing they agreed on could not count.

Three real defects came out of that run, hand-verified and filed as **#659**, **#660**, **#661**. None was reported.

## Why not just widen the identity to file level

That is precisely what the deleted `REGRESSION: agreement is OVERLAP, not equality` test argued against, and its reasoning still holds: `formatter.ts` genuinely holds two distinct defects — the exit-code bug at `:39/:46` and the envelope bug at `:43` — so file-level matching would collapse them and hide one.

The trade is real in **both** directions. Rather than swap one silent loss for another, precision now rests on the filters that measure the defect itself rather than the wording of its citation: journal-ref resolution, a deterministic 3× replay, and an adversarial verifier panel that defaults to `refuted`.

This also matches the UI hunter's design, which never had agreement for the same predicted reason.

## `maxVerified` replaces it as the cost bound

Agreement was quietly gating how many candidates reached the two-verifier stage — it admitted **zero of six** on that run, so verification was free. Removing it without a cap would have opened twelve verifier sessions.

Default 4, per charter (not per run, so one charter cannot eat another's budget). **What the cap drops is shown in the report, not just counted** — same reason the agreement-cost section existed: a truncated run must not read as a thorough one.

## Funnel keys

`corroborated` and `soloReproduced` are gone with the stage. Two replace them:

- **`refutedAfterReplay`** is now *the* precision signal — reproduced deterministically and still refused by the panel, which is the judgement agreement used to share. Excludes `unjudged` drops, or the signal would read worse the flakier the API got.
- **`cappedUnverified`** is the recall cost of the cap.

## Incidental fix

One transient API error used to cost a **whole charter**: a failed sample left fewer than 2 for intersection, so the charter reported nothing regardless of what the survivors found. That happened on the run before last, and it is why I originally mis-attributed that run's emptiness.

## Not covered by a test

The cap counter is inline in `cmdRun`, which is not unit-testable without extracting it. It is observable through `cappedUnverified` and the report section, both of which **are** tested.

`intersectSamples` and its four tests are **deleted** rather than left unused — dead code with passing tests is worse than no code, because the tests imply something is exercised. The reasoning is preserved in the commit message and in git history.

## Test plan

- `cd scripts/agent && node --test *.test.mjs` — 646 pass
- `pnpm verify:self` — 11/11 lanes
- Report rendering covered for both new funnel keys and the cap section

## Summary by CodeRabbit

* **Improvements**
  * Candidate review now includes reproduced results held back by verification limits, along with their supporting evidence.
  * Reports clearly distinguish verified results, verifier refutations, and candidates awaiting review.
  * Processing can continue when an individual sample fails, provided other samples succeed.
  * Validation messages and pipeline output now better reflect the updated review process.

* **Bug Fixes**
  * Improved reporting of excluded candidates and verification outcomes.
  * Preserved relevant evidence while applying verification limits.

### DIFF
--- a/scripts/agent/hunt-gate.mjs
+++ b/scripts/agent/hunt-gate.mjs
@@ -29,11 +29,18 @@
 //   keepUnrefuted         drops only on a refute     a crashed verifier reports
 //
 // The replacements below (`huntSeverity`, `coerceCandidates`, `dedupeCandidates`,
-// `intersectSamples`, `isFilingVerdict`) are the same five ideas with the failure
-// direction flipped. (The panel's null-verdict behaviour now lives in
+// `isFilingVerdict`) are those ideas with the failure direction flipped. There was a
+// fifth, `intersectSamples`, inverting `unionSamples` by keeping only what >=2
+// samples agreed on. It is gone: once the probe budget stopped starving samples they
+// explored deeply enough to diverge, and it discarded 6 of 6 reproducing findings
+// over citations one line apart. Precision now rests on replay and the panel.
+//
+// (The panel's null-verdict behaviour now lives in
 // `annotateFindings`/`keepUnrefuted`, which drop a finding only on a concrete
-// refutation — so a verdict that never arrived still KEEPS it there.) They are NOT thin wrappers — do not "simplify" them back
-// into their panel counterparts.
+// refutation — so a verdict that never arrived still KEEPS it there.)
+//
+// The survivors are NOT thin wrappers — do not "simplify" them back into their
+// panel counterparts.
 //
 // What IS safely shared, because it is polarity-neutral: `CITATION` from
 // citation.mjs (the repo's own shared home for it), `globToRegExp` from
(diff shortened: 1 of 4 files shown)


harrykim8672 (approved):
Thanks for contributing

PR #664, opened 2026-08-04
Route the stage-detail capture out of the runner, from both panels

dlgpdmsly2:
## The bug

#641 added the per-lens stage-detail capture and predicted *"the first managed PR after this merges will produce the artifact."* It did not, and neither did the four after it. Every panel run since that merge which provably reached the lenses — `30799533061`, `30803409664`, `30803718950`, `30836949850`, `30837818619` — logged:

```
No files were found with the provided path: .agent-review/*/stage-detail.json
```

**The capture has never left a runner.**

Nothing in the script was wrong, and the same job proves it. In run `30837818619` all six lenses reported `success`, so each passed the unconditional `writeStageDetail` call (it sits after the skip and fail-closed returns, before the verdict write). `STAGE_DETAIL_CAPTURE` echoes empty, which `stageDetailCaptureEnabled` resolves to ON exactly as intended. No `stage-detail capture failed` line, so nothing threw. And the per-lens directories held real content: each check run's `output.summary` is 836–2,838 characters read straight out of `.agent-review/<lens>/summary.md`, written by `writeVerdict` into the **same** `lensOut` from the **same** `mkdirSync`.

The fault is the glob. `upload-artifact` resolves patterns through `@actions/glob` with `excludeHiddenFiles: !include-hidden-files`, and the input defaults to `false`. The globber's search root is the pattern's non-wildcard prefix — for `.agent-review/*/stage-detail.json` that is the **directory** `.agent-review` — and it skips any traversed item whose basename starts with a dot:

```js
// @actions/glob/lib/internal-globber.js
if (options.excludeHiddenFiles && path.basename(item.path).match(/^\./)) continue;
```

The root is the first item on the stack, so it is skipped **before `readdir`** and the whole subtree is never enumerated. Not "the files were filtered out" — the directory was never opened. `if-no-files-found: ignore` then swallows the report, which is why five rounds passed silently.

What made this read as a *write* bug is the sibling step: the execution-log upload, in the same job, from the same hidden directory, **works** — because it names its two files literally, so each search root is the FILE, whose basename is not hidden, and no hidden component is ever inspected. A glob and a literal path resolve through different branches of `getSearchPaths`, so "same directory, same job, one worked" says nothing about the other.

`ci.yml`'s `.harness-reports/` upload already carries `include-hidden-files: true` for exactly this reason; #641 did not carry it across.

Reproduced against `@actions/glob@0.5.0` with `upload-artifact`'s own glob options:

| pattern | `excludeHiddenFiles` | files found |
|---|---|---|
| `.agent-review/*/stage-detail.json` | `true` (today) | **0** |
| `.agent-review/*/stage-detail.json` | `false` (fixed) | 2 |
| the two literal execution-log paths | `true` | 2 |
| the two literal execution-log paths | `false` | 2 |

That is the production log, both steps, exactly.

## The second producer

`agent-review-on-demand.yml` now uploads the same capture — same name, path, retention and flag.

#641 filed this under "Not built", on the grounds that *"advisory `@claude review` invocations are not the population a tuning corpus should be scored on."* That is right about scoring **the gate** and does not apply to the measurement this feeds. Severity-weighted precision has `minor` and `nit` buckets and the independent reviewer we compare against is nit-heavy, so separating *"we missed it"* from *"we found it and called it minor"* needs the panel's **non-blocking** findings. `samples` is the only channel carrying them — its own sibling `verifications` is blocking-only, and every permanent channel (check-run `output.text`, the rendered comment) is critical/major. Measured over all 17 real on-demand comments: **249 blocking against 387 non-blocking**.

Advisory rounds are still not scored *as gate decisions*. Their per-sample severities are the comparison data.

`STAGE_DETAIL_CAPTURE` / `STAGE_DETAIL_DIFF_CONTENT` are now passed on the on-demand panel step too. A **no-op today** — both repo variables are unset, which is the same ON/OFF pair this job already resolved by passing neither — but without them `AGENT_STAGE_DETAIL_CAPTURE=false` would silence the gating capture while this one kept writing, and `DIFF_CONTENT=true` would have one producer emit `lensDiff` and the other silently omit it. Passed raw and resolved in the script, for #641's reason: a YAML default puts the inverted logic where no test can reach it.

## Why one PR

The test written for the bug is what makes the parity step correct. It parses every `upload-artifact` step out of the real workflows and requires `include-hidden-files: true` on any path that makes the globber **walk** (a wildcard, or a trailing `/`) through a dot-prefixed segment. It fails on all three regressions:

| tree | assertion that fires |
|---|---|
| gating step without the flag (= #641 as shipped) | `agent-review-panel.yml "Upload per-lens stage detail" globs the hidden path … without include-hidden-files: true` |
| on-demand step without the flag (= a naive copy) | same message, `agent-review-on-demand.yml` |
| on-demand step deleted | `both the gating and the on-demand panel must upload the stage-detail capture` |

So the parity half needed no test of its own. Comment lines are stripped before parsing — these workflows quote both the flag and the path in prose, the same trap #630/#640/#651 each hit. A literal file path is exempt, because it genuinely is immune. The named assertion uses `filter`, not `find`: with two identically-named steps a `find` would assert about whichever file the directory listing yielded first and leave the other free to regress.

## Scope

**No script change, and no change to what is captured.** `writeStageDetail`, `buildStageDetail`, the gate and the payload are untouched and were never wrong. Both producers already wrote these files; only the route out is new.

**No new permission, and none granted.** `upload-artifact` uses `ACTIONS_RUNTIME_TOKEN`, not `GITHUB_TOKEN`, so it needs no scope — the `deps` jobs upload today holding only `contents: read`. The on-demand `review` job's permissions are byte-identical before and after: `{contents: read, pull-requests: read, checks: read, issues: write}`. No `checks: write`, no check run, no conclusion, no required check. **Nothing here makes an advisory review gating.**

The glob cannot pick up a branch-planted `.agent-review/evil/stage-detail.json` because "Clear stale verdicts" does `rm -rf .agent-review` before the panel runs. The gating `review-panel` job has the identical untrusted-cwd property and already uploads two artifacts.

**Producer only, still.** Nothing reads `stage-detail.json` after this — which is the reason to do both halves together rather than in sequence: the collector gains a second source with no new case to handle.

## Verification

- `node --test "scripts/agent/*.test.mjs"` — **658 tests, 657 pass, 0 fail, 1 skipped**; 657/656 on `main` (`dfc7ec674`).
- Both workflows parse; the on-demand `review` job's permissions are unchanged.
- The invariant holds across every existing upload step: `ci.yml`'s `.harness-reports/` already sets the flag; `docker-publish.yml`'s `${{ runner.temp }}/digests/*` has no hidden segment; every `claude-execution-output.json` and `.tgz` upload is a literal path.
- Artifact layout after the fix is unchanged and confirmed rather than assumed: one search path, so `search.ts` returns `rootDirectory: searchPaths[0]` and the artifact holds `<lens>/stage-detail.json` — the layout the collector was planned against.

## Still open, deliberately

- **The artifact itself is unverified end to end** and cannot be from here: the gating side needs one managed PR to run the panel, the on-demand side one `@claude review`. The number to check against #641's size table is the largest per-lens file — the check #641 asked for and never got.
- **Two producers under one artifact name is a collector requirement.** `stage-detail.json` carries no field naming the workflow that wrote it, so a collector resolving "artifacts named `review-panel-stage-detail` for PR X" gets both producers with nothing in the payload to tell a gating round from an advisory one, and would double-count one head sha as two rounds. Recoverable from run metadata (`event: issue_comment` versus the panel's `workflow_run`). Must be settled in the collector, not here.

Both are recorded as unchecked boxes in the task doc.

## Summary by CodeRabbit

* **New Features**
  * Added detailed review-stage capture and artifact uploads.
  * Added optional capture settings for on-demand reviews.
  * Artifacts now retain generated stage-detail files for 30 days, including files in hidden directories.

* **Bug Fixes**
  * Fixed missing stage-detail files caused by hidden-file upload exclusions.

* **Tests**
  * Added workflow validation and regression coverage for hidden-path artifact uploads.

### DIFF
--- a/.github/workflows/agent-review-on-demand.yml
+++ b/.github/workflows/agent-review-on-demand.yml
@@ -321,6 +321,19 @@ jobs:
           # Via env, never interpolated into the `run:` string — this step holds
           # the Claude token. Same rule as the twin step in agent-review-panel.yml.
           BASE_SHA: ${{ steps.diff.outputs.base }}
+          # Per-lens stage-detail capture, PASSED RAW and for parity with the twin
+          # step in agent-review-panel.yml — not as a way to switch it on. Both
+          # resolve an UNSET repo variable in the script (capture ON, diff content
+          # OFF), which is what this job already got by passing neither.
+          #
+          # Passing them is what stops the two producers from drifting once someone
+          # DOES set one: without these lines `AGENT_STAGE_DETAIL_CAPTURE=false`
+          # would silence the gating capture while this one kept writing, and
+          # `..._DIFF_CONTENT=true` would have one producer emit `lensDiff` and the
+          # other silently omit it — a difference no consumer could see in the
+          # payload.
+          STAGE_DETAIL_CAPTURE: ${{ vars.AGENT_STAGE_DETAIL_CAPTURE }}
+          STAGE_DETAIL_DIFF_CONTENT: ${{ vars.AGENT_STAGE_DETAIL_DIFF_CONTENT }}
         # --prior-findings and --rebuttals are both tolerant of a missing file
         # (existsSync guard in review-panel.mjs → treated as []), so a
         # skipped/failed read step above is safe.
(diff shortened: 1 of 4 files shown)


harrykim8672 (approved):
Thanks for contributing

PR #665, opened 2026-08-04
Add the UI hunter's prediction protocol: the model proposes, code decides

hwisoo-kim:
## Summary

PR 2 of the UI hunter. **Still no AI in it** — pure logic plus one runner change.

PR 1's free oracles only catch defects that announce themselves: a crash, a console error, a broken DOM. Most real UI bugs do none of that, and every one this repo's design doc cites is of that kind — "increase font size" made text smaller, "edit link" inserted one, typing landed in the wrong cell. Nothing threw.

Catching those means letting the agent say what it expected. The moment it may say *"that was wrong"* it is both prosecutor and judge, which is the documented route to curl's inbox. This takes the judge's job back — and it is now the whole precision argument for the UI hunter, since cross-sample agreement came out of the pipeline in #662.

## The protocol

A prediction is a **named reader plus a comparison**, never prose, submitted **with** the action:

```js
act({
  action: { type: "key", key: "Control+z" },
  expect: {
    read: "sheet.cellValue", args: ["A1"],
    op: "equals", value: "@read:4",
    ground: "A", because: "undo restores the value I read before editing",
  },
})
```

The runner performs the read in the **same round-trip** and reports `actual` without comparing anything. That atomicity is the point: a separate read would let the caller act, look, and only then decide what it had "expected".

Six operators, all total over reader values and decidable without executing anything the model wrote. **No regex operator**, though the plan listed one — a model-supplied pattern is code this process would execute, and `contains` covers what it was for.

**Ground A requires `value` to be a journal reference, never a literal.** A literal is the model asserting a fact from its own belief about how the app ought to behave, which is exactly what "the app contradicts itself" must exclude. `@read:4` resolves to what a reader actually returned. Grounds B and C are checked against the docs scope and the page snapshot; D is never eligible.

**Unevaluable is not violated.** Numbers demanded and a string read, a selection that changed length — never a finding. Collapsing that into `violated` would turn every malformed prediction into a defect report.

## Two things measured while building this, both of which changed the code

**`clip()` was corrupting values.** It returned a JSON string, so `[11,18,32]` became `"[11,18,32]"`. Values leaving the runner are now *compared*, not just displayed, and a truncated array compares **wrongly** rather than not at all — a confident wrong answer. Oversized values become a marker object the protocol treats as unevaluable.

**Two false findings appeared within minutes of the protocol first running** — both ground A, both traceable to a real earlier read, both reproducing deterministically:

1. *"I edited a cell, then undid, so the old value is back"* — violated, because `MemStore.undo()` is a no-op (`"No-op for memory store (no history tracking)"`). The product is fine; the sheet surface mounts a test double without history. **Now mechanically avoidable**: the bridge exposes `sheet.canUndo` / `doc.canUndo`, and they disagree — `false` and `true`, verified in the oracle lane.

2. *"I typed XYZ, then undid, so the text is back"* — violated, because docs undo is **per-keystroke**: one undo left `"XY"`. Whether undo should coalesce a typing burst is a convention judgement, not a self-contradiction, and no mechanical check separates that from a real defect.

Case 2 is why the verifier panel stays load-bearing and why its rubric must attack the expectation *before* the behaviour. Both cases are documented at the top of `hunt-ui-expect.mjs` rather than left for someone to rediscover, because the honest claim is that grounding removes a **class** of bad predictions, not all of them.

## Oracle lane

Gains an undo-capability check asserting the **positive** capability only — docs really can undo after an edit. It deliberately does **not** assert that sheets cannot: that is a limitation worth fixing, not a contract worth freezing, and pinning it would make an improvement look like a regression. The sheet value is reported, not enforced.

## Test plan

- `cd scripts/agent && node --test *.test.mjs` — 686 pass (29 new)
- `pnpm verify:self` — 11/11
- `pnpm verify:hunt:oracles` — 10 oracle checks + cell targeting + undo capability
- **All 11 protocol guards mutation-tested** — each fails with its guard removed
- Live browser run: a grounded violation produced an eligible candidate (`expected "10", read "999"`, traced to journal read #1); a prediction that held produced nothing

## Not yet wired

Nothing calls this. PR 3 gives the model the tool; PR 4 is the orchestrator.

## Summary by CodeRabbit

* **New Features**
  * Added automated UI issue hunting with browser-based checks and application-state validation.
  * Added supported predictions for verifying values, relationships, and numeric conditions during UI actions.
  * Added checks for document and sheet undo capability.
* **Bug Fixes**
  * Improved validation for invalid or unsupported predictions and reader names.
  * Improved handling of unstable, oversized, unavailable, or unevaluable values.
* **Documentation**
  * Updated UI issue-hunting documentation and task tracking to reflect the completed capabilities.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1354,7 +1354,8 @@ prediction protocol are new.**
 Canvas, so the a11y tree covers the React chrome and essentially nothing where the
 content is. `window.__WB_HUNT__` on the DEV-only `/harness/hunt` route answers a CLOSED
 registry of named readers instead — `doc.text`, `doc.runs`, `doc.fontSizes`,
-`sheet.cellValue`, `sheet.cellCenter`, `canUndo` — over `MemStore`/`MemDocStore`, with
+`sheet.cellValue`, `sheet.cellCenter`, `doc.canUndo`, `sheet.canUndo` — over
+`MemStore`/`MemDocStore`, with
 the real `DocsFormattingToolbar` mounted. No backend, no login, no cleanup risk. A
 registry rather than an `evaluate(<model JS>)` hook, so the reachable surface is bounded
 by reviewed code; membership is an own-property test, because a plain object literal
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1341,6 +1341,91 @@ it left alone. `--run` is mandatory so there is no "delete everything that looks
 a fixture" mode to reach by accident, and the run id is in the prefix so concurrent
 runs cannot delete each other's fixtures. Docker lifecycle is NOT reimplemented —
 `scripts/verify-integration-docker.mjs` owns it, and `preflight` reports what is missing.
+### Phase 28: UI Issue Hunting
+
+Phase 26's own residual-risk list names the ceiling this addresses: the CLI reaches
+neither Canvas rendering, CRDT collaboration, nor frontend interaction, which is where
+most open UI bugs live. This is the Playwright hunter that section called "the natural
+next surface", and it reuses Phase 26's precision apparatus wholesale — inverted gate,
+3x replay, adversarial verifier panel, novelty ledger. **Only the probe layer and the
+prediction protocol are new.**
+
+**The sensor is a bridge, not the accessibility tree.** Sheets, docs and slides render to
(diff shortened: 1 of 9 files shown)


harrykim8672:
@coderabbitai full review

harrykim8672:
@claude review

harrykim8672 (approved):

PR #666, opened 2026-08-04
Make @claude rerun actually re-engage the loop

hwisoo-kim:
## Summary

`@claude rerun` on #632 and #648 announced it had dropped `agent:blocked`. Both
stayed blocked. **Labels on a pull request need `pull-requests: write`, not
`issues: write`** — and the summary claimed the drop regardless of whether it worked.

## The evidence

From both run logs:

```
##[warning]Could not remove agent:blocked (Resource not accessible by integration).
```

A PR is an issue for most of the API, and its **comments** are reachable with
`issues: write` — which is why the paged-comment deletion in the very same step
succeeded and reported *"cleared 1 paged marker(s)"*. Its **labels** are not. The job
held `pull-requests: read`.

And the message was built unconditionally, so a human read "dropped `agent:blocked`"
with the label sitting right next to it. Same failure shape as #648's page claiming
the requested changes were not applied when they had been: **a summary asserting the
happy path instead of reporting the outcome.**

It now says which of *dropped* / *was not set* / *could not drop* actually happened.
The three branches were extracted from the YAML and executed, since inline
`github-script` JS holds no test.

## The same grant was missing from `review-panel`, and there it was silently dead

**`agent:reviewing` has never appeared on a PR.** Not once. Same `pull-requests: read`,
and `set-state.mjs` is fail-safe — any API error logs and exits 0, which is correct
for a label that gates nothing — so the "Set state → reviewing" step reported success
from the day it was written.

Confirmed by timeline rather than reasoned: **#648, #632, #605 and #633 all go
implementing → fixing → blocked** with no reviewing state between them.

The comment above that grant said *"labels use the issues API"* — the misconception
itself, now corrected in place.

**Why widening it is safe** is the argument its `issues: write` already rested on: the
SDK step receives no GitHub token (`CLAUDE_CODE_OAUTH_TOKEN` only), so neither grant
widens what prompt-injected branch code can reach. This is the hunk to scrutinise if
you scrutinise one — it's a permission widening on the job that holds the Claude
credential.

## Enumerated, not guessed

Jobs that write labels with the default `GITHUB_TOKEN`:

| job | effective `pull-requests` | |
|---|---|---|
| `agent-iterate-ci::iterate` | `write` | inherited workflow-level |
| `agent-review-panel::fix` | `write` | declared |
| `agent-review-panel::stalled` | `write` | declared |
| `agent-review-panel::review-panel` | ~~`read`~~ → `write` | **fixed** |
| `agent-rerun::rerun` | ~~`read`~~ → `write` | **fixed** |

My first scan also flagged `agent-implement`. **That was wrong** — its label write
happens inside the agent's *prompt* using the App token, whose installation
permissions the workflow block doesn't govern, and #648's timeline shows
`agent:implementing` applied successfully by `yorkie-agent[bot]`. Corrected before
touching it.

## The stale labels

Removed by hand on #632 and #648. Both PRs' paged markers were already gone — the
rerun's comment deletion worked — so the label was the only thing left wrong, and
removing it makes the state honest rather than papering over anything.

## Verification

`pnpm verify:self` green (**11/11**) · both workflows parse · the re-scan above
returns `write` for every label-writing job.

**No test covers this.** The bug is a workflow permission, which nothing in the test
suite can observe — it only manifests against the real API. I verified it by reading
the run logs and the label timelines, but the next regression of this kind will be
found the same way: by someone noticing a label didn't change.

**`set-state.mjs`'s fail-safe exit is load-bearing and now demonstrably hides bugs.**
It should stay fail-safe, but a broken grant looked identical to a working one for
months. Reporting the outcome where a human reads it — as the rerun summary now does —
is the cheap half; a periodic reconcile that flags "the label I set isn't there" is
the thorough half, and isn't here.

## Linked Issues

None — reported directly from #632 and #648.

## Author checklist

## Verification checklist

## Risk Assessment

- **Data/security risk:** two permission widenings, both `pull-requests: read → write`
  on jobs that already hold `issues: write`. `rerun` runs no untrusted code (API calls
  only). `review-panel` runs the SDK, but that step receives no GitHub token, which is
  the same reasoning already documented for its `issues: write`.
- **Rollback:** revert. Label writes go back to failing silently, and the rerun summary
  goes back to claiming a drop it didn't make.

## Notes for Reviewers

- **The interesting artefact** is that a fail-safe write plus an optimistic summary
  hid a broken permission for months. Neither alone would have — the fail-safe kept
  the pipeline alive, the summary kept anyone from looking.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pull request label management during review and rerun workflows.
  * Status messages now accurately report whether labels were removed successfully, were absent, or could not be removed.
  * Review states can now be applied reliably when appropriate.
  * Prevented duplicate review panels during CI reruns while ensuring eligible reruns are processed.

* **Documentation**
  * Documented required permissions and label-operation behavior for pull request workflows.
  * Clarified how CI events and reruns affect automated review handling.

### DIFF
--- a/.github/workflows/agent-rerun.yml
+++ b/.github/workflows/agent-rerun.yml
@@ -49,8 +49,14 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     permissions:
-      pull-requests: read # resolve the PR (head repo/branch/labels)
-      issues: write # delete the paged comment, drop agent:blocked, post the note
+      # WRITE, not read. Label operations on a PULL REQUEST need
+      # `pull-requests: write`; `issues: write` covers its COMMENTS but not its
+      # labels, so with `read` the paged comments were deleted successfully and
+      # `removeLabel` failed with "Resource not accessible by integration" — while
+      # the summary below still announced it had dropped the label. Observed on #632
+      # and #648: both reported success and stayed `agent:blocked`.
+      pull-requests: write # resolve the PR, and drop agent:blocked
+      issues: write # delete the paged comment, post the note
       actions: write # re-run CI to re-fire the panel
     concurrency:
       group: agent-rerun-${{ github.event.issue.number }}
@@ -140,10 +146,22 @@ jobs:
 
             // Drop the single-value blocked label (best-effort; a fresh round re-derives
             // the correct state). removeLabel 404s if it isn't set — that's fine.
+            //
+            // The OUTCOME IS RECORDED, not assumed. This used to warn on failure and
+            // then announce success regardless, so #632 and #648 both said "dropped
+            // agent:blocked" while staying blocked — a message a human reads and
+            // believes, contradicted by the label right next to it. A summary must
+            // report what happened, not what was attempted.
+            let labelDropped = null; // true | false | null (was not set)
             try {
               await github.rest.issues.removeLabel({ owner, repo, issue_number: pr_number, name: 'agent:blocked' });
+              labelDropped = true;
             } catch (e) {
-              if (e.status !== 404) core.warning(`Could not remove agent:blocked (${e.message}).`);
+              if (e.status === 404) labelDropped = null;
+              else {
+                labelDropped = false;
+                core.warning(`Could not remove agent:blocked (${e.message}).`);
+              }
             }
 
             // Re-run the latest CI run for the head SHA so the panel fires again: on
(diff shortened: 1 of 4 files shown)


harrykim8672:
@coderabbitai review

harrykim8672:
@claude review

harrykim8672 (approved):
Thanks for your contribution

PR #667, opened 2026-08-04
Derive the harvest cutoff from the panel's own check runs

dlgpdmsly2:
## The problem

`harvest.mjs` keyed *"when did the panel let go of this PR"* off the `<!-- agent-handoff -->` comment, with the `ready_for_review` timeline event as a fallback — and then read *"what had the panel concluded"* **through** that cutoff. Four problems, and the fourth is the one that cost data.

| # | Problem | Consequence |
|---|---|---|
| 1 | **No marker, no harvest.** `mark-ready.mjs` posts the marker inside a `try/catch` *after* flipping the PR ready, so a genuinely promoted PR can carry none. | Measured: **18 of 33 agent PRs** have neither a marker nor a `ready_for_review` event, having been opened ready rather than promoted. |
| 2 | **`ready_for_review` answers a different question.** It fires whether or not the panel ever spoke. | On a manually-readied PR, every later human commit became a candidate. It over-fired on the PRs it covered and covered none of the PRs it was reached for. |
| 3 | **A rebase erased it.** The cutoff was a comment timestamp compared against **committer dates**. | A force-push after promotion pushed every commit past the cutoff, emptied `beforeHandoff`, and left `headAtHandoff` undefined. |
| 4 | **Signature 2 read the panel through signature 1's cutoff.** The comparison set came from whichever commit was head at hand-off. | No hand-off ⟹ no comparison set ⟹ **every CodeRabbit candidate withheld on a PR whose check runs were readable the whole time.** And one set was applied to every finding regardless of which commit it was about. |

Signature 2 never needed a hand-off. *"CodeRabbit flagged something our panel reviewed and did not raise"* is exactly as true on a human PR reviewed via `@claude review`, where nothing was handed off at all.

## The change

The panel's own check runs are the only signal that is present whenever the panel ran, carries a server timestamp, and means "the panel approved":

|  | marker comment | `ready_for_review` | check runs |
|---|---|---|---|
| present whenever the panel ran | no (`try/catch`) | no | **yes** |
| survives a rebase / force-push | yes | yes | **yes** |
| means "the panel approved" | yes | **no** | **yes** |

`completed_at` is stamped by GitHub, so the rebase that rewrites every committer date on the PR cannot move it.

- **`panelRounds(commits, runsFor)`** — one round per commit the panel *concluded* on. Pure, with `runsFor` injected, so the ordering decision is testable without an API.
- **`panelApprovedAt(rounds, markerAt)`** replaces `handoffTime`. **`ready_for_review` is removed outright, not demoted** — keeping it as a last resort would preserve the over-fire in exactly the cases nothing else can check. Its timeline request is gone with it. The marker stays as a fallback for a PR with no readable round, because it *does* mean the panel approved; it is only unreliably present.
- **`commitIndex` + `roundsUpTo`** — the comparison set is the union of rounds **at or before** the commit a finding is about, resolved **per CodeRabbit comment**.
- **`commentRounds`** — an `@claude review` is a round too, with `conclusion: ""`. Not a missing value: it reached no gate conclusion, which is also what keeps it out of `panelApprovedAt`. It used to be a whole-PR fallback reached only when check runs were absent; as a round it gets the same per-commit treatment the gating arm does.
- **`panelRoundAt`** splits the *cheap* half of a verdict (conclusion, reviewedSha, completedAt — all in the check-runs LIST response) from the expensive half. Two phases: one list call per commit to establish the history, and the per-run `withFullOutput` refetch only for rounds a candidate is actually compared against, cached by sha. `panelVerdictAt` now calls it, so the aggregate conclusion rule ("one failing lens means the panel did not let it through") stays in one place rather than growing a second copy.

### One deliberate semantic change

Gating and advisory rounds are now **unioned** rather than ranked. The record's claim is "the panel did not raise this", and both arms run the same lens panel — so a finding either raised is a finding the panel raised. Ranking check runs above the comment made sense when there was one set per PR and the question was which *rendering of one round* to trust; across rounds it discards real findings and files them as misses. `panelSaw` still prefers the gating round, which is where "check runs are the structured record" is actually observable in the output.

## Corrected while building — real data caught it

The first implementation took the **newest** all-success round. Every human fix opens a new round, and that round approves too — so on **#548 the cutoff landed on 2026-07-28, four days after the three human fixes of 2026-07-25, and lost all three, including rows a human has already curated into `misses.jsonl`.** It is the **first** approval, matching `markerHandoffAt`'s own "FIRST marker, not the last": a later re-approval does not un-say the first one.

Also corrected: an unreadable commit now yields an `unreadable: true` round rather than nothing, which keeps the property the previous code deliberately had — *"we know which commit the panel was looking at and we could not read what it concluded"* — and makes the comparison set refuse rather than resolve to `[]`, which would file every CodeRabbit finding on the PR as a miss against a verdict nobody has seen. Timestamps are compared as parsed instants, not sorted as strings.

## Fail directions

| path | on doubt | why |
|---|---|---|
| `panelApprovedAt` | **null** | null means "cannot classify" and signature 1 goes silent. "Approved at time zero" makes every commit a candidate — the over-fire removing `ready_for_review` exists to end |
| an unreadable round | **`null` blockers**, never `[]` | `[]` claims "this round raised nothing blocking" about a round nobody has seen |
| one unreadable commit | costs its round, not the walk | a PR is not forfeited by one bad request |
| an unplaceable finding or round | **widen** the comparison set | what cannot be placed cannot be excluded. Widening can only over-suppress a finding the panel really did raise; narrowing on a guess writes a false row, and a curated false row is permanent |
| the `ROUND_WALK_LIMIT` cap | proceed, and **log** | a silently truncated history makes "never approved" indistinguishable from "we stopped looking" |

## Scope

**No schema change.** `wafflebase/miss@1`, same fields, same order. `handoffAt` keeps its name and now means "when the panel approved", which is what every consumer already assumed. **Ids are unchanged, so `dedupeById`'s first-wins rule still protects every curated `verifiedBy`** — verified on #548, which re-harvests to the same four ids.

No change to `attributeToPanel`, `bestMatch` or `clusterFindings`. No widening to `minor`/`nit` CodeRabbit findings (that needs the panel's non-blocking findings, i.e. the stage-detail collector). `--append` untouched, including both refusal paths, and not run.

## Verification

- `node --test "scripts/agent/*.test.mjs"` — **676 tests, 675 pass, 0 fail, 1 skipped**; 662/661 on `main` (`35206e585`). 14 new tests.
- `eslint scripts` — clean, under the lockfile's pinned `eslint@9.24.0`.
- **Each bug pinned by a test named after the failure**, with a before/after harness confirming the old code behaved differently on 4 of 5:

| case | before | after |
|---|---|---|
| 1 · no marker anywhere | 0 candidates, `skipped` | human-fix + coderabbit, not skipped |
| 2 · manually readied, never approved | 1 human-fix *(the over-fire)* | 0, `skipped` |
| 3 · rebase rewrote every committer date | CodeRabbit withheld | compared against the real verdict |
| 4a · finding earlier than the round that raised it | not suppressed | not suppressed *(regression guard, not a fixed bug)* |
| 4b · finding on the commit the panel raised it at | filed as a miss | suppressed |

- **Real PRs, read-only, before vs after — #548, #559, #578, #582, #591, #649, and the no-marker population #581, #590, #605, #652. Candidate sets identical on all ten.** #548's `handoffAt` moves 22 seconds (the check run at 15:48:06 rather than the marker at 15:48:28) with the same four ids. #591 and #605 change only their `skipped` reporting, correctly: the panel reviewed but never approved, and there is no CodeRabbit input.
- **API cost, measured.** #548 (9 commits) 13 → 24 calls; #582 (7 commits) 4 → 10. One check-runs LIST per commit, minus the timeline request that is now gone. Real PRs here run 1–9 commits, so `ROUND_WALK_LIMIT = 100` never binds.

## The honest limit of this change

Stated plainly because the plan predicted otherwise: **there is no measurable gain on today's data.** #581, #590 and #605 have no marker, but the panel never *approved* them either — so signature 1 has no referent under the new cutoff any more than the old one — and they carry no CodeRabbit blocking findings, so signature 2 has no input. The gains here are demonstrated **synthetically**, not measured.

What this buys is a mechanism that is correct for the next PR. The rebase case and the per-comment anchoring are latent bugs rather than observed losses, and #548 shows the change is behaviour-preserving where the corpus already has rows.

**Not built:** a third signature. `scripts/agent/eval/github-signals.mjs` derives an `is_real` vote from an author's reply to a bot finding ("Done in `<sha>`" / "not a defect because…"). It needs no cutoff at all — it reads a thread, not a timeline — and is the natural next one.

## Summary by CodeRabbit

* **New Features**
  * Review findings now use panel check-run history to determine approval and comparison cutoffs.
  * Supports advisory and unreadable review rounds, commit-specific comparisons, and fallback handoff evidence.
  * Code review findings are evaluated independently from handoff status.

* **Bug Fixes**
  * Improved attribution across rebases, on-demand reviews, and multiple review rounds.
  * Corrected cutoff handling and skip reporting when review evidence is incomplete.

* **Documentation**
  * Added guidance covering panel-round derivation, fallback behavior, limitations, and verification results.

### DIFF
--- a/docs/tasks/active/20260804-harvest-panel-rounds-todo.md
+++ b/docs/tasks/active/20260804-harvest-panel-rounds-todo.md
@@ -1,0 +1,146 @@
+# Harvest: derive the cutoff from the panel's own check runs
+
+The feedback corpus proposer keyed "when did the panel let go of this PR" off a comment
+and, through it, "what had the panel concluded" as well. Both now come from the panel's
+check runs, and the second is resolved per CodeRabbit comment rather than once per PR.
+
+## The problem
+
+`harvestPr` had one cutoff, `handoffTime`, built from the `<!-- agent-handoff -->`
+comment with the `ready_for_review` timeline event as a fallback. Four things were wrong
+with it, and the fourth is the one that cost data.
+
+1. **No marker, no harvest.** `mark-ready.mjs` posts the marker inside a `try/catch`
+   AFTER flipping the PR ready, so a genuinely promoted PR can carry none. Measured: 18
+   of 33 agent PRs have neither a marker nor a `ready_for_review` event, because they
+   were opened ready rather than promoted.
+2. **`ready_for_review` answers a different question.** It fires whether or not the
+   panel ever spoke, so on a manually-readied PR every later human commit became a
+   candidate. It over-fired on the PRs it covered and covered none of the PRs it was
+   reached for.
+3. **A rebase erased it.** The cutoff was a comment timestamp compared against COMMITTER
+   DATES. A force-push after promotion rewrites every one of those, so every commit
+   post-dated the cutoff, `beforeHandoff` emptied, and `headAtHandoff` went undefined.
+4. **Signature 2 read the panel THROUGH signature 1's cutoff.** The comparison set came
+   from whichever commit was head at hand-off. So (a) no hand-off meant no comparison
+   set and every CodeRabbit candidate withheld on a PR whose check runs were readable
+   the whole time, and (b) one set was applied to every finding regardless of which
+   commit it was about — wrong in both directions depending on where the hand-off commit
+   sat relative to the comment.
+
+Signature 2 never needed a hand-off. "CodeRabbit flagged something our panel reviewed and
+did not raise" is exactly as true on a human PR reviewed via `@claude review`, where
+nothing was handed off at all.
+
+## The change
+
+- [x] `panelRounds(commits, runsFor)` — one entry per commit the panel CONCLUDED on,
+      `{sha, index, conclusion, reviewedSha, completedAt, runs}`. Pure, with `runsFor`
+      injected, so the ordering decision is testable without an API.
+- [x] `panelApprovedAt(rounds, markerAt)` replaces `handoffTime`. Check runs first,
+      marker as a fallback, `ready_for_review` **removed outright** rather than demoted —
+      keeping it as a last resort would preserve the over-fire in exactly the cases
+      nothing else can check. The timeline request is gone with it.
+- [x] `commitIndex` + `roundsUpTo(rounds, index)` — the comparison set is now the union
+      of rounds AT OR BEFORE the commit a finding is about, resolved per comment.
+- [x] `commentRounds` — an `@claude review` is a round too, with `conclusion: ""`. It was
+      a whole-PR fallback; as a round it gets the same per-commit treatment the gating
+      arm does, and it can never become signature 1's cutoff.
+- [x] `panelRoundAt` splits the CHEAP half of a verdict (conclusion, reviewedSha,
+      completedAt — all in the check-runs LIST response) from the expensive half. Two
+      phases: one list call per commit to establish the history, and the per-run
+      `withFullOutput` refetch only for rounds a candidate is actually compared against,
+      cached by sha. `panelVerdictAt` now calls it, so the aggregate conclusion rule
+      ("one failing lens means the panel did not let it through") stays in one place.
+- [x] `markerHandoffAt` keeps "FIRST marker, not the last".
+
+## Corrected while building
+
+- **The newest approval is self-defeating, and only real data showed it.** The first
+  implementation took the NEWEST all-success round. Every human fix opens a new round,
+  and that round approves too — so on #548 the cutoff landed on 2026-07-28, four days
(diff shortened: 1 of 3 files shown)


harrykim8672 (approved):
Thanks for contributing

PR #668, opened 2026-08-04
Record the lane the gate routed on in the stage-detail capture

dlgpdmsly2:
The stage-detail capture records every finding that reached the verifier and what the verifier said. It omits what the **gate** decided: the lane.

`annotateFindings` returns copies — `{ ...f, lane }`, which is what makes it safe — and the capture is handed the array that went in, so the annotated one is never named and nothing can reach it.

**Two of the three lanes survive that; one doesn't.** `discarded` is recoverable (`dropped` on each row is `isDroppingVerdict` over the recorded verdict) and `blocking` is the residue. `backlog` is not: it comes from `novelty.origin`, i.e. a `git blame` of the tree *as it stood during the review*. Once the branch moves, that answer is unobtainable — and the artifact expires at 30 days anyway.

So a capture without the lane can't be told apart from one where nothing was demoted: the difference between *"this finding blocked the PR"* and *"this finding was waved past."* That's the question a gate is scored on, and #608 opened this corpus on the grounds that nothing records it. It's also why this is worth doing now rather than when a consumer asks — every other missing field could be backfilled later.

## The change

```js
const annotatedFresh = annotateFindings(detected, verdicts, await noveltiesFor(detected));
const kept = keepUnrefuted(annotatedFresh);
…
fresh: annotatedFresh,   // was: fresh: detected
```

**No behaviour change.** `kept` is still what flows to the merge, the cluster and the verdict — no lens, verifier, lane assignment, gate decision, check run or permission moves; the new name is read by the capture only.

No new argument, no field computed anywhere, no extra `git blame` — the novelties were already awaited on that line for the gate's own use. `annotateFindings` **maps**, so the array is the same length in the same order as `detected`, which is what lets it stand in without disturbing the `freshVerdicts` pairing.

`prior` stays raw on purpose: prior-round findings are routed with `null` novelties deliberately, so their lane is derivable from the verdict already on the row. And the lane isn't recomputed inside `buildStageDetail` — that would be a second router able to disagree with the gate's own.

One reader contract, documented where a consumer will look: **a missing lane means unknown, never blocking.** Every capture written before this has none, and defaulting absence to `blocking` would score demoted findings as gating ones — this bug reintroduced at the other end. Non-blocking findings are the exception; they never carry a lane by design.

## Tests

Three. Two behavioural — a fresh row carries the lane (both findings `confirmed`, so `dropped` is `false` on both while the lanes still come out `backlog`/`blocking`, which is the proof the lane carries information no other recorded field does), and the annotated array preserves verdict alignment where `kept` would not. Passing `kept` is the plausible wrong fix: it's annotated too, so lanes appear and the capture *looks* right while every row after the first drop is paired with another finding's verdict.

The third reads the source, and only because the call site is genuinely unreachable — inside `main`'s per-lens closure, unexported, needing a git repo and live model sessions. It's structural rather than a literal match (per the inertness note in that file, the variable name is read out of the source): it fails on `fresh: detected`, fails on `fresh: kept`, and stays green when the variable is renamed with the wiring intact.

## Verification

- **666 agent tests, 0 fail**, against 663 on `main` — exactly +3. With the Agent SDK installed that's 665 pass / 1 skipped vs 662 / 1; without it (the `agent:tests` lane) 660 / 6 vs 657 / 6. Measured from the committed tree.
- `eslint scripts` clean, on the lockfile's pinned `eslint@9.24.0`.
- **+198 bytes** per blocking fresh finding worst case (a `relocated` origin with both shas and an `alsoAt`), +94 ordinarily — under 1% of a 5–30 KB per-lens capture.
- Still unverified in a real artifact, because **none has left a runner yet** — the same open box #664's section records. The first capture to arrive should show a `lane` on every blocking fresh row and none on a minor.

### DIFF
--- a/docs/tasks/active/20260803-panel-stage-detail-capture-todo.md
+++ b/docs/tasks/active/20260803-panel-stage-detail-capture-todo.md
@@ -610,3 +610,152 @@ no new scope. The on-demand job's permissions are byte-identical before and afte
       advisory one, and would double-count one head sha as two rounds. Recoverable
       from run metadata (`event: issue_comment` versus the panel's `workflow_run`).
       Not a blocker here; it must be settled in the collector.
+
+---
+
+# Follow-up: record the lane the gate routed on
+
+Two lines. The capture records every finding that reached the verifier and what the
+verifier said, and omits the one thing the **gate** decided: which lane the finding
+was routed to. That field cannot be added later, which is the only reason this is
+worth its own change rather than waiting for a consumer to need it.
+
+## The problem
+
+`annotateFindings` attaches `lane` — `blocking`, `discarded` or `backlog` — and it
+returns copies (`{ ...f, lane }`), which is what makes it safe. The capture is handed
+the array that went *in*:
+
+```js
+const kept = keepUnrefuted(
+  annotateFindings(detected, verdicts, await noveltiesFor(detected)),  // labels land here
+);
+…
+writeStageDetail(lensOut, buildStageDetail({ fresh: detected, … }));   // this gets the unlabelled array
+```
+
+The annotated array is never named, so nothing else can reach it.
+
+Two of the three lanes survive that loss; one does not.
+
+| lane | in a capture written today | why |
+|---|---|---|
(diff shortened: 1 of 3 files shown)


harrykim8672 (approved):
Thanks for your contribution

PR #669, opened 2026-08-04
Report the review panel's real wall-clock as Total-time

hwisoo-kim:
## Summary

The agent-effort comment's **Total-time** for a review round was wildly overstated — #665 and #666 each ran ~10-15 min but were reported as **36 min and 63 min**.

**Cause:** for a `review` record, `metrics.mjs` sums `duration_ms` across *every* SDK result message in the execution log. But the panel runs its six lenses — and each lens's samples + verifier calls — **concurrently**. Summing concurrent durations overcounts by roughly the concurrency factor, so ~12 min of real elapsed time becomes ~60.

## Fix

- `review-panel.mjs` stamps its own **wall-clock elapsed** (orchestrator start→finish) into a sibling `review-timing.json`.
- `metrics.mjs` (`readWallMs`) reads that sibling and uses it as the round's `durationMs`, **falling back to the old summed value when the file is absent** — so pre-instrumentation logs are unchanged.
- The autonomous panel's artifact upload carries the new file to the separate metrics-record job; the on-demand review path reads it as a same-job sibling (no workflow change needed there).

The summed per-call durations are still a fine measure of *compute*, but Total-time should be the elapsed time a maintainer actually waited — that's what this restores.

Demonstration (8 concurrent sessions summing to 68m, real elapsed 12m):

```
BEFORE (summed concurrent durations): 68m
AFTER  (true wall-clock):             12m
```

## Test plan

- New `readWallMs` unit test: sibling present/absent, explicit-path override, non-positive/malformed/missing all fail-safe to `null`.
- `metrics.test.mjs` green (29 tests); full agent suite green (662 pass, 1 pre-existing skip).
- `review-panel.mjs` / `metrics.mjs` `node --check` clean; workflow YAML validates.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -957,6 +957,7 @@ jobs:
           path: |
             .agent-review/review-execution.json
             .agent-review/review-lens-stats.json
+            .agent-review/review-timing.json
           if-no-files-found: ignore
           retention-days: 1
 
--- a/scripts/agent/metrics.mjs
+++ b/scripts/agent/metrics.mjs
@@ -137,6 +137,29 @@ export function sumExecutions(messages, kind = "review") {
   };
 }
 
+/**
+ * The panel's TRUE wall-clock duration for one round, from review-panel.mjs's
+ * `review-timing.json` (written next to the execution log). This exists because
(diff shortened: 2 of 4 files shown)


harrykim8672:
@claude review

harrykim8672 (approved):
Thanks for your contribution

PR #672, opened 2026-08-05
Add @claude fix, an on-demand fixer for the panel's verdict

hwisoo-kim:
## Summary

Adds **`@claude fix`** on a PR: a maintainer comments it and one fix agent runs against the review panel's standing verdict, immediately.

The fixer only ever ran on the loop's terms — inside `MAX_REVIEW_ROUNDS`, only after a fresh panel round, and not at all once the loop had paged. The closest handle was `@claude rerun`, which restarts the whole cycle (~13 min of CI, then a full panel) to reach a fixer that already had everything it needed.

### 1. Eligibility: a panel ran, and nothing landed after it

`scripts/agent/fix-eligible.mjs` reduces both halves to one unforgeable question: does the **current head sha** carry completed `agent-review-*` check runs produced by `github-actions`? A commit *moves* the head, so a verdict attested against the head proves the panel ran **and** that nothing has landed since. No timestamp comparison — that would depend on two clocks and on commit dates, which the author controls (`git commit --date`).

Fails toward **ineligible** in every branch (API error, unreadable check list, unknown head, unknown `GITHUB_REPOSITORY`), because this authorises a bot to push to someone's branch. A refusal is an *answer*, not a failure: the job stays green and the reason replaces the placeholder comment.

It deliberately does **not** check the paged latch — a PR the loop gave up on is the main reason to reach for this verb. The latch stays in place, so the loop stays parked; `@claude rerun` remains the verb that clears it.

### 2. The fix/skip report, read by the next panel round

`scripts/agent/fix-report.mjs` posts a readable table plus a hidden payload. The next round folds each claim into the **existing adjudication pass** — not the verifier. `adjudicateFinding` states that the verifier path is the one path never handed author-written text, and a report is author-written text, so it goes to the component built to receive it.

The asymmetry between the two statuses is the safety property:

| status | maps onto | outcome |
|---|---|---|
| `fixed` | `not-present` | Can clear the finding — but only if the adjudicator reads the code and cites locations itself. A false "I fixed it" is upheld. |
| `skipped` | **nothing** | `OVERTURN_GROUNDS` has no entry for "I did not do it" (Phase 28: *undeliverable is not wrong*). Can only be upheld — and skipping the same finding twice trips `upheldTwice` and pages a human. |

"This finding is *wrong*" stays a different claim with a different channel (a rebuttal). Both fixers are told the difference in their prompt. The advisory `@claude review` panel reads reports too, so it reaches the same verdict the gating panel would.

### 3. A separate effort comment

`metrics.mjs effort` posts one comment per fix run under `<!-- agent-fix-effort -->`, which matches neither `METRIC_PREFIX` nor `SUMMARY_MARKER` — so `summarize`'s sweep-and-repost cannot take it. The session is still recorded into the PR-wide ledger as `review-fix`, so the total stays right.

It reports the **outcome**, not just the spend. Both `@claude rerun` attempts on #632 and #648 died at turn 1 in 18 seconds on:

```json
{ "subtype": "success", "is_error": true, "api_error_status": 429,
  "terminal_reason": "api_error", "num_turns": 1,
  "result": "You've hit your session limit · resets 12:10pm (UTC)" }
```

An account quota window — from outside, indistinguishable from a cheap successful round, and the pipeline's only signal was the generic *"the fixer agent failed and the branch head is unchanged"* page, which sends a maintainer looking for a code defect. The comment now separates a **session limit** (wait for the reset, then re-run), a **turn ceiling** (re-running hits it again) and a **transient** (retry now), and renders sub-minute durations in seconds so "it died in 18s" is not hidden behind `formatMinutes`' 1m floor.

### One brief, one builder

The fixer's work list moves out of ~100 lines of inline `github-script` into `scripts/agent/fix-brief.mjs`, shared by both fixers — the lesson `prior-findings.mjs` records, applied to a *prompt* input. That puts its two bounds (40 items / 16k chars) and the `proseOnly` cut under test, and replaces `core.setOutput`'s implicit random delimiter with an explicit one: the value is previous-round model output, so a guessable `$GITHUB_OUTPUT` delimiter would let a finding append step outputs of its own.

### Trust boundary

Everything that **decides** (eligibility) or composes the agent's **prompt** (the brief) runs from the trusted `main` checkout, before the PR branch is checked out over it. Author-side acts run from the branch afterwards. Both fixers now invoke the report/rebuttal CLIs from a staged copy of main's scripts, so a PR branched before a script shipped can still file one.

### Behavioural changes

- A round with **zero failing lens checks** now fails the `fix` job instead of spending an agent run on an empty work list and paging afterwards. Same destination, one run cheaper. The other zero-finding case (failing lenses whose `output.text` doesn't parse) is unchanged — the honest pointer, bodies uncut.
- `@claude fix` on an **issue** still starts `agent-implement.yml`; the two gate on opposite sides of `github.event.issue.pull_request`, asserted by a test.

## Test plan

- `pnpm verify:fast` — green.
- `pnpm verify:entropy` — green.
- `cd scripts/agent && node --test *.test.mjs` — **772 passing**, 1 skipped.
- New: `fix-brief.test.mjs` (19), `fix-eligible.test.mjs` (15), `fix-report.test.mjs` (22), plus 7 in `metrics.test.mjs` and 5 workflow-invariant tests in `checks.test.mjs`.
- **Mutation-checked**, not just green: moving the eligibility gate after the branch checkout, restoring the inline checklist builder, negating the PR-surface guard, dropping the effort step, moving `defence()` out of `buildAdjudicatorPrompt`, removing the TOCTOU re-check, reverting the ` -->` escape, reverting the rebuttal-precedence rule, reverting `classifyFixResult`, and using all reports instead of the latest — each fails the intended test.
- Both CLIs exercised end-to-end against a stub `gh`, including the real `$GITHUB_OUTPUT` heredoc emission and the actual 429 execution log.

## Defects found and fixed before this was opened

Recorded because they are the interesting part. Each is covered by a test that fails if the fix is reverted.

Found while building:

- The tie-refusal tests passed **vacuously** — `findingSimilarity` scores two-word summaries at 0, so the matcher never entered its loop.
- The fork check **failed open** when `GITHUB_REPOSITORY` was unset: it skipped the comparison instead of refusing.
- The refusal message told people to run `@claude review` to become eligible — that verb is advisory and records **no** check runs, so following it could never work.
- The adjudicator fence holds here only because `defence()` lives in `buildAdjudicatorPrompt`, not the serializer, which `toRebuttalRecords` bypasses.

Found by an independent adversarial review of the finished diff:

1. **Every *successful* fix run reported FAILURE.** `classifyResult`'s success arm requires `structured_output`, an Agent-SDK field a claude-code-action transcript never carries — so a clean run came back `failed (no-output) — subtype=success. Not retryable; a human should take a look.` The happy-path test hand-wrote `{ok:true}` and so could never have caught it.
2. **Folding reports into `rebuttals` killed the rebuttal loop.** A disputed finding always produces both a rebuttal and a report item with identical lens/file/summary, and `matchRebuttal` refuses a tie — so the grounded rebuttal was never adjudicated, and `upheldTwice` could never fire. The documented "skipping twice pages a human" guarantee was false.
3. **One item quoting ` -->` discarded the entire report.** `JSON.stringify` doesn't escape it, the non-greedy parser stopped there, and the round-trip guard then posted *nothing at all*.
4. **Adjudication cost was unbounded** — one 20-turn Opus session per still-gating finding, inside a 45-minute job.
5. **Eligibility had a TOCTOU** — the gate pinned a sha, the checkout took the branch tip minutes later.
6. A failed gate step stranded the placeholder forever, and the "passed all N lenses" refusal over-counted `neutral` lenses.

## Not built

The panel's rendered summary does not surface "N findings the author reported as skipped" — the fix agent's own comment is the human-visible record, in the same thread. And `MAX_REVIEW_ROUNDS` counts only autonomous rounds, so an on-demand fix is outside that budget by design.

## Summary by CodeRabbit

* **New Features**
  * Added a maintainer-only `@claude fix` command for eligible pull requests.
  * Fixes now require a completed review with actionable failing checks.
  * Added structured reporting for fixed and skipped findings, including follow-up adjudication.
  * Added execution summaries with outcome, effort, duration, and diagnostic details.
* **Improvements**
  * Review results now incorporate fix reports and preserve unresolved skipped findings.
  * Added safeguards for stale branches, unauthorized requests, malformed reports, and failed executions.
* **Documentation**
  * Documented the on-demand fix workflow and its safety rules.

### DIFF
--- a/.github/workflows/agent-fix.yml
+++ b/.github/workflows/agent-fix.yml
@@ -241,16 +241,26 @@ jobs:
           ref: ${{ steps.pr.outputs.branch }}
           fetch-depth: 0
           token: ${{ steps.app-token.outputs.token }}
-      - uses: pnpm/action-setup@v4
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
         if: steps.eligible.outputs.eligible == 'true'
       - uses: actions/setup-node@v4
         if: steps.eligible.outputs.eligible == 'true'
         with:
           node-version: 22.x
           cache: pnpm
+      # --ignore-scripts because the workspace is the UNTRUSTED PR branch and an App
+      # token is already in the environment: the branch's own install lifecycle
+      # scripts must not run. The only two are the root `postinstall` (`git config
+      # core.hooksPath .githooks`, unwanted here anyway — we do not want the branch's
+      # hooks) and packages/backend's (`prisma generate`), which `pnpm backend test`
+      # inside verify:fast needs, so it is run explicitly below — for the first-party
+      # backend package only, not for every dependency's postinstall.
       - name: Install dependencies
         if: steps.eligible.outputs.eligible == 'true'
-        run: pnpm install --frozen-lockfile
+        run: pnpm install --frozen-lockfile --ignore-scripts
+      - name: Generate the Prisma client (skipped postinstall)
+        if: steps.eligible.outputs.eligible == 'true'
+        run: pnpm --filter @wafflebase/backend exec prisma generate
 
       # Read the remote tip before the agent runs, so the step after it can tell
       # "fixed and pushed" from "ran and pushed nothing". The REMOTE, not local
@@ -476,6 +486,15 @@ jobs:
             if (neverRan) {
               body = '🛑 `@claude fix` could not start the fix agent — a setup step failed before it '
                 + `ran, so nothing was changed. [Workflow run](${process.env.RUN_URL})`;
+            } else if (advanced && failed) {
+              // The commit landed but the agent then died — e.g. during the
+              // `fix-report.mjs post` call. Line ~500 sets the job red, so a plain
+              // "✅ pushed" would sit next to a red X and point at a report comment
+              // that was never filed. Report the partial outcome honestly instead.
+              body = `⚠️ \`@claude fix\` pushed \`${after.slice(0, 8)}\` to \`${branch}\`, then failed before `
+                + 'it finished. The commit landed, but its fix report may be missing or incomplete, so the '
+                + 'next panel round cannot credit the claims. See the effort comment above and the '
+                + `[workflow run](${process.env.RUN_URL}).`;
             } else if (advanced) {
               body = `✅ \`@claude fix\` pushed \`${after.slice(0, 8)}\` to \`${branch}\`. `
                 + 'CI and a fresh review panel will run on it. See the fix agent\'s report comment '
--- a/.github/workflows/agent-fix.yml
+++ b/.github/workflows/agent-fix.yml
@@ -154,12 +154,19 @@ jobs:
       # Not eligible → say why and stop. A refusal is an ANSWER, not a failure:
       # the job stays green so the reason is what the maintainer sees, rather than
       # a red X that reads as a broken workflow.
+      # `always()`, NOT the implicit `success()`. `fix-eligible.mjs` exits 2 on a
+      # broken invocation (unreadable lenses.json, missing gh, unset
+      # $GITHUB_OUTPUT), which fails the step — and with the default condition
+      # this was skipped along with everything downstream, leaving the PR showing
+      # "🤖 Working on @claude fix…" forever next to a red X. A refusal must always
+      # reach the person who asked.
       - name: Explain the refusal
-        if: steps.eligible.outputs.eligible != 'true'
+        if: always() && steps.eligible.outputs.eligible != 'true'
         uses: actions/github-script@v7
         env:
           COMMENT_ID: ${{ steps.placeholder.outputs.id }}
           REASON: ${{ steps.eligible.outputs.reason }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         with:
           github-token: ${{ steps.app-token.outputs.token }}
           script: |
@@ -172,7 +179,10 @@ jobs:
             await github.rest.issues.updateComment({
               owner: context.repo.owner, repo: context.repo.repo,
               comment_id: Number(process.env.COMMENT_ID),
-              body: `⏸️ \`@claude fix\` did not run.\n\n${process.env.REASON}`,
+              // REASON is empty when the gate step itself failed rather than
+              // returning a decision — say that, instead of posting a blank refusal.
+              body: `⏸️ \`@claude fix\` did not run.\n\n${process.env.REASON
+                || 'The eligibility check could not complete — see the [workflow run](' + process.env.RUN_URL + ').'}`,
             });
 
       # The fixer's work list, from the SAME builder the autonomous loop uses.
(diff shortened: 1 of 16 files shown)


harrykim8672:
@claude review

hwisoo-kim:
Addressed all nine findings — CI is green (verify-self / verify-browser / verify-integration / codecov all pass). Each fix is mutation-checked: reverting it fails the test that covers it.

**CI failure (`agent:tests`) — my own test was environment-dependent.** It passed `repo: undefined` to `readPrHead`, and a default parameter substitutes for `undefined` — so the call silently became "whatever `GITHUB_REPOSITORY` happens to be": green locally where it is unset, red in Actions where it is `wafflebase/wafflebase`. It now clears and restores the variable, so it tests the production path deterministically in both. Verified by running the suite under both env states.

**`agent-review-on-demand.yml` — valid, and worse than it looks.** Fixed to `needs.authorize.outputs.pr`, and the tolerance comment now names the third flag. This wasn't cosmetic: the reader was copied from `agent-review-panel.yml`, where the number comes from a `pr` *step* this workflow doesn't have, so `steps.pr.outputs.number` was the empty string, the `if` was permanently false, and **the step never ran** — no `/tmp/fix-reports.json`, `review-panel.mjs` tolerating the missing file, and the advisory panel silently ignoring every fix report. Fail-safe and completely invisible. That is exactly the second-copy rot `fix-brief.mjs` was extracted to prevent, reproduced in the plumbing around it, so there is now a test asserting the on-demand reader resolves to an output this workflow actually produces.

**`review-panel.mjs` `applySkipClaims` lens — valid, and it's a pre-existing bug, not a new one.** `findingSimilarity` returns 0 unless both sides agree on `lens`. Prior findings carry it (stamped by `tagPriorFindings` from the check-run name); a fresh finding is raw model output and `coerceFindings` never adds one — and when a fresh finding and its carried-forward twin merge, the **fresh** representative survives. So any defect the fresh pass re-found — i.e. any defect still present, the common case — reached adjudication with no `lens` and could never match a rebuttal. **Phase 28's rebuttal→adjudication loop has been inert for exactly those findings since it shipped**, and the fix-report claims inherited it. Confirmed by reproduction before and after.

Fixed differently from the suggestion, for a reason worth recording: the stamp goes on `merged`, not on the gating subset passed to `applySkipClaims`/`adjudicateRebuttals`. `mergedAfter` removes overturned findings from the summary by object **identity** (`overturnedOut.has(f)`), and `gatingFindings` filters without copying — so copying at the gating step would make every dropped finding un-removable from the summary. One stamp upstream fixes both the skip path and the rebuttal path with identities intact.

Worth noting: my first draft of that line referenced an undefined `str`, inside an arrow that `import` alone never evaluates — the same shape as #657's `retryAt`. `no-undef` caught it, which is the rule #658 existed for.

**`fix-eligible.mjs` check order — valid.** `readPrHead` returns `{head: "", isFork: true}` on a failed lookup, because fork is the safe default for unknown provenance. Checking `isFork` first therefore told a maintainer "this is a fork" whenever the API merely failed, sending them to look for a fork problem on a same-repo PR. Both outcomes refuse, so no decision changes — only whether the stated reason is true.

**Nitpicks — all four valid, all fixed:**
- `checks.test.mjs`: extended, as described above.
- `fix-brief.mjs`: the header claimed a `findings` step output that is never written. Corrected in both the header and the `buildBriefFrom` doc.
- `fix-report.mjs` JSDoc: correct — my own insertion left the `toRebuttalRecords` rationale orphaned above `MAX_FIX_ADJUDICATIONS`, documenting nothing. Moved onto the function, and its `skipped` bullet was also **stale**: those claims no longer reach adjudication at all, they take the session-free uphold path.
- `fix-report.mjs` line 93: correct — the comment had lost its own `-` and read as "`-` is the JSON escape for `-`". Restored; implementation untouched.

**`harness-engineering.md` code span** — fixed by rewording to "space-then-`-->` terminator", which removes the leading space from the span while staying accurate about what the parser actually splits on (the terminator does include the space).

Nothing skipped.

hwisoo-kim:
Split the lens-match fix out into #674, as it changes the gating path on every PR and deserves its own review and bisect point.

**#674 must merge first.** This branch is now rebased on top of it and no longer carries the stamp — but it still *depends* on it: fix-report claims are matched by the same `findingSimilarity` gate, so without #674 the reports here match nothing. Fail-safe (every finding stands), but the feature is inert.

The diff against #674's branch is this PR's own work only — `review-panel.mjs` here contains no stamp lines. Rebased, `verify:fast` green, 845 agent tests passing.

While splitting it out I found the reason the loop was never *observed* to be broken: no structured rebuttal has ever been filed on any PR, and part of why is that `rebuttal.mjs` landed with #633, so branches cut before it (#632, #648) don't contain the CLI their fixer prompt tells them to run. That half is fixed here by staging main's scripts into `runner.temp`; #674 fixes the matching half. Details in #674.

harrykim8672:
@claude review

PR #674, opened 2026-08-05
Stamp the lens on findings, so a rebuttal can be matched at all

hwisoo-kim:
## Summary

The rebuttal→adjudication loop (#633) has been **inert since it shipped**, for exactly the findings it exists to serve. Split out of #672, where I found it, so it gets its own review and its own bisect point — it changes the gating path on every PR.

`findingSimilarity` returns 0 unless both sides agree on `lens`, and it is the gate `matchRebuttal` scores through:

```js
if (trimmed(a.lens) !== trimmed(b.lens)) return 0;   // rounds.mjs:303
```

But `lens` is **not in the `FINDING` schema** — a lens is never asked for it — and nothing in `review-panel.mjs` assigned one. Prior findings *do* carry it (`tagPriorFindings` stamps it from the check-run name), and when a fresh finding merges with its carried-forward twin the **fresh** representative survives. So the lens was lost, and the outcome inverted the intent:

| round N+1 state | matched | adjudicator |
| --- | --- | --- |
| fresh pass re-found it (± carry-forward) | 0 | **never called** |
| carry-forward only (fresh pass missed it) | 1 | called |

Only the second row worked, and it is the one the loop was **not** built for: a rebutted finding means the author did *not* change the code, so the next round's fresh pass almost always re-finds it.

Reproduced against `main`'s own code before fixing:

```
gating findings: 1 | lens on it = undefined
adjudicator invoked: 0 time(s)
tally: {"matched":0,"overturned":0,"upheld":0,"errored":0}
```

## Why nothing observable ever broke

The channel was never exercised. **No structured rebuttal has ever been filed on any PR** — #564, #605, #632, #633, #648, #649, #657, #666 all have zero `<!-- agent-rebuttal -->` comments.

A second defect explains part of that: `scripts/agent/rebuttal.mjs` landed *with* #633, so branches cut before it do not contain the CLI their fixer prompt tells them to run. #632 (`agent/625-notes-mermaid-preview`) and #648 (`agent/586-cli-system-exit-code`) both 404 on that path. That half is fixed separately in #672, which stages main's scripts into `runner.temp`.

Both defects fail safe — the finding stands — which is why this was latent rather than harmful.

## Where the stamp goes, and why it matters

`stampLens` fills the blank. Two constraints dictated the placement:

- **It stamps `merged`, not the gating subset.** The round loop then removes overturned findings from that same array by object **identity** (`overturnedOut.has(f)`), and `gatingFindings` filters without copying. A copy made at the gating step would leave every overturned finding un-removable from the summary — dropped from the check run and still rendered as blocking for a human.
- **Blanks only, never an overwrite.** `priorForLens` is already filtered to `p.lens === lens.id`; overwriting would let one lens claim another's finding and silently re-target its rebuttals.

## Test plan

- `pnpm verify:fast` — green. `pnpm verify:entropy` — green.
- `cd scripts/agent && node --test *.test.mjs` — 758 passing, 6 skipped (SDK-absent lane), green with `GITHUB_REPOSITORY` both set and unset.
- Four new tests assemble the **real** shape — fresh pass + carry-forward → merge → gate → `adjudicateRebuttals` with an injected adjudicator — because both halves passed their own unit tests while the loop was inert end to end. One asserts the pre-fix behaviour explicitly, so the fix cannot regress into "it never matched anyway".
- **Mutation-checked.** All three fail a test: removing the call from the round loop, copying unconditionally (breaks the `mergedAfter` identity), and overwriting a foreign lens.

## One note on the tests

The first draft defined a local copy of the stamp expression and every test passed **with the production line deleted** — they were exercising the test's own helper. That is why `stampLens` is a named export rather than an inline `.map`, and why there is a source-level test asserting the round loop routes `merged` through it: that loop lives in `main()`, where no unit test reaches.

## Interaction with #672

#672 depends on this. Its fix-report claims are matched by the same `findingSimilarity` gate, so until this lands its `@claude fix` reports match nothing. **Merge this first**; #672 is rebased on top of it and I've removed the duplicate stamp from that branch.

Unrelated, but noticed while writing this: `docs/design/harness-engineering.md` has two sections numbered "Phase 28" (UI Issue Hunting and Structured Rebuttal). Left alone here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved review adjudication so newly identified and carried-forward findings are correctly attributed to the active review perspective.
  - Rebuttals are now matched only to findings from the same perspective.
  - Restricted rebuttal processing to trusted review-agent submissions.
  - Added safeguards against crafted finding content altering adjudication instructions.

- **Documentation**
  - Added guidance describing structured rebuttal adjudication and review-finding attribution.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1575,6 +1575,42 @@ visible only in the check body today), and any measurement of how often a
 rebuttal is *right* — which is a `misses.jsonl` question, since an overturn that
 should not have happened is a false negative like any other.
 
+#### The loop shipped inert, and stayed inert until the lens was stamped
+
+`findingSimilarity` returns 0 unless both sides agree on `lens`, and it is the gate
+`matchRebuttal` scores through. But `lens` is not in the `FINDING` schema — a lens
+is never asked for it — and nothing in `scripts/agent/review-panel.mjs` assigned
+one. Prior findings do carry it (`tagPriorFindings` stamps it from the check-run
+name), and when a fresh finding merges with its carried-forward twin the FRESH
+representative is the one that survives. So the lens was lost in exactly the case
+that matters, and the outcome inverted the intent:
+
+| round N+1 state | matched | adjudicator |
+| --- | --- | --- |
+| fresh pass re-found it (± carry-forward) | 0 | never called |
+| carry-forward only (fresh pass missed it) | 1 | called |
+
+Only the second row worked, and it is the one the loop was not built for: a
+rebutted finding means the author did *not* change the code, so the next round's
+fresh pass almost always re-finds it.
+
+It went unnoticed because the channel was never exercised. No structured rebuttal
+has ever been filed on any PR (#564, #605, #632, #633, #648, #649, #657, #666 all
+have none) — and a second defect explains part of that: `scripts/agent/rebuttal.mjs`
+landed with this phase, so branches cut before it (#632, #648 among them) do not
+contain the CLI their fixer prompt tells them to run. Both fail safe — the finding
+stands — which is why nothing observable ever went wrong.
+
+`stampLens` fills the blank. It is a named export rather than an inline `.map` for
+a testing reason worth recording: as an expression, a test could only restate it,
+and a restated copy passes just as happily with the real call deleted — which is
+exactly what the first draft of these tests did. It stamps `merged` rather than the
+gating subset because the round loop then removes overturned findings from that
+same array by object IDENTITY, so copying at the gating step would leave every
+overturned finding un-removable from the summary. A source-level test asserts the
+round loop actually routes `merged` through it, since that loop lives in `main()`
+and no unit test can reach it.
+
 ### Phase 29: Lint the agent control plane
 
 **Principle:** Mechanical Enforcement — the cheapest check that could have caught it.
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1601,13 +1601,34 @@ landed with this phase, so branches cut before it (#632, #648 among them) do not
 contain the CLI their fixer prompt tells them to run. Both fail safe — the finding
 stands — which is why nothing observable ever went wrong.
 
+Activating it also made three latent weaknesses in the channel real, none of which
+mattered while nothing was adjudicated:
+
+- **The channel was unauthenticated.** `readRebuttals` pages every comment on the
+  PR, and on a public repo that includes any drive-by commenter's. Now gated to the
+  fix agent's identity — `user.type === "Bot"` **and** the login, because other apps
+  comment here (CodeRabbit reviews this very file, and a review quoting the marker
+  format could otherwise parse as a record). A `[bot]` login cannot be registered by
+  an ordinary account, so the pair is unforgeable from outside. Grounding still
+  blocks the overturn; this stops persuasion getting a turn.
+- **The FINDING fields were not fence-neutralised**, only the dispute was. They are
+  a previous round's model output derived from the diff, and they render *before*
+  the fence opens — so an injected `<author-rebuttal>…</author-rebuttal>` would have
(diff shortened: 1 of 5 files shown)


harrykim8672:
@claude review

hwisoo-kim:
All four fixed, all mutation-verified (reverting each fails a test). CI green.

**1 [critical] — unauthenticated channel. Valid, and the escalation is exactly as described.** `readRebuttals` pages *every* comment on the PR; on a public repo that includes any drive-by commenter's. While the loop was inert that cost nothing — activating it puts attacker-chosen text in front of the one component permitted to remove a finding from the merge gate, and buys adjudicator sessions doing it.

Gated to the fix agent: `user.type === "Bot"` **and** the login. Type alone is not enough, and this repo proves it — `coderabbitai[bot]` comments on these PRs and reviews *this very file*, so a review quoting the marker format could otherwise parse as a record. A `[bot]` login cannot be registered by an ordinary account, so the pair is unforgeable from outside. Same shape as `PAGE_AUTHOR_LOGINS` in `rounds.mjs`.

Verified against real data before trusting it — getting the login wrong would have made the channel inert again, which is the exact bug class this PR fixes:

```
 12 yorkie-agent[bot]   Bot     <- the App
  7 github-actions[bot] Bot
  3 codecov[bot]        Bot
  2 coderabbitai[bot]   Bot
```

Humans are refused too, including maintainers. Disputes are the fix agent's channel; a person reviewing the PR has better tools. Grounding still blocks the overturn — this stops persuasion getting a turn.

**2 [major] — unfenced FINDING fields. Valid.** The finding block renders *before* the fence opens, and its fields are a previous round's model output derived from the diff — so a contributor can get chosen text quoted into `summary`/`evidence`. Unneutralised, an injected `<author-rebuttal>…</author-rebuttal>` would have placed a complete fake dispute ahead of the real one. `defence()` now applies to `lens`/`file`/`severity`/`summary`/`evidence`; the test asserts exactly one opening and one closing tag survive.

**4 [major] — fill-blanks-only re-opened origin spoofing. Valid, and this one is mine.** `prior-findings.mjs:104` carries the comment explaining why `lens` goes **last** — "a finding cannot spoof its own origin by carrying a `lens` key, since these come from a previous round's model output" — and I cited that file while doing the opposite. A fresh finding is model output too; `lens` is absent from the FINDING schema but nothing rejects an extra key, and that field decides which rebuttals `findingSimilarity` matches. It now overwrites.

The identity constraint I'd used to justify the weaker rule doesn't apply: what callers depend on is *where* the stamp runs (as `merged` is built), not whether it copies. Corrected the comment and the test that asserted the wrong behaviour.

**3 [major] — implicit lens partition. Valid, fixed, with a caveat worth stating.** `adjudicateRebuttals` now filters by `lensId` itself. Once finding 4 overwrites, `finding.lens` is trustworthy and `findingSimilarity`'s lens gate already enforces the partition — so this is defence-in-depth, making it a property of the function rather than an invariant every caller maintains.

That made it hard to test honestly: a foreign rebuttal against a correctly-stamped finding fails either way, and my first test passed with the guard deleted. The distinguishing case is robustness to the **caller** — a finding and rebuttal that agree with each other but not with `lensId`. Similarity is happy; only the partition refuses. That's what it asserts now, with a positive control.

---

**One thing this surfaced beyond the four findings.** The same critical hole existed in #672's fix-report channel — and worse: one report carries up to 80 items where a rebuttal carries one, an 80x amplification. I'd been told about it in an earlier review round and it never landed. Now closed in #672 by reusing `fromRebuttalAuthor` rather than re-deriving it, since both channels have exactly one legitimate writer and two copies of "who may write" is how one ends up wrong.

Also worth noting for the record: no structured rebuttal has ever been filed on any PR, so none of these three weaknesses has been exercised. All fail safe — the finding stands.

851 agent tests, `verify:fast` and `verify:entropy` green.

harrykim8672 (approved):

PR #678, opened 2026-08-05
Give the UI hunter a live session and one MCP tool

hwisoo-kim:
## Summary

PR 3 of the UI hunter (Phase 28). PR 1 (#642) shipped the executor, harness and
oracles; PR 2 (#665) the prediction protocol. This one makes the machinery
drivable by a model: a persistent browser session and the single MCP tool that
exposes it.

**A persistent session, because the numbers demanded one.** The plan assumed a
spawn-per-action tool, matching the CLI hunter. Measuring retired that: a
1-action plan takes 6095ms and a 3-action plan 6103ms, so Vite plus Chromium
boot is ~6.1s and each subsequent action ~4ms. At `maxActions: 80` that is ~8
minutes of pure boot against a whole-run probing budget of ~15 minutes. So
`hunt-ui-runner.mjs` gains a `--serve` mode — newline-delimited JSON, one
response per request, one page for the session — and `hunt-ui-session.mjs` is
the client. Measured through it: ready 856ms, first action 5.4s, later actions
33–44ms. Serve and plan modes share one `observeAction`, extracted rather than
duplicated.

**Replay deliberately does not use it.** `runUiPlan` keeps its `spawnSync` path
against `--plan`, so the determinism gate still gets a fresh process and a fresh
browser context per attempt. Exploration is a session; replay is a clean room.
One mechanism for both would mean either a slow explorer or a replay that
inherits state, and the second is how phantom repros get through.

**The tool description is the map.** PR 1's readers were reachable but
undiscoverable — nothing told a model that `doc.fontSizes` existed, so an agent
asked about formatting would reach for the snapshot and read an almost-empty
accessibility tree off a canvas. The tool now lists the readers, with arity,
scoped to the run's surface. That list is a second copy of the bridge registry,
so a test parses `bridge.ts` and fails on any divergence: a stale copy is
silent, because a reader never called is indistinguishable from an area with no
defects.

**The per-run surface selector is enforced, not requested.**
`assertSafeActionPlan` bounds reader *namespaces*, which is the right check for
it to make and not sufficient here. The tool checks exact names at three doors —
the action's own reader, a click target's `reader` (a point comes from
`sheet.cellCenter`, not from coordinates), and `expect.read` — plus `goto`'s
surface, or an agent could navigate out of its assignment and then read legally.

**What comes back is less than what happened.** A non-`read` action reports only
whether it landed; a prediction reports its verdict and never the measured
`actual`; no page snapshot is returned unless `dom.snapshot` was read by name.
Handing back the value invites re-describing a violated prediction as some
weaker claim that happens to fit it, which is the rationalisation the
same-round-trip design exists to prevent.

Also fixes a **literal NUL byte** in `hunt-ui-probe.mjs` (from #642) that made
the file binary to `grep` and `git diff` — `grep '^export'` printed nothing at
all, which is how the wrong export names nearly got built against here. The
escape sequence produces the identical string, so no fingerprint changes.

Nothing is filed automatically. The output is a local report; the orchestrator,
personas and first live run are PR 4.

## Test plan

- `verify:self` — all 11 lanes green in 133s, including `agent:tests` (2.0s) and
  `verify:frontend:chunks` (the hunt route stays DEV-gated, production still 147
  chunks).
- `agent:tests` — 902 pass, 0 fail. 34 new cases for the tool, 13 for the
  session, both driven through injected stubs so neither needs a browser (the
  lane runs with `scripts/agent/node_modules` absent, so every third-party
  import stays lazy).
- `verify:hunt:oracles` — 10 oracle checks plus cell targeting and undo
  capability, all pass.
- `verify:entropy` — 0 issues; the design doc's new refs resolve.
- **Mutation testing: 22 mutants across both new modules, all 22 caught.**
  Including the two that matter most: `atIndex` off by one (which reopens the
  self-reference finding generator) and `redactSecrets(text, secrets)` with the
  old array signature, which silently delivers no secrets at all. The second one
  survived the first round — the test had used a `wfb_`-shaped value that
  `redactSecrets`' own generic pattern caught regardless, so it never verified
  that the run's live key was wired in. It now uses an opaque value.
- **Live end-to-end run** through the real runner, real Vite, real Chromium:
  ready 2.9s, `goto doc` 5.6s, reads 44–60ms. The surface pin refused
  `sheet.cellValue` on a doc run and listed what *was* available; the vocabulary
  bound refused `evaluate`; a self-referencing prediction came back
  `unevaluable`. One grounded ground-A candidate surfaced — clicking "Increase
  font size" with no selection leaves `doc.fontSizes` unchanged — and was
  reported for investigation rather than auto-filed, which is the intended
  behaviour.

## Summary by CodeRabbit

* **New Features**
  * Added persistent UI exploration sessions to reduce repeated browser startup.
  * Added scoped UI exploration with browser actions, navigation, predictions, notes, and on-demand observations.
  * Added concise action and prediction responses, with detailed results available when requested.
  * Added validation, timeouts, graceful shutdown, and malformed-request handling.

* **Bug Fixes**
  * Improved resilience when actions fail, runners crash, or protocol messages are malformed.

* **Documentation**
  * Documented persistent serve mode and isolated replay workflows.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1426,8 +1426,48 @@ state instead (`boundingBox`, `combinedBoundingBox`, `framesApproxEqual` are alr
 exported). Renderer bugs — model right, paint wrong — remain the visual lane's job, which
 already owns 220 baselines with Docker font pinning. Two instruments, two failure classes.
 
+**Exploration is a session; replay is a clean room.** The CLI hunter spawns a fresh
+process per probe because that costs milliseconds. Measuring the same shape here retired
+it: a 1-action plan takes 6095ms and a 3-action plan 6103ms, so Vite plus Chromium boot is
+~6.1s and each subsequent action ~4ms. At `maxActions: 80` a spawn-per-action tool would
+spend ~8 minutes on boot alone, against a whole-run probing budget of ~15 minutes. So
+`packages/frontend/scripts/hunt-ui-runner.mjs` gained a `--serve` mode — newline-delimited
+JSON, one response per request, one page for the session — and `scripts/agent/hunt-ui-session.mjs`
+is the client (measured through it: ready 856ms, first action 5.4s, later actions 33-44ms).
+Both modes share one `observeAction`, extracted rather than duplicated.
+
+Replay deliberately does NOT use it. `runUiPlan` keeps its `spawnSync` path against
+`--plan`, so the determinism gate still gets a fresh process and a fresh browser context
+per attempt. Sharing one mechanism would mean either a slow explorer or a replay that
+inherits state, and the second is how phantom repros get through. A failed action in serve
+mode is an observation with `ok:false`, never a protocol error, because "the click missed"
+is data — and the process stays up through a malformed request, since killing the browser
+over one bad line discards the boot this mode exists to amortise.
+
+**The tool description is the map.** PR 1's readers were reachable but undiscoverable:
+nothing told a model that `doc.fontSizes` existed, so an agent asked about formatting would
+reach for the snapshot and read an almost-empty a11y tree off a canvas.
+`scripts/agent/hunt-ui-tool.mjs` therefore lists the readers, with arity, SCOPED TO THE
+RUN'S SURFACE. That list is a second copy of the bridge registry, so a test parses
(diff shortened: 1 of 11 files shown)


hackerwins:
@coderabbitai review this.

hackerwins:
@coderabbitai review this.

harrykim8672 (approved):

PR #679, opened 2026-08-06
Clear 35 of 39 Dependabot alerts across both lockfiles

hackerwins:
## Summary

39 open Dependabot alerts, spanning 13 packages and 4 kinds of fix. All
were transitive (lockfile) findings — 34 in `pnpm-lock.yaml`, 5 in
`scripts/agent/package-lock.json`. One commit per fix class:

1. **Refresh stale `pnpm.overrides` (27 alerts)** — brace-expansion,
   undici, fast-uri, postcss, js-yaml, tmp and shell-quote were still
   pinned at the previous sweep's patch level, so newer advisories on the
   same packages matched again. Adds two new pins for body-parser.
2. **Update the agent-scripts lock (5 alerts)** — hono, ip-address and
   @hono/node-server. `npm update` alone clears them:
   `@modelcontextprotocol/sdk` 1.30.0 widens its `@hono/node-server` range
   to `^1.19.9 || ^2.0.5`, so the patched v2 line resolves without forcing
   an override.
3. **Bump react-router 7.15.1 → 7.18.2 (4 alerts)** — open redirect,
   constructor injection, unauthenticated DoS, RSCErrorHandler protocol
   validation. The only direct-dependency move in this PR.

## Not fixed here

- **Alert #134** (react-router RSC-mode CSRF) is patched only in 8.3.0
  with no 7.x backport. The frontend mounts a plain `BrowserRouter` and
  imports no RSC API, so the path is unreachable — dismissed as not
  applicable rather than taking a major upgrade.
- **vite ×3** remain via `vitepress@1.6.4`, which pins `vite ^5.4.14`.
  The advisories are ranged `<= 6.4.2` with no lower bound, so 5.4.21
  still matches even though the app is on 6.4.3. Docs-site build only,
  not app runtime; same conclusion as the 2026-06 sweep.

## Test plan

- `pnpm verify:fast` green after each commit
- `pnpm frontend build` green after the react-router bump
- `npm test` in `scripts/agent` — 942 tests pass; `npm audit` clean
- Cross-checked every alert's `vulnerable_version_range` against the
  versions actually resolved in both lockfiles via `semver.satisfies`:
  35 of 39 no longer match, the remaining 4 are the exceptions above

## Summary by CodeRabbit

* **Security & Maintenance**
  * Updated third-party dependencies to address reported security vulnerabilities and enforce safer minimum versions.
  * Improved dependency consistency across the application and supporting tools.
  * Verified resolved alerts and documented remaining items requiring follow-up.

* **Documentation**
  * Added guidance for handling dependency alerts, validating lockfile updates, and refreshing generated application assets.
  * Updated the active task index with recent dependency maintenance work.


hackerwins:
@coderabbitai review this.

PR #680, opened 2026-08-06
Hash the full lens configuration in config_hash

dlgpdmsly2:
## The problem

The panel is tuned by editing `lenses.json` — a model, a sample count, a reasoning effort, which file classes a lens reads. **Nothing here can answer "what settings produced that review?"** A check run records a verdict; the manifest behind it is whatever was on `main` at the time, and two tuning commits later that is not recoverable.

This lands the fingerprint that answers it, and the machinery to freeze a config and load it back. A fingerprint has two ways to be wrong and they are not symmetric — hashing the **same** when behaviour differs merges two reviewers with no symptom anywhere; hashing **differently** when behaviour is identical splits them visibly. The logic it replaces (from the fork eval harness, never upstreamed) was wrong in both directions, six times.

| # | Where | What | Direction |
|---|---|---|---|
| 1 | `normalizeLens` | omits `scopeClasses` | **false negative** — a code-only lens and an everything lens hash identically |
| 2 | `normalizeLens` | omits `effort` | **false negative** — two lenses differing only on the panel's main cost dial hash identically |
| 3 | `normalizeLens` | hand-mirrors the panel's `samples` default | **false positive** — `2.5` hashed as `2.5` where the panel floors to `2`; `1e999` as `Infinity` where the panel gives `2` |
| 4 | `buildConfig` | rebuilds each lens from a seven-field allowlist — drops `scopeClasses` **and** `effort` going in | silent loss |
| 5 | `materializeLenses` | same allowlist, drops both again coming out | silent loss |
| 6 | `normalizeLens` | omits `maxTurns` — **found by the guard this PR adds** | **false negative** |

Five of the six are one mistake: **a field was added and nobody updated a hardcoded list.** And 4 and 5 are the same bug at the two ends of one round trip, which is why neither was caught — the old test asserted that materialising a snapshot and rebuilding it gave the original hash, and it passed. **A round trip cannot see a field neither of its ends carries.**

### Two of these cost something

**The routing strip.** `scopeClasses` dropped means `lensScope` sees an omission, which it treats as EVERYTHING by design — so every replayed lens read the **whole PR diff** instead of its file-class slice.

**The cost dial.** A materialised lenses dir carried no `effort` on any lens, so the panel saw `undefined`, `assertEffort` passed it (unset is legal), and all six lenses ran at the SDK default `high`. Production runs **five of six at `medium`**. Materialising the real manifest through both versions:

```
before -> [high,   high, high,   high,   high,   high  ]
after  -> [medium, high, medium, medium, medium, medium]   (security keeps no effort key)
```

That is **the exact silent cost regression `assertEffort` was written to prevent** — *"no error, no changed output, nothing in the logs"* — arriving through a door it cannot watch. It catches a **wrong** value; it cannot catch one **deleted upstream of it**. A validator only guards the door it stands in.

## The change

- `normalizeLens` hashes the **effective** value of every field the panel reads, not the raw manifest value. `appliesWhen` already worked this way; the rest now do.
- **The panel's defaults arrive by import**, not by copy: `FILE_CLASSES` and `sampleCountFor` from `review-panel.mjs`. A hand-copied default that drifts from the panel *is* the false negative this exists to prevent. The import costs nothing — `ask.mjs` loads the SDK lazily, so every new test passes with no `node_modules` at all.
- **Both directions of `config-build.mjs` copy the whole lens object.** The seven-field allowlists are gone; what remains is a denylist of three derived keys. Adapters widen, never narrow.
- **The guard**, which is the actual deliverable: three assertions hold the hashed-field list against the real `lenses.json`, against the fields `review-panel.mjs` actually *reads* off a lens, and against what `normalizeLens` emits. Excluded fields carry a written reason **as data**, so a test can require one. Deliberately not a golden hash string — that goes red on every legitimate change and teaches people to update the constant.
- Smaller: `snapshot.captured_at` is injectable (byte-reproducible snapshots); `--sdk-version` reads the pin from `scripts/agent/package.json` instead of a duplicated literal, and refuses a range rather than record a version that does not identify a build.

`config_hash` covers the lens composition, not the panel's code. The other half of that key already exists as `panelSha` in every capture's `meta.json` (#673), so the pooling key is the **pair** — no hand-maintained `pipeline_version`, for the same reason this PR exists.

`review-panel.mjs` is untouched beyond importing two already-exported symbols. `lenses.json` is byte-identical. No lens's behaviour changes. No model is invoked and nothing is spent.

## Verification

| | |
|---|---|
| the `agent:tests` lane's own command | **1049 tests · 0 fail**, against **1002** on `main` — exactly the 47 new ones |
| #677's `eval/test-lane.test.mjs` | passes with both new suites present, so they genuinely run in CI |
| `npx eslint scripts` | exit 0 at the lockfile's pinned `eslint@9.24.0` |
| with **no `node_modules` at all** | 1049 · 1043 pass · 0 fail · **6 skip** — the same 6 as `main` |
| mutations | **19 applied one at a time, 19 caught** |

Both hash directions asserted for every field. `samples` is a **property over 12×12 values**: two manifests hash alike *exactly when* `sampleCountFor` would run the same number of samples. The round trip runs against the **real `lenses.json`**, never against its own output. The drift guard was fired by adding a field to the real manifest — it names the file and the field.

**One correction worth flagging**, since the audit that scoped this got it backwards: the panel's authoritative `samples` default is the exported `sampleCountFor`, which returns **2**. The `Math.max(1, Number(samples) || 1)` in `createWarmupGate` is a floor on an already-normalised value, not a default. The fix is to call `sampleCountFor` rather than mirror it — which also kills the fractional and `Infinity` divergences instead of the two known inputs.

**It has already earned its keep on a real change.** #671 reworded four lens rubrics and touched no field of `lenses.json`. The hash moved:

```
82c7519d7  sha256:efafe9def96d0c21d66e9625982c7458db8297ffc10c97fe577eeca434c92a70
bb21ff953  sha256:1c7853debf4edf92646d2299b0c924cb48cca89d6bb68b81648c57508a762f01
```

A reworded rubric is a different reviewer, and nothing else here would have said so.

**One follow-up, named rather than left to be found:** `scripts/agent/eval/README.md` is #677's file, and its "What exists today" table does not list these two modules while the line under it still says config identity is not built. Two lines, deliberately not taken here because the branches were written in parallel.

## Summary by CodeRabbit

- **New Features**
  - Added reproducible evaluation configuration snapshots and manifests.
  - Added configuration identity and content hashing based on effective evaluation settings.
  - Added validation for configuration fields, SDK versions, and reproducible timestamps.
  - Added support for restoring snapshots into panel-compatible lens files while preserving relevant settings.

- **Documentation**
  - Documented configuration identity, hashing behavior, validation, drift protection, failure semantics, and verification results.

- **Tests**
  - Added comprehensive coverage for snapshot round trips, field preservation, hashing, defaults, validation, and configuration drift.

### DIFF
--- a/docs/tasks/active/20260805-eval-config-identity-todo.md
+++ b/docs/tasks/active/20260805-eval-config-identity-todo.md
@@ -1,0 +1,425 @@
+# Config identity: what settings produced this review?
+
+Two modules under `scripts/agent/eval/`. `config-hash.mjs` reduces a config manifest
+to a fingerprint of the reviewer it describes; `config-build.mjs` turns the lenses
+directory into that manifest plus a self-contained snapshot, and turns a snapshot
+back into a lenses directory the panel can load.
+
+Neither invokes a model. Nothing here changes any lens's behaviour, and the panel
+keeps reading `scripts/agent/lenses/lenses.json` exactly as it does now.
+
+## The problem
+
+The panel is tuned by editing `lenses.json`: a model, a sample count, a reasoning
+effort, which file classes a lens reads. **Nothing in this repository can answer
+"what settings produced that review?"** A check run records a verdict; the manifest
+behind it is whatever was on `main` at the time, and after two more tuning commits
+that is no longer recoverable. The sharper form — *were these two reviews even
+produced by the same reviewer?* — has no answer at all.
+
+A fingerprint of the manifest answers both. But a fingerprint has two ways to be
+wrong and they are **not symmetric**:
+
+| | consequence |
+|---|---|
+| hashes the **same** when behaviour **differs** | merges two reviewers into one population. Nothing fails, nothing is logged, and the error surfaces as a comparison nobody can explain |
+| hashes **differently** when behaviour is **identical** | splits a population, which shows up immediately as fewer matching results than expected |
+
+The version this replaces was wrong in **both** directions at once.
+
+### Five divergences, and the one mistake underneath four of them
+
+Verified at `upstream/main` `82c7519d7`, re-derived at `bb21ff953`.
+
+| # | Where | What | Direction |
+|---|---|---|---|
+| 1 | `normalizeLens` | omits `scopeClasses` | **false negative** — a code-only lens and an everything lens hash identically |
+| 2 | `normalizeLens` | omits `effort` | **false negative** — two lenses differing only on the panel's main cost dial hash identically |
+| 3 | `normalizeLens` | hand-mirrors the panel's `samples` default | **false positive** — see below; not the divergence the audit predicted |
+| 4 | `buildConfig` | rebuilds each lens from a hardcoded seven-field list, dropping `scopeClasses` **and** `effort` on the way IN | silent loss |
+| 5 | `materializeLenses` | the same seven-field list, dropping both again on the way OUT | silent loss |
+
+1, 2, 4 and 5 are one mistake: **a field was added to `lenses.json` and nobody
+updated a hardcoded list.** `scopeClasses` arrived with #582's file-class routing;
+`effort` arrived later. Both lists were written before either.
+
+**Divergences 4 and 5 are the same bug at the two ends of one round trip, which is
+why neither was caught.** The old test asserted
+`buildConfig(materializeLenses(snapshot)) === the original hash` — and it passed,
+because a snapshot built and materialised by the same broken pair is perfectly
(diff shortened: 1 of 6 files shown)


harrykim8672:
@claude summarize

harrykim8672 (approved):

PR #682, opened 2026-08-06
Replay a frozen corpus item through the review panel

dlgpdmsly2:
The review panel is a **non-deterministic judge, and there is no way to ask it the same question twice.** A check run records a verdict; the diff behind it moves; two commits later the lens manifest that produced it is gone. So "did that tuning change help?" is unanswerable — the input never held still and neither did the reviewer.

`scripts/agent/eval/run.mjs` takes a frozen corpus item (#677), spawns `review-panel.mjs` over it, and stores an immutable envelope of what happened — per lens, per finding, with the exit code, the novelty gate's own self-report, and the panel's own wall-clock. Idempotent and resumable by `--run-id`: an item already stored is not re-run.

`review-panel.mjs` is unchanged. No lens's behaviour and no gate decision moves, and nothing in the panel imports anything from `eval/`.

## Four of the five defects fail in exactly the same direction: they turn a broken review into a clean one

1. **The lane was thrown away.** Every finding was rebuilt as `{lens,severity,file,summary,evidence}`. The lane rests on a `git blame` of the tree under review, and that tree is gone after merge — a replay that drops it destroys the only copy. Upstream fixed this exact bug once in `normalizeFindings` and published the postmortem. Adapters widen, never narrow: the whole finding is copied and `lens` added, overwriting — which is `stampLens`'s own documented rule, because a fill-the-blank version would let a finding declare which lens raised it.
2. **The exit code was discarded.** `runAgent` resolved `{outDir, code, stdout, stderr}` and the caller assigned none of it, and `runAgent` never rejects by design — so no non-zero exit could become a non-`ok` item. Fixed structurally rather than by remembering: `captureArtifacts` now takes the run result and refuses a directory.
3. **The gate's self-report went with it.** The panel prints one of three lines saying whether the novelty gate ran; it reached `stdout` and was dropped on the same statement. It is parsed into a named state and recorded in every envelope — without it a scorer cannot tell a run that had the gate from one that did not, and pooling those two is a measurement error.
4. **Missing output became "found nothing."** `readJson(panel.json) ?? []` turned a panel that crashed before writing into `findings: []`, `stageDetail: {}`, `status: "ok"` — indistinguishable from a genuinely clean pull request.
5. **Latency came from the wrong field.** `sumExecutions`'s flat sum overcounts by the concurrency factor (#669: a ~12-minute panel reported as 36–63). It now comes from `review-timing.json` via #669's `readWallMs`, or is `null` with `duration_source: "absent"` — never the sum as a *fallback*. On the end-to-end replay below those two read **742318** and **1189600**.

In a precision metric a false clean review is not noise, it is a **perfect score**. So `status` has two values, `ok` means "poolable as a real verdict", and three misconfiguration reasons (`gate-degraded`, `gate-unreported`, `no-lens-diff`) abort the whole run rather than spend on nineteen more items that would fail identically. The failing item is stored first, because the evidence is what makes it fixable.

**The gate assertion is deliberately asymmetric.** No `--base-sha` and the gate reports OFF is expected today — the replay tree comes from `git archive`, which has no `.git` — so it is recorded, not failed on. A `--base-sha` passed and the gate still OFF is a hard error: that is the silent-degradation case, and an inert gate is *safe* (every finding keeps gating), which is exactly why it must be loud. `runAgent` accepts a `baseSha` so the assertion is live and testable; the runner passes none. That is the seam, not the work.

## Also

- **`STAGE_DETAIL_DIFF_CONTENT` is on for replays, and the capture is asserted** rather than the flag trusted. #644 made it opt-in for good reasons, but its own docblock names a replay as the flag's single consumer, and the bytes are already in our frozen corpus item. Without it the fixture records the whole PR diff as every lens's input. The assertion counts the `lensDiff` *key*, never the bytes — `""` is a real value for a lens whose slice is empty.
- **`sdk_version` is read from `scripts/agent/package.json`** instead of the literal `"0.3.217"` it was written down as, via #680's `pinnedSdkVersion`.
- **`repo_context_files` is stable across replicates.** The marker was written inside the archived tree and counted before the write on first materialisation, after it on every cache hit — so one commit reported `N` then `N+1`.
- **Every envelope carries `panel_sha`**, and the store refuses one without it. The runner resolves the panel as its own *sibling*, so running from a feature branch measures that branch's panel; `(config_hash, panel_sha)` is what "same reviewer" means. A dirty tree outside `eval/` is refused too — HEAD's sha does not describe modified bytes.
- **The store gains the run slice** PR #677 deliberately left out. Run items are write-once for the opposite reason to corpus items: an item is a deterministic extraction whose re-write is a determinism check, a run item is an observation of a non-deterministic judge whose re-write would delete a data point. `config.snapshot.json` is identity and a *differing* one is refused rather than ignored, compared by `config_hash` rather than bytes (a snapshot carries `captured_at`, so identical configs differ in bytes by construction).

`adapters/reviewer.mjs` has never had a test file and it owns the subprocess contract. It has one now, and every test drives a **real subprocess for free** — `panelScript` is injected, so `adapters/stub-panel.mjs` writes canned output and exits with a code the test chose. The six-flag contract is asserted against what the stub *received* and cross-checked against the panel's own usage block.

## Verification

- **1116 agent tests · 0 fail · 0 skip**; 1057 on `main` (`3247650ae`). Delta +59.
- No new skips without `node_modules`: **6**, exactly the baseline (1 Agent SDK + 5 `lint-config`) — every subprocess test here runs with no dependencies installed.
- `npx eslint scripts` exit 0. `eval/test-lane.test.mjs` green with the new suites, including `eval/adapters/` — the depth the old flat glob missed.
- **28/28 mutations caught.** Four survived a first pass and all four were weaknesses in the tests, not the code; the `sdk_version` one was tautological in exactly the way `pinnedSdkVersion`'s docblock warns about, and its `readFile` is injected now.
- **Idempotence shown, not claimed:** two invocations at one `--run-id` over a real extracted corpus item — the second re-ran nothing and `shasum` over both stored files is byte-identical.
- **A full free end-to-end:** `extract-corpus.mjs --prs 664` → a 3176-file materialised tree → stub panel → a stored envelope carrying `lane: "backlog"` and `novelty: {origin: "relocated", …}` intact.
- Verified from the committed tree, not a working copy.

**No replay against the real panel has been run, by design** — this change must not spend money, so every test uses the stub. The first real one belongs with the worktree. Nothing here is that work: no real git worktree, no `--base-sha` passed, `--require-repo-context` still default off, no cost caps. The seams for all four are named explicitly in the task doc.

## Summary by CodeRabbit

* **New Features**
  * Added an offline evaluation runner for replaying frozen items through the review panel.
  * Added resumable runs with immutable results, configuration snapshots, failure classification, and aggregated metrics.
  * Added repository snapshot validation, gate and diff checks, timing data, and detailed run metadata.
  * Added persistent run and item storage with atomic writes and filtering.
* **Documentation**
  * Documented replay commands, storage formats, error handling, limitations, and verification coverage.
* **Tests**
  * Added comprehensive coverage for replay execution, persistence, resumption, validation, subprocess behavior, and failure handling.


dlgpdmsly2:
Note for the record — a gap in this PR that I'm fixing in a different PR, not here.

`runAgent` spawns the panel with no timeout, and `child.on("error", …)` resolves the
promise without killing the child. The tests have no `after()`/`afterEach` cleanup and
no `kill()` anywhere. So a panel that never exits leaves the promise unresolved *and*
the child alive.

This happened while iterating on this branch: a CI run hung for 31 minutes
(run `31073290840`). All 305 tests printed `ok`, no `# tests` summary line was ever
printed, and the job teardown reported six orphan processes — `node`, `sh`, `node`,
`sh`, `node`, `node`. Nothing failed; the runner just wouldn't exit. The change that
made it reachable is reverted, but the missing timeout and kill are still here.

Fixing it in a later PR rather than this one: on a real replay a panel that never exits is a
runaway bill, not a stuck CI job, so it belongs with the per-run cost caps PR 6 already
owns. Filed here so it's on record against the code it describes.

Separately, `ci.yml` sets no `timeout-minutes`, so that run was on course for GitHub's
360-minute default. Repo-wide and unrelated to this PR — its own change.

harrykim8672 (approved):

PR #681, opened 2026-08-06
Add Phase 1 observability to the agent PR pipeline

hwisoo-kim:
## Summary

Makes the autonomous review-and-fix loop legible from the GitHub UI alone (Phase 1 of the pipeline-observability plan):

- **`scripts/agent/loop-status.mjs`** — one sticky `<!-- agent-loop-status -->` PR comment, updated in place: a round-by-round table (head SHA, non-panel checks, per-lens verdicts, what happened next), the fix-round budget (`N of 3`, rerun-floor aware), the latest loop decision, and effort totals from the metric ledger. Re-derived **in full** from the unforgeable signals (commits, check runs, the paged latch) on every update, so a missed update self-heals; it is a projection, never a gate, and nothing reads it back. The upsert is author-gated (same trust rule as the paged latch) so a stranger on this public repo cannot plant a hijackable marker comment.
- **`scripts/agent/guard-verdict.mjs` + `review-round-guard.mjs`** — every guard path, including the previously **silent PROCEED**, now emits a one-line `verdict` step output (fed to the status comment) and a `$GITHUB_STEP_SUMMARY` block: rounds used vs cap, stall-detector reason, standstill count, infra, rerun hold. The CI arm's attempts guard gets matching SKIPPED/PAGED/PROCEED summaries plus an `attempts` output.
- **`scripts/agent/panel-job-summary.mjs` / `session-job-summary.mjs`** — the panel run page gets a per-lens verdict/verifier/cost table (fail-closed message when `panel.json` is missing); the implement and fixer runs get turns/tokens/cost/duration/outcome tables (seconds under a minute, so an agent that died in 18s doesn't read as "1m"); the CI-fix diagnosis is teed into the job summary instead of living only inside the fixer's prompt.
- Wired into `agent-review-panel.yml` (panel / promote / fix / stalled), `agent-iterate-ci.yml`, and `agent-implement.yml`.

**Display-only: no loop decision logic is altered.** Every new step is `continue-on-error`, every new script exits 0 on error, and jobs whose workspace is the PR branch run the trusted `main` copy (or skip quietly where a branch predates the scripts).

## Why

[...]

PR #688, opened 2026-08-06
Surface what the panel decided: enriched check bodies, a session ledger, run links on pages

hwisoo-kim:
Phase 2 of the pipeline-observability plan (follow-up to #681). Rendering only — no loop decision changes. Answers the plan's remaining questions: what each review found (Q2), what remains/was skipped/how disputes were adjudicated (Q3), where a failure happened (Q4), and per-iteration cost (Q5).

## 1. Enriched lens check-run bodies (`scripts/agent/severity.mjs`)

Each blocking row in a lens check body now renders:

- **A `file:line` locator** — the finding's own `line`, else the first *same-file* citation in its evidence (`novelty.mjs::findingLocation` is the one rule for both; a foreign-file citation never lends its line).
- **The per-finding verifier outcome** — `confirmed, high/low confidence`, the existing could-not-settle wording (unchanged), or `UNVERIFIED — the verifier session errored`. Stamped by `annotateFindings` as a reporting-only `verification` field from the same null-verdict signal `verifierTally` counts, so the per-finding marker and the aggregate outage banner cannot disagree. Stamped only when a verdicts array was actually supplied — a caller passing none must not fabricate an outage.
- **The adjudicator's decision *and prose reason*** on a finding that survived a dispute or fix-claim. The reason is computed every round and was discarded from every human surface — only the `upheld` integer is carried in `output.text`, by design (the 60k budget); the prose now lands in `output.summary`, the surface built for humans.
- **An "Author-reported skips" section** with the author's note — the fix-report section's documented not-yet-built item. Skipped findings still gate and stay listed in their severity section; this section makes the skip readable as a skip.

Trust and contract notes, applying #681's lessons up front:

- Lens bodies are copied verbatim into a **bot-authored** PR comment (`.github/workflows/agent-review-on-demand.yml`), and the paged latches trust the bot identity — so the two author-adjacent strings on this surface (adjudication reason, skip note) are `<!--`-neutralized with the `scripts/agent/fix-report.mjs` ZWNJ technique. Pinned by test.
- The fixer-prompt cut contract holds: every new section arrives via the same `\n### ` marker, and a skipped (blocking) finding cannot leak above the cut. Pinned by test.
- `scripts/agent/harvest.mjs`'s corpus reader round-trips the enriched rows — locator `:line` stripped into `file`, kept in `evidence`; sub-bullets and the skips section parse as zero additional findings. Pinned by a cross-module test against the **real renderer**, not a hand-typed copy of its output.
- A finding carrying none of the new fields renders byte-identically to today, so pre-existing bodies and old rounds are unchanged.

## 2. Per-session ledger (`scripts/agent/metrics.mjs::renderLedger`)

The effort summary's aggregates answer "what did everything cost" but not "what did round 3 cost". A `<details>` fold now lists every recorded session chronologically — kind, turns, weighted tokens, cost, duration (seconds under a minute, same rule as the fix-effort card) — with `review`/`review-fix` rows carrying their round ordinal (the nth panel record IS round n, the same ledger order `detectFlips` reads). Zero new data collection; rendering happens in `cmdSummarize` before any sweep, so `--final` summaries keep the full table.

Two rules guard it:

- A missing value renders `—`, never `0` — `Number(null) === 0`, and a legacy record with no `turns` did an unmeasured amount of work, not zero.
- `kind` is **allow-listed** to the pipeline's own kinds (anything else renders `other`): metric records are parsed from ANY comment on a public repo, so `kind` is the one free-text field an outsider could steer into this bot-authored body — the #681 ledger-injection shape.

## 3. "Where to look" on every 🛑 page

Every page comment now ends with a link to the failed run, the job/step that decided or died, and the transcript artifact where one exists:

- The round guard's five page paths, via new pure helpers `guard-verdict.mjs::whereToLookLine` / `runUrlFromEnv`. The URL is built only from the runner's own `GITHUB_*` env and is null on any missing piece — never a partial `.../runs/undefined` link — so pages posted outside Actions render exactly as today.
- The CI arm's attempts-cap page and no-advance page (job `iterate`, `claude-ci-fix-execution-output`).
- The panel fix job's no-advance page (job `fix`, step "Address panel findings", `claude-fix-execution-output`) and the `stalled` safety net (which names the failed job in its own reason; the link points at the run that contains it).
- `agent-fix.yml`'s pages already carried run links and are untouched. The `github-script` sites carry inline copies of the line for the usual cannot-import reason; they are display-only strings, so no byte-identity pin.

## Deliberately not done here

- **A pre-existing behavior bug found while tracing the adjudication data flow:** the incremented `adjudication.upheld` lives only on the copies in the gating array (`adjudicateRebuttals`/`applySkipClaims` return new objects), while `writeVerdict` persists `mergedAfter` — the pre-adjudication originals. So the count in `verdict.json`/`output.text` can never exceed 1 and the two-upheld **standstill page can never fire**. Kept out of this PR to stay rendering-only; a separate fix is in flight.
- Phase 3 of the plan: visible rebuttal bodies, warnings on silent `continue-on-error` steps, a comment when an implement run dies without opening a PR.

## Validation

- `cd scripts/agent && node --test *.test.mjs`: 1062 pass, 0 fail (23 new tests across severity/review-panel/metrics/guard-verdict/harvest).
- `eslint scripts` clean; every touched workflow parses; `verify:entropy` fully green (doc refs path-qualified).
- `docs/design/harness-engineering.md` updated (second observability slice under Phase 21) and the task doc extended; `tasks-index` regenerated.

## Summary by CodeRabbit

- **New Features**
  - Added richer review findings with file locations, verifier and adjudication outcomes, and sanitized author notes.
  - Added a dedicated section for author-reported skipped findings while preserving their blocking status.
  - Added chronological per-session effort ledgers showing rounds, tokens, costs, and durations.
  - Added “Where to look” links to failure and paging messages, including workflow, job, step, and available transcript details.

- **Documentation**
  - Updated observability documentation and added the `eval config identity` task to the active task list.

### DIFF
--- a/.github/workflows/agent-iterate-ci.yml
+++ b/.github/workflows/agent-iterate-ci.yml
@@ -430,7 +430,7 @@ jobs:
               '<!-- agent-paged -->' \
               '🛑 The CI-fix round did not advance the branch head, so CI stays red and the loop cannot re-trigger itself (no new push → no new CI). This can mean the fixer produced no new commit, did not converge within its turn budget, hit tool-permission denials, could not push, the branch head could not be determined, or an earlier step errored — see the run log and the `claude-ci-fix-execution-output` artifact. **A human should take over** on this PR.' \
               '' \
-              "Where to look: [this run]($GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID) → job \`iterate\`; transcript in the \`claude-ci-fix-execution-output\` artifact." \
+              "Where to look: [this run]($GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID) → job \`iterate\`; a fixer transcript, when one was captured, is in the \`claude-ci-fix-execution-output\` artifact (its upload warns instead of failing when the fixer died before writing one)." \
               | gh pr comment "$PR" --body-file -
             node ./scripts/agent/set-state.mjs "$PR" blocked || true
           fi
--- a/.github/workflows/agent-iterate-ci.yml
+++ b/.github/workflows/agent-iterate-ci.yml
@@ -201,7 +201,10 @@ jobs:
               core.info(`CI failed ${failedCi} times (limit ${max}); paging a human.`);
               await github.rest.issues.createComment({
                 owner, repo, issue_number: pr.number,
-                body: `${PAGED}\n🛑 CI failed ${failedCi} times on this branch (limit ${max}) without going green. **A human needs to take a look** — the harness could not self-resolve it.`,
+                // OBSERVABILITY (Phase 2): every 🛑 page carries a "Where to
+                // look" line. All three pieces come from the runner's own
+                // context, never from PR content.
+                body: `${PAGED}\n🛑 CI failed ${failedCi} times on this branch (limit ${max}) without going green. **A human needs to take a look** — the harness could not self-resolve it.\n\nWhere to look: [this run](${context.serverUrl}/${owner}/${repo}/actions/runs/${context.runId}) → job \`iterate\`, step "Attempts guard"; the failing CI runs are on this branch's checks tab.`,
               });
               // The agent:blocked label is set by the "Set state → blocked" step
               // below (single-value state via set-state), gated on this `paged`.
@@ -426,6 +429,8 @@ jobs:
             printf '%s\n' \
               '<!-- agent-paged -->' \
               '🛑 The CI-fix round did not advance the branch head, so CI stays red and the loop cannot re-trigger itself (no new push → no new CI). This can mean the fixer produced no new commit, did not converge within its turn budget, hit tool-permission denials, could not push, the branch head could not be determined, or an earlier step errored — see the run log and the `claude-ci-fix-execution-output` artifact. **A human should take over** on this PR.' \
+              '' \
+              "Where to look: [this run]($GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID) → job \`iterate\`; transcript in the \`claude-ci-fix-execution-output\` artifact." \
               | gh pr comment "$PR" --body-file -
             node ./scripts/agent/set-state.mjs "$PR" blocked || true
           fi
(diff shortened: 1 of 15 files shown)


harrykim8672:
@claude review

hwisoo-kim:
Verified all four CodeRabbit findings against the code; all four were valid and are fixed in e55357793.

1. **Absent-artifact references in the page bodies** — confirmed: both uploads use `if-no-files-found: warn` and the no-advance pages fire under `always()`, so on a hard fixer error the transcript genuinely may not exist. Fixed with wording, not workflow logic: the "Where to look" lines now say "a fixer transcript, when one was captured, is in the … artifact" — the same hedge the `stalled` page already uses. Adding shell conditionals to trusted YAML for a display string seemed like the wrong trade.

2/3. **Trailing space inside the `` `\n### ` `` code spans** (harness doc + task doc) — rewritten as `` `\n###` `` with the trailing space described in prose, keeping the semantic point that the space is part of the exact delimiter the fixer cut splits on.

4. **Unbounded `renderLedger` rows** — capped at `MAX_LEDGER_ROWS` (30). Real risk beyond tidiness: metric records are parsed from ANY comment, so an open-ended table could push the summary past GitHub's comment-size cap, failing the post and silencing the whole summary. Rows are built over the full chronological list first (so the `#` column and `review (round n)` ordinals are identical whether or not truncation happened), the NEWEST rows survive (they're what a reader is diagnosing; the totals above still cover everything), and the omission is stated inside the fold rather than silent. Regression test added, including the exactly-at-cap case.

Validation: 1063 tests pass / 0 fail, `eslint scripts` clean, all touched workflows parse, `verify:entropy` fully green.

hwisoo-kim:
Verified — the panel is right, and the shape it names is exactly the persisted one: the output.text carry strips adjudication to `{ upheld: N }` (no `verdict`, no `reason`) by design, so every carried-forward finding with an adjudication would have hit the fallback and been labeled "the overturn lacked grounded evidence" — a decision that never happened this round.

Fixed in the latest push: the label chain is now enumerated-verdicts-only. `"overturned"` on a *kept* finding is the one string `adjudicateRebuttals` stores that maps to the ungrounded-overturn label; absent, empty, or unknown verdict strings render **no** decision line at all — the same silent direction those findings rendered before the sub-bullet existed. Regression test covers `{ upheld: 2 }`, `verdict: ""`, and an unknown future value.

(Today this path is mostly latent — the upheld increment never actually reaches verdict.json, the pre-existing standstill-persistence bug called out in the PR description — but the fix for that is in flight, and this rendering must be correct the moment history starts persisting.)

Validation: full suite passes (1064 tests, 0 fail), eslint clean.

harrykim8672 (approved):

PR #689, opened 2026-08-06
Persist the adjudicated copies so the standstill bound can count

hwisoo-kim:
## Summary

The rebuttal-standstill bound — "a finding disputed `MAX_REBUTTAL_ROUNDS` (2) times and upheld every time pages a human" — was structurally inert: the `adjudication.upheld` counter died in the same panel round that computed it, so it could never persist past 1 and the standstill page in `scripts/agent/review-round-guard.mjs` could never fire. A finding could be re-disputed forever, bounded only by the round cap (which pages with a far less specific message).

**This is a behaviour change, not a refactor: the standstill page becomes reachable for the first time.** A finding disputed and upheld in two rounds now pages a human instead of buying a third adjudication session.

The counter was being dropped in three places, fixed together because they are the same bound:

1. **Persistence (the main defect).** `adjudicateRebuttals` and `applySkipClaims` in `scripts/agent/review-panel.mjs` RE-CREATE an upheld finding (`{...f, adjudication}`), and those copies lived only in `gating` — while `writeVerdict` persisted `merged`'s pre-adjudication originals into verdict.json. verdict.json is what the panel workflow builds the check run's `output.text` from, and `output.text` is the unforgeable channel both the next round's carry-forward and the round guard's `exhaustedFindings` count from. So every round re-derived `upheldCount(prior) + 1` from a prior count of 0. The new `substituteAdjudicated` replaces each original with its adjudicated copy in what `writeVerdict` persists, paired back by object identity (`applySkipClaims` is a positional map; `adjudicateRebuttals` now returns its re-creations as `[original, copy]` pairs). Mapping copies back to originals also fixes a latent half of the same bug: a finding that was skip-claimed and *then* overturned appeared in `dropped` as the skip copy, which the old identity filter over `merged` could never remove — overturned for the check run, still rendered as blocking for the human.
2. **The cluster fold.** `mergeCluster`'s `fold` dropped `adjudication` from folded wordings — while the workflow comment above the `output.text` builder already claimed the opposite ("carried on folded wordings too … see upheldCount"). A rebutted finding means the code did not change, so the next fresh pass almost always re-finds it, and the fresh representative wins the cluster slot — so the carried count had to survive in `mergedFrom`, where `upheldCount` takes the cluster max. It now does, integer only, exactly as the workflow trims it.
3. **The dedup collision.** `dedupeFindings` handed a byte-identical collision to the fresh copy and discarded the carried count with it. The winner now absorbs the loser's higher `upheld` count (`absorbUpheld`) — the counter is carried by the collision, never decided by it.

No workflow changes: `.github/workflows/agent-review-panel.yml` already reads `adjudication.upheld` (own and `mergedFrom`) out of verdict.json into `output.text`, and `scripts/agent/review-round-guard.mjs` already pages on it. The field simply never arrived until now.

## Why

The bound exists to route "the loop cannot settle this" to a human (Phase 28 of `docs/design/harness-engineering.md`). Without persistence the bound is a promise the code cannot keep: each dispute costs an adjudicator session, upholds, and is forgotten. The design doc gets a new subsection documenting the failure and the fix, following the existing "the loop shipped inert" precedent.

## Linked Issues

None — found by tracing the adjudication data flow end to end (`adjudicateRebuttals` → `writeVerdict` → verdict.json → `output.text` → `review-round-guard.mjs`).

## Author checklist

## Verification

Locally: `cd scripts/agent && node --test *.test.mjs` (1042 pass / 0 fail), `pnpm lint:scripts`, `pnpm verify:entropy`, and `pnpm verify:fast` (via the pre-commit hook) all green.

Skip reason (if applicable): n/a

## Risk Assessment

- User-facing risk: none — agent-pipeline scripts and one design doc only; no product code touched.
- Data/security risk: none new. Paging is the fail-safe direction: the change can only *add* human looks, never remove one. The count still travels only on the unforgeable check-run channel, never on author comments. The one behaviour to be aware of: PRs where the fixer disputes or skips the same finding in two consecutive rounds will now page instead of looping silently — that is the designed destination, but expect the standstill page to start appearing.
- Rollback plan: revert the commit; the counter falls back to resetting every round (pre-fix behaviour), nothing else depends on the new field reaching verdict.json.

## Notes for Reviewers

- New tests in `scripts/agent/review-panel.test.mjs`: the persistence regression (the array `writeVerdict` persists carries `adjudication.upheld` after an upheld dispute), the skip-claim pairing and overturn-after-skip identity case, an end-to-end two-round test driving the real wiring (merge → gate → adjudicate → substitute → carry forward) showing `upheldCount` reach 2 through **both** merge paths (cluster fold and dedup collision), and a source-level test that the round loop routes `writeVerdict`'s input through `substituteAdjudicated` at all (same pattern as the existing `stampLens` wiring test, since `main()`'s per-lens loop is not unit-testable).
- AI assistance: this PR was written with Claude Code (analysis, implementation, and tests); I reviewed and signed off on the full diff.
- UI changes: none.
- Follow-up work: none required; the "adjudication record in the metrics comment" gap noted in Phase 28 remains open and is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rebuttal progress now persists correctly across review rounds and can reach standstill limits.
  * Upheld finding counts are preserved when findings are merged, deduplicated, or restated.
  * Skip-claimed and overturned findings are handled correctly in persisted verdicts.
  * Adjudicated findings replace outdated copies, ensuring accurate final results and reporting.

* **Tests**
  * Added coverage for multi-round rebuttals, standstill limits, finding identity, clustering, deduplication, and persisted adjudication outcomes.

### DIFF
--- a/docs/design/harness-engineering.md
+++ b/docs/design/harness-engineering.md
@@ -1695,6 +1695,56 @@ the gating step would leave every overturned finding un-removable from the summa
 round loop actually routes `merged` through it, since that loop lives in `main()`
 and no unit test can reach it.
 
+#### The bound was still inert: the count died in the round that computed it
+
+Stamping the lens made disputes *match*; it did not make the counter *persist*.
+An upheld finding is RE-CREATED (`{...f, adjudication}`) by `adjudicateRebuttals`
+and `applySkipClaims` in `scripts/agent/review-panel.mjs`, and those copies lived
+only in `gating` — what the check's conclusion is computed from — while
+`writeVerdict` persisted `merged`'s pre-adjudication originals into verdict.json.
+verdict.json is where the panel workflow reads `adjudication.upheld` into the
+check run's `output.text`, and `output.text` is the unforgeable channel both the
+next round's carry-forward and the round guard's `exhaustedFindings` count from.
+So the increment died in the same round that computed it: every round re-derived
+`upheldCount(prior) + 1` from a prior count of 0, the counter could never reach
+`MAX_REBUTTAL_ROUNDS`, the standstill page could never fire, and a finding could
+be re-disputed forever — bounded only by the round cap, which pages with a far
+less specific message than the one this bound exists to send.
+
+`substituteAdjudicated` closes it: the adjudicated copies replace their originals
+in what `writeVerdict` persists, paired back by object identity
+(`applySkipClaims` is a positional map; `adjudicateRebuttals` returns its
+re-creations as `replaced` pairs). Mapping copies back to originals also fixed a
+latent half of the same bug — a finding that was skip-claimed and *then*
+overturned appeared in `dropped` as the skip copy, which the old identity filter
+over `merged` could never remove.
+
+Persisting the count exposed two more places the same integer silently died on
+the way into round N+1, both in the merge:
+
+- **`mergeCluster` dropped `adjudication` from folded wordings**, while the
+  workflow comment above the `output.text` builder already claimed the opposite
+  ("carried on folded wordings too … see upheldCount"). A rebutted finding means
+  the code did not change, so the next fresh pass almost always re-finds it — and
+  the fresh representative wins the cluster slot, so the carried count had to
+  survive in `mergedFrom`, where `upheldCount` takes the cluster max. It now does,
+  integer only, exactly as the workflow trims it.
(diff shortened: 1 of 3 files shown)


hackerwins:
@coderabbitai review this.

hackerwins (approved):
Thanks for your contribution.

PR #691, opened 2026-08-06
Stop the determinism gate scoring a replay that never ran

hwisoo-kim:
## Summary

The first live run of the UI hunter (PR 4b). It found **four defects — three in the hunter, one in the eval runner** — none of which was reachable by reasoning, all of which were obvious the moment code ran against a real browser and a real hook.

## The hunter would have "reproduced" every candidate it ever found

Replay scored **3 of 3 candidates as `reproduced` and `deterministic` on plans where nothing executed.**

`resolveActionRefs` builds the replay plan from the *cited* refs — and a model cites the actions that **demonstrate** a finding, not the ones that set up the state. All three candidates cited ranges starting at 9 or 32; none included the opening `goto`. Replaying just those meant a browser that had never navigated, so every read failed with `hunt bridge is not installed`. All three attempts then failed **identically**, their `uiPlanKey`s matched, and the stage whose whole purpose is killing phantom repros returned a perfect score.

- Replay now runs the **journal prefix** up to the failing action — which is what the design always said stage 3 does.
- Plus a **fail-closed guard**: a run in which every action errored is not a reproduction, however consistently the attempts agree.

## The turn ceiling sat below the action budget

`explorerMaxTurns: 60` against `maxActions: 80`. Every action costs at least one turn, so the budget was unreachable *by construction* — both briefs died at `error_max_turns` having proposed nothing. `validatePersona` now refuses it; the CLI charters sit at 1.25–1.4×. Also `verifierMaxTurns: 20 → 35`, which every verifier hit on the second run.

## A failed brief discarded its journal

`explore-raw.json` was written as `[]` — the one file that exists to explain a zero run did not survive the *most likely* failure. The partial journal now rides out on the error.

## And `-C` does not scope git

Found while trying to push. `GIT_DIR` in the environment overrides both `-C` and `cwd`, and anything inside a git hook has it set — `pre-push` runs `verify:self`, so the whole agent suite inherits it.

- **`materializeRepoAt` (production)** passed `-C repoSource` to `git archive` with the ambient environment. Under an inherited `GIT_DIR` it archives the *ambient* repository at that commit instead of `repoSource`, silently — git succeeds, the tar extracts, and the panel reviews a tree from the wrong repo. Same shape as the `#521` note a few lines above it.
- **`eval/run.test.mjs`** built its fixture repo with `cwd: src` alone, so under a hook it ran against the real repository: it added `a.mjs`/`b.mjs` to the developer's index and committed them as `t <t@t> "two files"`, then its own `rm -rf` left every tracked file reading as deleted. That commit was real and had to be dropped by hand. It is the only fixture-git test in the suite not already using `fixtureGitEnv` — `novelty`, `git-env` and `collect-captures` all scrub.

Net effect before this: `pnpm verify:self` fails from inside any hook, so **`git push` is blocked for everyone**, and the suite writes commits into the repository under test.

## What the run cost, and what it proved

Two runs, **$9.38**. After the budget fix the explorer found the seeded defect independently in *both* briefs, with exactly the predicted shape:

> `[critical] Typing into the document silently drops every second character`
> `expect: doc.text contains "@input:13" ground=A` → `violated`, `eligible`

So the explorer, the prediction protocol, the grounding tiers and the journal all work end to end against a real browser. Replay did not, and now does.

## Test plan

- `cd scripts/agent && node --test '**/*.test.mjs'` with `node_modules` hidden — **1297 pass / 0 fail / 1 skipped**
- The same suite **with `GIT_DIR`/`GIT_INDEX_FILE` set**, reproducing the hook environment — 1298/1298. Before this PR that combination failed.
- `pnpm lint:scripts`, `pnpm verify:entropy`, `hunt-ui.mjs preflight` — clean
- **Mutation-tested**, each guard reverted and the named test confirmed to fail: the prefix plan, the all-failed-run guard, that guard biting in the pipeline, the turn-ceiling check, the journal surviving the throw, and the failed brief still being persisted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository handling to ensure evaluations and temporary test runs target the correct project.
  * Strengthened UI issue reproduction by requiring successful actions and validating replay attempts.
  * Preserved partial exploration details when investigations fail, making skipped results more traceable.

* **Quality Improvements**
  * Expanded validation and automated coverage for UI exploration, replay behavior, action limits, and failed investigations.
  * Tuned exploration and verification limits to support more thorough issue discovery while reducing unnecessary actions.

### DIFF
--- a/scripts/agent/charters-ui/personas.json
+++ b/scripts/agent/charters-ui/personas.json
@@ -32,12 +32,12 @@
     "maxCandidates": 8,
     "reportableSeverities": ["critical", "major"],
     "actionBudget": {
-      "maxActions": 80,
+      "maxActions": 50,
       "perActionTimeoutMs": 30000,
       "totalTimeoutMs": 900000
     },
-    "explorerMaxTurns": 60,
-    "verifierMaxTurns": 20
+    "explorerMaxTurns": 70,
+    "verifierMaxTurns": 35
   },
   {
     "id": "sheet-author",
@@ -71,11 +71,11 @@
     "maxCandidates": 8,
     "reportableSeverities": ["critical", "major"],
     "actionBudget": {
-      "maxActions": 80,
+      "maxActions": 50,
       "perActionTimeoutMs": 30000,
       "totalTimeoutMs": 900000
     },
-    "explorerMaxTurns": 60,
-    "verifierMaxTurns": 20
+    "explorerMaxTurns": 70,
+    "verifierMaxTurns": 35
   }
 ]
--- a/scripts/agent/eval/run.mjs
+++ b/scripts/agent/eval/run.mjs
@@ -36,6 +36,7 @@
 
 import { mkdtempSync, mkdirSync, existsSync, writeFileSync, readFileSync, rmSync, readdirSync } from "node:fs";
 import { execFileSync } from "node:child_process";
+import { repoScopedEnv } from "../git-env.mjs";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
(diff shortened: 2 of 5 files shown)


harrykim8672:
@coderabbitai review this PR

harrykim8672 (approved):

PR #719, opened 2026-08-07
Strip blockquote markers before parsing a CodeRabbit review body

dlgpdmsly2:
## Summary

Follow-up to #714. That PR reported **100.0% recovery** of CodeRabbit's review-body
findings against CodeRabbit's own declared counts. The number was right; the `detail`
field behind it was wrong on **146 of 1626 findings (9.0%)**.

Blockquote markers are now stripped once, at the two entry points, and
`CR_PROSE_END`/`CR_WRAPPER` go back to being anchored on plain text.

## Why

CodeRabbit wraps a tier section in a GitHub alert block (`> [!CAUTION]`) whenever its
findings fall outside the diff, and that prefixes **every line** of the section with
`> `. Neither new nor rare: **96 of this repo's 558 CodeRabbit review bodies**, back to
2026-04. Caught on #716's review body, posted twelve minutes after #714 merged.

The counts survived it — `CR_LOCATOR` already tolerated a leading `>`, so locators,
headers, tiers, files and the recovery rate were all correct. `CR_PROSE_END` did not:
with no `>` in its leading class it never matched on a quoted body, so the prose boundary
was never found and `detail` ran to the end of the finding, swallowing the
`🤖 Prompt for AI Agents` block.

That block is the specific thing `codeRabbitDetail`'s own docblock spends a paragraph
explaining must never reach the comparison: its boilerplate contributes `code`, `fix`,
`issues`, `changes`, `validate` to every comparison and inflates containment on the
vocabulary every finding already shares — eating real misses exactly where the panel had
the most findings.

**The lesson, one level in from #714's:** that PR verified the *count* and never the field
the count produced. A denominator establishes whether you found everything. It says
nothing about whether what you found is right.

**The first fix was wrong, and it is the tempting one:** add `>` tolerance to each of the
four line-anchored patterns. It fails, because an **empty** quoted line is `>` alone — not
blank, and matching no wrapper — so the header-below-the-locator search stops on it and
reads no header at all. The new test for that case failed against that fix, which is the
only reason it was caught before pushing. One normalisation beats four permissive
character classes, which are four chances to miss one.

Stripping is line-anchored, so a `>` **operator** in prose or backticked code is
untouched. All six surviving cases were checked by hand and every one is a real comparison
(`marL + marR > cell width`, `valueFields.length > 1`, `next > 0`), not a quote marker.
The strip only removes markup, so it cannot reduce the finding count.

## Linked Issues

None — follow-up to #714.

## Author checklist

## Verification

Skip reason (if applicable): *left unticked until CI posts its comment — these get ticked
from CI's own result, not from a local run.*

Measured locally, from the committed tree, with no `node_modules` present:

| | |
|---|---|
| `cd scripts/agent && node --test '**/*.test.mjs'` | **1386 tests · 0 fail · 6 skip** |
| baseline on `upstream/main` (`1f9ee8864`, now contains #714) | **1382 · 0 fail · 6 skip** |
| delta | **+4 tests, no new skips** |
| `npx eslint scripts` @ pinned `9.24.0` | **exit 0**, clean |

- **Boilerplate in `detail`: 0 of 1626** review-body findings and **0 of 1485** inline
  findings, down from 146. Counts unchanged — still 1626 of 1626 declared (100.0%), 0
  unrecognised sections, 3111 total.
- **#716's real body**, the one that exposed this: 1 of 1 declared, tier
  `outside-diff-range`, file `scripts/agent/eval/run.mjs` from the enclosing `<summary>`,
  locator `421-426`, lens `security`, `detail` 804 chars of title + prose and nothing
  else.
- **3 mutations**, each red on the test that names it, then restored green.

## Risk Assessment

- **User-facing risk:** none in the gating path. `harvest.mjs` is a hand-run offline CLI —
  no workflow invokes it and nothing imports it but its own test. This change only removes
  markup from a text field, so it cannot change any count: recovery stayed at 1626 of
  1626 and the inline total at 1485.
- **Data/security risk:** none. No author or login check touched, no new endpoint, no new
  scope. `misses.jsonl` and its write path are unchanged.
- **Rollback plan:** revert. Nothing consumes `detail` from the review-body path yet — it
  is parsed and filed nowhere — so a revert restores #714's behaviour exactly.

## Notes for Reviewers

- **AI assistance:** written with Claude Code, as #714 was. The 96-body and 146-finding
  figures were measured against the live API over all 638 PRs, and the three mutations
  were run to confirm each new test fails on the code it protects.
- **One constructed fixture, labelled as such in the test** — a quoted section whose
  header sits on the next line. Both halves are attested independently (96 quoted bodies;
  header-below-locator in #11 and #477), but no body currently does both at once, so it is
  a union rather than an observation. If it is ever cut, cut the test with it.
- **Unrelated, and not fixed here:** `eval/run.test.mjs:659` scans `tmpdir()` globally for
  `eval-item-*`, which is shared across runs and processes. It failed once during this
  work with 1355 stale directories present from repeated local lane runs, and passed again
  once they were removed. That is #713's test and looks like a real latent flake under the
  concurrent runner, but no code here is involved.
- UI changes: none.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of CodeRabbit reviews displayed within GitHub blockquotes.
  * Findings now consistently retain their severity, file, location, title, and clean details.
  * Removed unwanted AI-agent boilerplate and metadata from extracted review details.
  * Improved handling of nested quotes and findings with titles below their code locations.

* **Tests**
  * Added regression coverage for blockquoted review sections and detail cleanup.

* **Documentation**
  * Documented the issue, resolution, and verification results.

### DIFF
--- a/docs/tasks/active/20260807-coderabbit-parser-widening-todo.md
+++ b/docs/tasks/active/20260807-coderabbit-parser-widening-todo.md
@@ -262,3 +262,110 @@ match on `(pr, title)`; the tier is on the finding so it can.
 
 **Not verified:** whether the 1626 review-body findings are useful *as corpus
 candidates* — none is filed, and that judgement belongs with the record schema.
+
+---
+
+# Follow-up: the alert-quoted review body, and the field the denominator did not cover
+
+Appended 2026-08-07, after #714 merged. Caught on **#716's review body**, which
+CodeRabbit posted at `09:05:21Z` — twelve minutes after #714 merged at `08:53:44Z`.
+
+## The problem
+
+CodeRabbit wraps a tier section in a GitHub alert block (`> [!CAUTION]`) whenever its
+findings fall outside the diff, and that prefixes **every line** of the section with a
+`>` and a space. Not new and not rare: **96 of this repo's 558 CodeRabbit review
+bodies**, back to 2026-04.
+
+Every count #714 reported was right on those bodies. `CR_LOCATOR` already tolerated a
+leading `>`, so locators, headers, tiers, files and the 100.0% recovery were all correct.
+
+**`detail` was not.** `CR_PROSE_END` had no `>` in its leading class, so on a quoted body
+the prose boundary was never found and `detail` ran to the end of the finding —
+swallowing the `🤖 Prompt for AI Agents` block. That block's boilerplate is the exact
+thing `codeRabbitDetail`'s own docblock spends a paragraph explaining must never reach the
+comparison: it contributes `code`, `fix`, `issues`, `changes`, `validate` to every
+comparison and inflates containment on the vocabulary every finding already shares —
+eating real misses where the panel had the most findings.
+
+**146 of 1626 review-body findings — 9.0%**, median `detail` length 1590 chars.
+
+The lesson is the one this module keeps re-learning, one level in: **#714 verified the
+count and never the field the count produced.** The recovery rate was 100.0% and honest;
+the record behind it was wrong. A denominator checks whether you found everything, not
+whether what you found is right.
+
+## The change
+
+`CR_BLOCKQUOTE_PREFIX` — `/^[ \t]*(?:>(?:[ \t]|$))+/gm` — stripped **once** at the two
+entry points: `codeRabbitReviewSections` (so every line-anchored pattern downstream sees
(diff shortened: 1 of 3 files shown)


hackerwins:
@coderabbitai review this.

harrykim8672 (approved):

PR #731, opened 2026-08-09
Snap board move and resize onto the visible grid

hackerwins:
## Summary

#722 shipped the board's background grid as display only and named this as its own follow-up. A grid you cannot land on is a backdrop, not a tool.

**The step is the grid the user can see** — the same `gridStep(zoom)` ladder that paints it — so snapping never puts anything on a line that is not drawn. Snap is a second toggle, independent of the display mode (`Grid: None` + snap on is a real preference, and Miro models it the same way), reachable from the toolbar's `Grid ▾` dropdown and from the empty-canvas context menu. On by default, like the display.

### The seam

Two additive `SlidesEditorOptions`. A slides mount passes neither, so slides behavior is unchanged:

```ts
getSnapGrid?: () => number | null;             // world step, null = off
hostCanvasMenuItems?: () => ContextMenuItem[]; // empty-canvas menu extension
```

`getSnapGrid` is a callback because the step depends on live zoom and the editor is built once. `hostCanvasMenuItems` is the general form of the bespoke `onFitToContent` hook — board-only menu entries should not each earn an editor option.

### How it behaves

- **Fallback, not candidate.** `snapDelta` runs its existing slide-centre / guide / element-edge contest first and only quantizes an axis nothing won. Aligning to another object is the stronger intent; a grid that outranked edges would make two shapes impossible to butt together unless their sizes happened to be multiples of the step.
- **No threshold** — it rounds. The nearest line is never more than half a step away, so reusing the 8-unit band would leave the toggle inert at low zoom (16% of the time at zoom 0.25). It emits no guide either: the painted grid is its own feedback.
- **Resize** rounds only the edges the dragged handle moves, anchor edge fixed. Suppressed per axis by an equal-size match, wholesale by Shift (aspect), and by Alt.
- **Move and resize drags only.** Drag-to-insert, connector endpoints, rotation and the arrow-key nudge do not snap — see Known limitations in the task doc.

### Two defects found by review, both pre-existing invariants this broke

- **A click could move an element.** `startDrag`/`startResize` arm on pointerdown with no movement threshold, because a select-click's release must stay a no-op. Every other snap passes that through by returning zero for a zero delta; the grid quantizes. Gated on `DRAG_THRESHOLD_PX` (3 px), the same threshold the slow-double-click classifier uses.
- **Shift axis lock inverted.** The lock re-derived its axis from the *corrected* delta, sound only while corrections were bounded. A horizontal Shift-drag whose X was cancelled onto the lattice could hand Y half a step, flip the dominant axis, and move the element sideways. The axis is now resolved once from the raw delta and re-applied.

Review also surfaced that `matchSize` never had the zero-delta property either (it snaps to a peer within 8 units of the *current* size, so it corrects before the pointer moves, and `startResize` commits unconditionally). That is a live defect independent of the grid; the same gate now covers it.

## Test plan

- `pnpm verify:fast` — exit 0.
- New unit tests: `grid-snap.test.ts` (8), `snap.test.ts` grid block (6), `board-grid.test.ts` persistence (6).
- New editor-level tests driving the real editor: move/resize quantization, Alt suspension, the no-grid path (slides unchanged), the context-menu hook, multi-selection resize, per-axis equal-size suppression, the click/tremor gate on move and resize, Shift axis-lock regression (×2), Shift-suppression on resize, and the `matchSize` zero-delta case.
- One pure test in `drag.test.ts` pinning *why* the lock must read the raw delta.
- Manual browser smoke in `pnpm dev`: snap on drag, Alt suspend/resume, click-does-not-move, resize edge behavior, context-menu toggle state, `None` + snap on.

Design: `docs/design/board/board.md` § Snap to grid. Task: `docs/tasks/active/20260809-board-snap-to-grid-{todo,lessons}.md`.

## Summary by CodeRabbit

* **New Features**
  * Added an enabled-by-default, independently configurable Snap to grid preference.
  * Grid snapping now applies to move and resize actions using the visible grid spacing.
  * Added toolbar and canvas context-menu controls, with preference persistence.
  * Existing guides and object-edge snapping take priority over grid snapping.
  * Added host-provided canvas menu item support for compatible editors.

* **Bug Fixes**
  * Prevented accidental snap corrections on clicks and very small movements.
  * Improved axis locking, modifier-key overrides, resizing, and rotated-element behavior.

### DIFF
--- a/docs/design/board/board.md
+++ b/docs/design/board/board.md
@@ -201,10 +201,59 @@ This also made `commitViewport` in `board-view.tsx` the single chokepoint
 for pan/zoom: minimap-navigate, wheel, and space/middle-drag each used to
 repeat the same `setViewport` + `repaintViewport` pair inline.
 
-**Snap to grid is deliberately not part of this.** Miro keeps it a
-separate toggle too. It extends `snapDelta` (`slides/view/editor/snap.ts`)
-with grid candidates and carries a drag / resize / connector regression
-surface that display alone does not.
+### Snap to grid
+
+A second, independent toggle (Miro keeps them separate too): the mode
+controls what is *drawn*, snap controls where things *land*, and
+`Grid: None` + snap on is a real preference. Off by default, unlike the
+display — drawing lines changes nothing about a board, while snapping
+changes where every subsequent drag ends up, and a Miro-imported board's
+elements sit at arbitrary coordinates that an on-by-default snap would
+relocate the first time anyone nudged one.
+
+The step is the **visible** grid — the same `gridStep(zoom)` the
+background is painted from — so the user always lands on a line they can
+see. Move and resize snap; connector endpoints and rotation do not.
+
+Two additive `SlidesEditorOptions` carry it, and a slides mount passes
+neither, so slides behavior is bit-identical:
+
+```ts
+getSnapGrid?: () => number | null;          // world step, null = off
+hostCanvasMenuItems?: () => ContextMenuItem[];
+```
+
+`getSnapGrid` is a callback because the step depends on live zoom and the
+editor is built once. `hostCanvasMenuItems` is the general form of the
+bespoke `onFitToContent` hook — board-only menu entries should not each
+earn an editor option — and carries the `Snap to grid` item that
+right-clicking the empty canvas offers alongside the toolbar's checkbox.
+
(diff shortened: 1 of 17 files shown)


hackerwins:
@coderabbitai review this.

Issue #700, opened 2026-08-06
Design editor: ship as a local Vite plugin, split into two packages


hotsunchip:
## Summary

Ship the design editor as a **dev-only Vite plugin a developer installs in their own project** (`@wafflebase/design-editor`), rather than as a hosted editor on `wafflebase.com` that loads arbitrary repositories. Split the existing `packages/design-sdk` work along that boundary into a published generic package and a private wafflebase reference app.

Design doc: [`docs/design/design-editor/design-editor-local-plugin.md`](docs/design/design-editor/design-editor-local-plugin.md) — landed in #701. **Its rollout table is the plan of record**; the ledger below tracks it against real PR numbers.

## Why

The hosted model required running a Vite dev server per user with the customer's repository checked out and installed. Every load-bearing mechanism in the engine assumes the source tree sits next to a running dev server — `apply: "serve"` middleware that *writes* source, `?wbFrame=` module ids, HMR pushes, a second HTML entry. Serving arbitrary repos that way is a multi-tenant dev-container orchestrator with arbitrary code execution, cold starts and npm supply-chain exposure. That is a separate product.

Running the plugin on the developer's own machine removes the hosting requirement **and** the credential surface: no GitHub App, no PAT, no branch/PR creation. Edits land in their working tree and `git diff` is the review surface.

### Two packages

| Package | Published | Contents |
| --- | --- | --- |
| `@wafflebase/design-editor` (`packages/design-editor`) | intended; `private: true` today | Vite plugin, AST mutators, bridge protocol, editor shell UI |
| `packages/design-sandbox` | no (`private: true`) | Wafflebase's instance: scene manifest, providers, fixtures, canvas scenes, `@wafflebase/core` token adapter |

**One-way dependency:** `design-sandbox` → `design-editor`, never the reverse. Nothing under `packages/frontend` imports either.

**The mechanical test:** the published package declares **no `@wafflebase/*` runtime dependency**. A reversal fails `pnpm install` in a consumer project instead of needing to be caught in review.

### One public entry point

```ts
designEditor({ root, scenes, providers, tokens })
```

Replacing today's six internal plugins (`mutationBridge`, `scenesRegistry`, `scenePatch`, `yorkieOffline`, `safelistFile`, `antlr4tsAssertShim`).

### Support matrix

[...]

PR #738, opened 2026-08-10
Add the data-wb-node stamper with a cross-consumer drift guard

hotsunchip:
> **Stacked on #718.** GitHub shows the cumulative diff until that merges; **this PR's own delta is 408 lines across 3 files**, all under `packages/design-editor/*/server/`. It cannot branch from `main`: `stamp.mjs` imports `jsx-nodes.mjs`, which lands in #718.
>
> Rebased onto #718's current head, so it carries the `allowJs` and fragment-ownership work rather than reverting it.

## Summary

Fourth PR of the design-editor series ([#700](https://github.com/wafflebase/wafflebase/issues/700)): the dev-only `data-wb-node` stamping transform. `stamp.mjs` arrives **byte-identical to its prototype**; everything else is tests, plus one deletion argued for below.

`stamp.mjs` is the **third consumer** of the `jsx-nodes.mjs` numbering. The extractor emits paths with it, the injector resolves anchors with it, and this writes it into the DOM so a click can be mapped back to source. Its header states the risk: three implementations of *"which child is index 2"* would drift, and the drift surfaces as **a click selecting the wrong node**. Nothing checked that the numbering was still shared.

### The load-bearing test

One component exercising every numbering rule simultaneously — two returns, a fragment, a ternary, a conditional, an iteration, and a sibling helper root — stamped, then asserted equal to what `walkJsx` independently yields.

A second test **guards the guard**: if both sides collapsed to nothing, or to the first return only, that equality would still hold. So the ids are also checked to contain both roots, both returns, and the `<li>` inside the `.map`.

Deriving the expectation from `walkJsx` rather than hard-coding it is why this suite survived #718's numbering changes without a single edit.

### Two counter-intuitive properties, pinned

**Stamping preserves every path but none of the fingerprints.** `fp` covers attribute *names*, and stamping adds three attributes. So `data-wb-fp` carries the **baseline** fingerprint, captured before the rewrite — which is precisely why the host must read the attribute rather than recompute it from the served source. Anyone who "optimises" that into a recompute breaks click resolution with no error anywhere.

**Stamping is not idempotent.** A second pass treats the injected attributes as ordinary ones and duplicates them.

### Splice correctness, from both ends

Eight fixtures re-parse with zero diagnostics, and `<Select<string> …>` gets its own test: `attributes.pos` sits after type arguments where `tagName.end` does not, so the naive offset would emit `<Select data-wb-node="…"<string> …>`.

Also covered: `data-wb-file` is **omitted rather than emitted broken** when a path contains `"`, `<`, `>` or `&`, while node and fp are still written.

## This deletes `stamp.d.mts`

An adjacent declaration file **shadows the module it describes**. `tsc --listFiles` loaded `stamp.d.mts` and dropped `stamp.mjs` from the program entirely, so the `// @ts-check` on its first line never ran, and the declaration was free to drift from the code it documented.

This is the same class of dead pragma #718 just fixed for `jsx-nodes.mjs`, but hidden a level deeper: the file being skipped was not the file carrying the pragma, so turning `allowJs` on did not surface it.

Measured rather than argued — planting one type error in `stamp.mjs`:

| | errors |
|---|---|
| with `stamp.d.mts` present | **0** |
| with it removed | **3** (`TS2362`, `TS18048`, `TS2345`) |

Nothing is lost by removing it. `allowJs` is now on, so a TypeScript importer gets the same signature from the JSDoc — which is how `jsx-nodes.mjs` has always been consumed. `stamp.mjs` typechecks clean under its own pragma with no annotation changes.

### Correcting myself

An earlier revision of this PR body claimed the `.mts` test made the declaration "genuinely gated." That was wrong. It verified the declaration was *self-consistent and matched its call sites* — it could not verify the declaration matched `stamp.mjs`, because `stamp.mjs` was not in the program at all. The deletion is what closes that gap.

The `.mts` test stays, with its purpose narrowed: the pragma now checks the annotations against the **implementation**, and this checks them against a **consumer** — importing exactly as the sandbox Vite config's dynamic import does. The behavioural suite cannot stand in for it, being `.mjs` with no pragma and therefore never type-checked at its call sites.

## Test plan

- `pnpm --filter @wafflebase/design-editor test` — 4 files / **111 tests** pass.
- `pnpm --filter @wafflebase/design-editor typecheck` — exit 0; proven to fail on a planted error in `stamp.mjs` (it did not before this deletion).
- `pnpm verify:entropy` — 0 dead-code, 0 doc-staleness.
- `stamp.mjs` diffed against `feat/design-system`: **byte-identical**.

### Merge order

This repo squash-merges. Merging a later PR in this chain first would fold the earlier ones into it under the wrong title and leave them open with an empty diff. **Order: #718 → this.**

### Local verification caveat

`--no-verify` on commit and push. `pre-push` runs `pnpm verify:self`, which fails on this machine in `verify:fast` → `packages/frontend` on two 5000 ms timeouts in files this diff does not touch, both passing in isolation. Documented on #717, where CI's `verify-self` passed. CI is the authority.

Tracking: #700

## Summary by CodeRabbit

- **New Features**
  - Added development tooling that automatically annotates JSX elements with stable node identifiers, fingerprints, and source-file metadata.
  - Supports reliable element tracking across nested components, repeated structures, fragments, and iterative edits.

- **Documentation**
  - Updated design editor engine documentation with the current server-module layout and type-checking guidance.
  - Refined the local plugin rollout plan, including review sequencing and cumulative-diff behavior.


hotsunchip:
@claude review

dlgpdmsly2 (approved):

PR #740, opened 2026-08-10
Add a workflow_dispatch-only workflow that replays the eval corpus in CI

dlgpdmsly2:
## Summary

Adds `.github/workflows/eval-replay.yml`, which replays the frozen eval corpus through the review panel on a runner, plus `eval/replay-plan.mjs` — a preflight that validates one dispatch and resolves the git refs it needs — and its tests.

This is the first workflow here that spends model budget on demand. It can only be started by hand, its cost ceiling is a required input with no default, and a free dry-run mode drives a stub panel end to end.

## Why

`eval/run.mjs` can replay a frozen pull request, and every replay so far has run on a laptop. The panel is a non-deterministic judge, so a result somebody dislikes can always be explained by "the environment differed" — and on a laptop there is no way to check. A runner is the same box every time, and the envelope already records which panel commit and which lens configuration produced each item; the missing half was a place to run it that someone else can reproduce.

The part that would otherwise have cost something: a corpus item's `review_commit` is a pull request head, and this repository squash-merges, so **it is never reachable from `main`.** Measured against a fresh clone carrying all 19 branches and 27 tags: every one of the seven frozen commits absent; after fetching `refs/pull/<n>/head` for each, all present. No checkout depth fixes that, because the objects are on no branch. Their bases *were* present, which is the half `fetch-depth: 0` buys. Without the fetch, every item refuses for free and the run reports a tidy nothing — so the lane fetches the refs the manifest names and then asserts each commit arrived.

Four mechanisms bound the spend, none redundant with the others: dispatch-only; a required cap; a per-item `--panel-timeout`; and a job `timeout-minutes` that the preflight *proves* is larger than what the first two permit, recomputed from the real item count so an eighth corpus item cannot silently start getting jobs killed mid-replay.

Two smaller decisions worth surfacing for review:

- **Sharded by replicate, not by item.** `run_id` already means "which replicate this is"; sharding by item would produce one run id per replay, dissolving the grouping reliability is computed over and leaving the cost cap bounding a single item. Each leg writes only under its own `runs/<leg-id>/` and pushes nothing, so two legs never author the same path; one final job commits them together.
- **The dry run lives in a separate run-id space** (`dryrun-` prefix). That is a correctness guard rather than a label: replays resume by skipping items already stored, so a dry run sharing a real run id would make the paid dispatch skip every item and report a complete run built from canned output.

## Linked Issues

None — this follows the eval harness work in `scripts/agent/eval/`.

## Author checklist

## Verification

Skip reason (if applicable):

Measured locally, from the committed tree, serially:

- `agent:tests` — **1462 tests · 0 fail · 1 skip**; baseline at the same sha **1439 · 0 fail · 1 skip** → **+23 tests, no new failures**.
- `actionlint` **1.7.7** clean on the new workflow, and clean on all 15 existing workflows. With `shellcheck` **0.10.0** wired in, the new file adds **0** findings to `main`'s existing 9. Both tools were proved non-vacuous against planted faults — without `shellcheck` on `PATH`, actionlint lints no `run:` body at all, which matters for a workflow that is mostly shell.
- `npx eslint scripts` exits 0.
- **21 of 21 mutations caught.** Two initially passed and were fixed: both assertions were scoped to the whole file and were being satisfied by matching lines in *other* steps — the fetch step's `exit 1` by the staging step's, and the commit step's dry-run gate by the replay step's identical one.
- A full free end-to-end against a pristine CI-shaped clone: preflight → pull-ref fetch → seven real worktrees (2470–3042 files, novelty gate ON for every item) → replay → the cost cap stopping the run as `capped` → a resume re-running nothing → artifact staging → commit, not pushed.

## Risk Assessment

- **User-facing risk:** none for users of the app — this touches no product code and no gating path. For maintainers the risk is **spend**. It cannot start without a human press (a test asserts `workflow_dispatch` is the only trigger), the cost ceiling is a required input with no default, and the workflow offers no way to request an unbounded run even though the CLI permits one. The realistic bad input is a cap set too high; the preflight prints `cap × replicates = maximum exposure` before any leg starts. A cap set too low truncates the run, which is recoverable by re-dispatching the same run id — nothing is paid for twice.
- **Data/security risk:** the workflow's `permissions:` are read-only, and a test asserts no write scope of any kind appears anywhere in the file. Write access lives only in `secrets.EVAL_STORE_TOKEN`, a fine-grained PAT scoped to one other repository. That token is held by the final job only, which runs no model and executes no checked-out code; the replay jobs build a worktree of an arbitrary past pull request and point a model at it, so they read the store unauthenticated (it is public) and persist no git credentials. A test asserts no job holds both credentials. This workflow is not in any `workflow_run` chain and must not be added to one — `ci.yml` → `agent-review-panel.yml` → `capture-collect.yml` is already at GitHub's three-level limit.
- **Rollback plan:** delete the file. Nothing depends on it, no other workflow references it, and it runs only when someone presses a button.

## Notes for Reviewers

- **AI assistance:** this change was written with Claude Code — the workflow, the preflight module, its tests, the task doc and this description. Every number quoted above was produced by running the command rather than estimated, and the end-to-end was executed against a real clone rather than reasoned about.
- **It has never been dispatched.** The GitHub-side behaviour — matrix expansion, the artifact round trip, the push — is reasoned, not observed. The intended first dispatch is the free `panel: stub` mode, which exercises all of it for nothing.
- One operational note that is not obvious from the diff: **a run that stops at its cap is a *prefix* of the corpus, not a sample.** Every replicate stops at the same items in manifest order, so a capped dispatch leaves the tail items with fewer replicates than the head. `run.json` records `status: "capped"` and names the items not replayed; resume before treating the data as complete.
- **Follow-up:** the free scoring lane on a schedule is deliberately not here — it scores stored runs, and neither the runs nor the scorers exist yet. It belongs in a separate file; mixing a free schedule into this one is one edit away from a nightly bill.
- The reasoning behind the five design decisions — required inputs, sharding, commit-vs-artifact, the dry-run design, and whether an environment approval gate is wanted — is in `docs/tasks/active/20260810-eval-replay-lane-todo.md`. The approval-gate one needs a maintainer decision: `CLAUDE_CODE_OAUTH_TOKEN` is an environment secret of `agent`, which eight workflows share, so required reviewers cannot be added there without gating every AI review in the repository. A gate for this lane would need its own environment.

## Summary by CodeRabbit

* **New Features**
  * Added a manually triggered evaluation replay workflow with planning, per-replicate execution, artifact collection, resumable runs, and dry-run support.
  * Added safeguards for cost, timeout, input validation, commit verification, and write access.
  * Added a preflight planner that reports validation errors and replay details before execution.

* **Documentation**
  * Documented replay setup, inputs, safeguards, persistence, failure handling, and known limitations.

* **Tests**
  * Added comprehensive coverage for planning, workflow configuration, permissions, validation, and execution safeguards.

### DIFF
--- a/.github/workflows/eval-replay.yml
+++ b/.github/workflows/eval-replay.yml
@@ -258,10 +258,27 @@ jobs:
 
       # THE STEP THE WHOLE LANE TURNS ON, and the one whose absence is invisible.
       # A squash-merged pull request's head is not reachable from `main`, so these
-      # objects are in no clone of this repository however deep. Fetched by pull ref
-      # rather than by bare sha: GitHub does not enable
-      # `uploadpack.allowReachableSHA1InWant`, so `fetch <sha>` is refused, while the
-      # pull ref always reaches the frozen commit.
+      # objects are in no clone of this repository however deep.
+      #
+      # THE PULL REF IS THE PRIMARY PATH AND A BARE SHA IS THE FALLBACK, in that order
+      # and for different reasons. `refs/pull/<n>/head` is a documented, stable GitHub
+      # feature, and one batched fetch serves every item — so it is what the lane asks
+      # for first. But it is a MOVING pointer: it tracks the pull request's current
+      # head, and a frozen `review_commit` that was later force-pushed away is no
+      # longer contained in it. Measured on the pilot's pr-415, whose corpus commit and
+      # whose pull-ref tip have diverged outright.
+      #
+      # So anything the pull ref did not carry is asked for by its full sha, from the
+      # same source repository. That works here — verified: this server does serve
+      # commits unreachable from every ref — but it rests on server configuration
+      # rather than on a documented contract, which is exactly why it is the second
+      # attempt and not the first.
+      #
+      # NO `--depth` on that fallback, and this is load-bearing rather than tidy: a
+      # shallow fetch marks the repository shallow, and a shallow tree cannot be
+      # blamed — which would silently disable the novelty gate this lane exists to
+      # exercise. Verified that the undepthed form leaves the clone unshallow and that
+      # `git blame` still works in a worktree built at the fetched commit.
       #
       # FETCHED FROM THE REPOSITORY THE CORPUS NAMES, NOT FROM `origin`. A corpus item
       # is a pull request OF a named repository — that is what `source_pr` means — and
@@ -289,6 +306,17 @@ jobs:
           source_url="$(cat "$RUNNER_TEMP/plan/source-repo-url.txt")"
           echo "eval-replay: fetching the corpus's refs from ${source_url}"
           xargs git fetch --no-tags "$source_url" < "$RUNNER_TEMP/plan/refspecs.txt"
+          # The second attempt, for a frozen commit its pull request has moved past.
+          # `|| true`: a failed retry falls through to the assertion below, so there is
+          # exactly ONE place that decides the step failed and exactly one message
+          # describing it.
+          while read -r commit item; do
+            if git cat-file -e "${commit}^{commit}" 2>/dev/null; then
+              continue
+            fi
+            echo "  retry   ${item} ${commit} — not carried by its pull ref, asking for the commit directly"
+            git fetch --no-tags "$source_url" "${commit}:refs/eval/item/${item}" || true
+          done < "$RUNNER_TEMP/plan/commits.txt"
           missing=0
           while read -r commit item; do
             if git cat-file -e "${commit}^{commit}" 2>/dev/null; then
@@ -299,8 +327,8 @@ jobs:
             fi
           done < "$RUNNER_TEMP/plan/commits.txt"
           if [ "$missing" -ne 0 ]; then
-            echo "eval-replay: ${missing} corpus commit(s) are not in this checkout after fetching the pull refs." >&2
-            echo "The pull request may have been force-pushed since the corpus was frozen, or its refs deleted." >&2
+            echo "eval-replay: ${missing} corpus commit(s) are not in this checkout after BOTH the pull-ref fetch and a direct fetch by sha." >&2
+            echo "The commit may have been garbage-collected upstream, or ${source_url} may not be the repository the corpus was frozen from." >&2
             exit 1
           fi
 
--- a/.github/workflows/eval-replay.yml
+++ b/.github/workflows/eval-replay.yml
@@ -1,0 +1,457 @@
+# Replay the frozen eval corpus through the real review panel, in a clean CI box.
+#
+# THIS IS THE ONLY WORKFLOW IN THIS REPOSITORY THAT SPENDS MODEL BUDGET ON DEMAND.
+# Every other job here is free or fails free. This one exists to call
+# `review-panel.mjs` over a frozen pull request, K times, and each of those calls is
+# a real bill. So the question asked of every line below is not "does this work" but
+# "what does this cost when it goes wrong with nobody watching" — and the answers are
+# four separate mechanisms, none of them redundant with the others:
+#
+#   workflow_dispatch ONLY   nothing but a human press can start it
+#   a REQUIRED cost cap      per replicate, no default, refused if absent
+#   --panel-timeout          one item cannot run forever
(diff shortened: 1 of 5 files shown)


dlgpdmsly2:
Pushed an update while this is still unreviewed — one substantive fix, found by checking who can actually press the button.

**The corpus refs were being fetched from `origin`, and that is wrong independent of any fork question.** A corpus item is a pull request *of a named repository* — that is what `source_pr` means, and every manifest already carries `source_repo` to say which. Reading the refs from `origin` instead ties a measurement to whichever checkout happens to be running it. The coupling is invisible until it bites, and it bites hard: a fork carries none of its parent's `refs/pull/*`. Measured today — `wafflebase/wafflebase` has all seven pilot refs, a fork has one, its own — and on a fork-origin checkout the old form dies with `fatal: couldn't find remote ref refs/pull/415/head`, after which every item would refuse for free and the run would look exactly like a panel that found nothing.

The remote now comes from the manifest, and a manifest with no usable `source_repo` is refused rather than guessed at.

**A second thing fell out of checking that: the bases are not carried by the pull refs.** I had assumed a `review_base` would be an ancestor of its own pull head. It is for pr-524 and pr-605 and **is not for pr-415**, so the bases come from `main`'s history. The lane now fetches the source repository's `main` too, which also stops base resolution depending on how fresh the running checkout's history is.

Verified on a simulated fork-origin checkout: the old form fails as above; the new one fetches all seven refs plus `main`, all seven bases resolve, and a stub replay completes with the novelty gate ON.

Re-verified from the committed tree at `2efe86c78`: **1471 tests · 0 fail · 1 skip** against a freshly measured baseline of **1446 · 0 fail · 1 skip** (+25). `actionlint` 1.7.7 + `shellcheck` 0.10.0 clean; `eslint` 0. **24 of 24 mutations caught**, including three new ones covering exactly this change — reverting the fetch to `origin`, defaulting `source_repo` instead of refusing it, and dropping `main` from the refspecs.

dlgpdmsly2:
Correction pushed as `36d1349`, on top of the branch update rather than over it.

**The comment on the fetch step was wrong.** It claimed GitHub does not enable `uploadpack.allowReachableSHA1InWant`, so `git fetch <url> <sha>` would be refused. That is false on this server — verified in a virgin `git init` repo with only a remote: fetching `51c01826`, a commit unreachable from *every* ref in this repository, exits 0 and the object arrives.

That correction changes the design rather than just the prose, because **a pull ref is a moving pointer.** `refs/pull/<n>/head` tracks the pull request's current head, so a frozen `review_commit` that was later force-pushed away is not contained in it. On pr-415 the corpus commit and the pull-ref tip have diverged outright, so the fetch-then-assert step would have refused that item and failed the whole dispatch.

So anything the pull refs did not carry is now requested by full sha, from the same repository the manifest names, ahead of an **unchanged** assertion — the retry is a second attempt, not a second failure point. The pull ref stays primary: documented, and one batched fetch serves every item, whereas the bare-sha path depends on server configuration rather than a contract.

**The fallback is deliberately not shallow**, and that one is load-bearing. `--depth=1` marks the repository shallow, a shallow tree cannot be blamed, and the novelty gate blames — so a shallow retry would leave the gate silently off on exactly the item that needed rescuing. Verified that the undepthed form leaves the clone unshallow and that `git blame` works in a worktree built at the fetched commit.

Demonstrated with pr-415 re-pinned to the stranded commit:

```
without the fallback:  MISSING pr-415 51c01826…  -> dispatch fails
with the fallback:     retry pr-415 … -> [new ref] 51c01826… -> refs/eval/item/pr-415
                       all 7 ok · shallow=false
```

Re-verified from the committed tree: **1493 tests · 0 fail · 1 skip** against a freshly measured baseline of **1468 · 0 fail · 1 skip** (+25). `actionlint` 1.7.7 + `shellcheck` 0.10.0 clean, `eslint` 0. **27 of 27 mutations caught** — three new ones cover this change (drop the retry, make it shallow, point it at origin), and a fourth fix was needed because the shallow mutation was initially caught by a neighbouring regex rather than by the guard that means it.

harrykim8672 (approved):

PR #743, opened 2026-08-10
Let the verifiers bound the claim, not just confirm it

hwisoo-kim:
> ⚠️ **This is an unvalidated hypothesis.** It encodes a belief about *why* the hunter's titles come out wrong; nothing yet shows it works. A cheap known-answer test exists — see the bottom — and I would run it before merging.

## The problem

The explorer's `title` is the one piece of model prose that reaches a report **unexamined**, and every real defect this hunter has found was described more broadly than its own evidence. Each was a **true observation wrapped in a false universal**:

| the hunter wrote | what is actually true |
|---|---|
| "Bold only removes bold — it can never apply it" | Bold applies fine to ordinary text |
| "after a scroll, clicks no longer select any cell" | Clicks work; the cells had scrolled offscreen |
| "never removes italic from a right-to-left selection" | It removes correctly when the preceding run shares the style |

Filed as written, each would have sent a maintainer after a bug that does not exist.

The third one is the sharpest: **both verifiers confirmed it at high confidence, citing exactly the right line.** They validated the observation and never examined the claim — because nothing asked them to.

## The fix, which the design had already invented once

`groundedIn` routes the code **location** through the verifier, precisely because the explorer cannot be trusted to get it right. That principle was never extended to the **claim**.

- `UI_VERIFIER_SCHEMA` gains a required **`scopedTitle`** — the defect restated no more broadly than the evidence and the source support.
- The report **headlines that**, with the hunter's original shown underneath (only when they differ) so a filed issue can be traced back to its run.
- **`overclaimed` is recorded but is NOT a gate input.** Overclaiming is a description defect, not a validity one — letting wording refute a finding would discard real defects, and all three above were overclaimed *and* real. It is counted so a drift stays visible.

The prompt names those three failures **verbatim** rather than asking for precision in the abstract, and points at the words where this goes wrong: *never, always, any, all, cannot*.

## Test plan

- `cd scripts/agent && node --test '**/*.test.mjs'` with `node_modules` hidden — **1445 pass / 0 fail / 1 skipped**
- `pnpm lint:scripts` clean; `hunt-ui.mjs preflight` ok
- **Mutation-tested:** the report headlines the scoped claim (not the hunter's); `scopedTitle` comes from the verifiers rather than falling through; `overclaimed` requires an explicit `true` so a missing field cannot inflate the count.
- One test replays the **actual** over-claim from the last live run verbatim, and asserts the scoped heading precedes the original in the rendered report.

## The known-answer test I would run before merging

Same controlled setup that validated #741 — same persona, same frozen issue corpus, same budgets, **only the verifier prompt changed**. If the title comes back as something like *"a toggle reads the run preceding a right-to-left selection"* rather than *"never removes italic"*, the fix works. ~$4.

Worth being blunt about why: twice in this effort I proposed a confident mechanism that measurement then contradicted (the verifier-ceiling "prompt growth" theory, and the "raising the ceiling treats a symptom" reasoning). A prompt change that does not change behaviour is worse than no change, because it reads as solved.

## Summary by CodeRabbit

* **New Features**
  * Reports now use evidence-based, verifier-scoped claim titles.
  * Original claim wording is preserved when a finding has been narrowed.
  * Reports track and identify claims that extend beyond the demonstrated evidence.
  * Verification now captures relevant preconditions and limits findings to supported evidence.

* **Tests**
  * Added coverage for scoped titles, overclaim detection, fallback behavior, and report rendering.


harrykim8672:
@claude review

harrykim8672 (approved):

PR #763, opened 2026-08-10
Fix formula reference XSS

taejinchoi-cbnu:
## Summary

Fix formula token rendering by escaping `REFERENCE` and `NUM ` token text before inserting it into `innerHTML()`.

https://github.com/user-attachments/assets/31068e0c-6cc9-4b59-898b-6e7a434bf4e9

Reproduction: In the current Wafflebase build, entering `=''!A1` into a sheet cell results in JavaScript execution via the injected onerror handler.

## Why

`=''!A1` is parsed as a single `REFERENCE` token.
Since the token text is inserted without escaping, the browser interprets the `` markup as a real DOM element and executes the onerror handler.

The issue may lead to XSS because formula values can also reach the document without going through the contenteditable editing path, including grid paste, spreadsheet file imports (e.g. XLSX, JSON, CSV/TSV), and the v1 cells API.
If such values are later rendered through the formula highlighting path, the browser can interpret them as HTML.

`NUM` cannot carry HTML with the current grammar, but it uses the same unescaped interpolation pattern.

Escaping it as well keeps both branches safe and avoids relying on the grammar remaining unchanged.

The fix uses the existing `escapeHTML()` helper already used by the other token types. No rendering structure or storage format changes are required.

## Linked Issues

Fixes #

## Author checklist

## Verification

Skip reason (if applicable):

## Risk Assessment

- User-facing risk: No functional behavior changes are expected. Formula highlighting and caret positioning remain unchanged.
- Data/security risk: This removes the XSS vector at render time. No storage format or CRDT schema changes are required, and existing stored values are safely escaped when rendered.
- Rollback plan: Revert the commits. No migration is required.

## Notes for Reviewers

- UI changes (screenshots/gifs if applicable): No UI changes
- Follow-up work (if any): none

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formula highlighting to safely display cell references, numbers, and sheet names containing HTML characters.
  * Prevented formula text from being interpreted as unintended HTML elements.
  * Preserved HTML entities as literal text in highlighted formulas.

* **Tests**
  * Added coverage for safe formula rendering in cell inputs and the formula bar.

### DIFF
--- a/packages/sheets/src/view/cellinput.ts
+++ b/packages/sheets/src/view/cellinput.ts
@@ -336,14 +336,14 @@ export class CellInput {
     for (const token of tokens) {
       if (token.type === 'REFERENCE') {
         contents.push(
-          `<span style="color: ${getFormulaRangeColor(this.theme, refIndex)}">${token.text}</span>`,
+          `<span style="color: ${getFormulaRangeColor(this.theme, refIndex)}">${escapeHTML(token.text)}</span>`,
         );
         refIndex++;
         continue;
       }
       if (token.type === 'NUM') {
         contents.push(
-          `<span style="color: ${this.getThemeColor('tokens.NUM')}">${token.text}</span>`,
+          `<span style="color: ${this.getThemeColor('tokens.NUM')}">${escapeHTML(token.text)}</span>`,
         );
         continue;
       }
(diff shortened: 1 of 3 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #767, opened 2026-08-11
Have the fixer report before it pushes, not after

hwisoo-kim:
## Summary

The fix agent's report goes missing on most rounds, and it is a race the fixer
runs against itself.

The push is what ends the job. Pushing re-triggers CI, CI re-triggers the review
panel, and `.github/workflows/agent-review-panel.yml` is `cancel-in-progress` —
so the fresh panel run cancels the run the fixer is still inside, about half a
minute later. Anything the fixer does *after* pushing is racing a cancellation it
caused itself.

**#757 measured it.** The fixer pushed on all three rounds and reported on one:

| round | commit | fix job cancelled | report |
|---|---|---|---|
| 1 | 07:54:45 | 07:55:19 (+34s) | lost |
| 2 | 08:47:14 | 08:47:48 (+34s) | **posted 08:47:41 — 7s of margin** |
| 3 | 09:32:58 | 09:33:29 (+31s) | lost |

Round 2 is the evidence that nothing is broken: the agent reports willingly, 27
seconds after committing. It simply does it after the push, and two times out of
three the axe falls first.

This also retro-explains the original symptom on **#737** and **#695**, where a
fix report was missing for rounds the fixer plainly ran. That was read as model
discretion at the time; it was this race.

### Ruled out first

- **Not out of turns.** The fixer pushed a commit on all three rounds, so it had
  budget left to work with.
- **Not the `readRebuttals` change from #747.** That merged at 06:40, before all
  three rounds, and round 2 succeeded under the same code — a systematic code
  break cannot produce a pattern where only the middle round works.

### The change

Ordering, not machinery: **commit, report, then push.** Nothing can cancel work
done before the push that causes the cancellation, and
`scripts/agent/fix-report.mjs`'s `--head` is the SHA the fixer acted *on*, so the
report is already complete before the new commit exists.

The prompt now states it twice — where the report is described, and again where
the push is. Both are needed because the agent reads prose, and `THEN REPORT`
sitting above a push section reads as *report last*.

`.github/workflows/agent-fix.yml` (the on-demand `@claude fix`) is **not** exposed
and is deliberately unchanged: it declares no workflow-level concurrency, so the
panel run its push triggers cancels nothing, and its prompt already ordered the
report first.

## Test plan

- `scripts/agent/fix-report.test.mjs` pins the ordering: the report **and** any
  rebuttal must precede the push instruction, and the ordering must be *stated*
  rather than merely arranged.
- **Two mutation tests, both caught** — removing the `BEFORE PUSHING` statement,
  and removing the `PUSH LAST` restatement, each turn the suite red.
- Full `scripts/agent` lane green; `pnpm lint:scripts` and `pnpm verify:entropy`
  clean.

### What this PR cannot verify about itself

Fork PRs skip the panel, so the prompt this changes will not run here. The check
is the next autonomous agent PR: every round that pushes a commit should now also
carry a `### 🛠️ Fix agent report`, and the loop-status `Fixer` column should read
a real count on every row rather than `🔧 dispatched`.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -1662,11 +1662,22 @@ jobs:
             - Run `pnpm verify:fast` ONCE at the end to confirm green — not after
               every edit.
 
-            THEN REPORT WHAT YOU DID. Required, and not paperwork: the next round
-            reads this report and hands each claim to an independent adjudicator,
-            so a finding you fixed in a way the re-check cannot see is re-raised
-            unless you say what you changed. Pass one `--fixed` or `--skipped` per
-            checklist item, so the report accounts for the whole work list.
+            REPORT WHAT YOU DID — AFTER COMMITTING, BEFORE PUSHING. Required, and
+            not paperwork: the next round reads this report and hands each claim to
+            an independent adjudicator, so a finding you fixed in a way the
+            re-check cannot see is re-raised unless you say what you changed. Pass
+            one `--fixed` or `--skipped` per checklist item, so the report accounts
+            for the whole work list.
+
+            THE ORDER IS LOAD-BEARING, and reporting after the push loses the
+            report. Your push re-triggers CI, CI re-triggers the review panel, and
+            this workflow is `cancel-in-progress` — so the new panel run CANCELS
+            the run you are still inside, killing this job about half a minute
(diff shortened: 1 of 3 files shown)


harrykim8672:
@claude review

harrykim8672 (approved):

PR #769, opened 2026-08-11
Source the agent pipeline from wafflebase/agent-pipeline@v0.1.1

hwisoo-kim:
## Summary

The pipeline's scripts now live in their own repo, `wafflebase/agent-pipeline`. This re-points every workflow that sourced them from **this** repo's `main` to source them from there instead, pinned to the immutable annotated tag `v0.1.0`.

The scripts are byte-identical to `scripts/agent/` as of this branch's base — verified file by file before the branch was cut. Nothing in `scripts/agent/` is deleted here; it simply stops being what runs.

## Why

Until the pipeline has one home, every fix to it lands twice and two copies of ~24k lines diverge quietly. Re-pointing the checkouts is the smallest change that closes that, and it is the step whose rollback is a single revert.

## Why the adapter steps

In its own repo the pipeline lives at `packages/pipeline`, not `scripts/agent`. A plain re-point would have moved **117 invocation paths** across these ten files — and most of those invocations are guarded like `[ -f … ] || { echo "skipping"; exit 0; }`, so a wrong path does not fail a job, it **silently skips a step**.

So each re-pointed checkout lands in `.pipeline-src` and is then moved to exactly where the steps already look:

```yaml
- uses: actions/checkout@v4
  with:
    repository: wafflebase/agent-pipeline
    ref: v0.1.0
    path: .pipeline-src
    sparse-checkout: packages/pipeline
    persist-credentials: false
- name: Place the trusted pipeline where the steps expect it
  run: |
    mkdir -p .trusted/scripts
    mv .pipeline-src/packages/pipeline .trusted/scripts/agent
    rm -rf .pipeline-src
```

**Zero invocation paths change in this PR.** The whole diff is checkout blocks. These adapters are deleted when the workflows themselves become reusable and everything moves to `$RUNNER_TEMP/pipeline` — so the paths get renamed once, not three times.

18 checkout sites across 10 workflows. Every one of the 6 remaining `ref: main` sites was the *first* checkout in its job, so switching from a root checkout to `path:` + move loses no workspace-cleaning behaviour.

## Verification

- `agent:tests` green: **1576 tests, 1570 pass, 0 fail** (6 pre-existing skips), run the way CI runs it — recursive glob, `node_modules` absent.
- All ten workflows parse as YAML.
- Audited: no line matching `node …scripts/agent`, `cp -R`, `[ -f `, or `--lenses` appears in the diff.
- `v0.1.0` confirmed byte-identical to this branch's `scripts/agent/` across all 41 moved files.

## Risk Assessment

The trusted-source property is unchanged. `workflow_run` still guarantees the default-branch copy of these files executes, so the pin cannot be altered by a pull request. The pin is a literal for now; once the workflows themselves move out, an OIDC `job_workflow_ref` assertion replaces it.

`v0.1.0` is protected by a ruleset (`deletion`, `non_fast_forward`, `update`) — mutation-tested: a force-move and a delete were both rejected by the remote.

Rollback is `git revert` of this one commit.

## Not covered

`agent-implement`'s post-run steps still run **this** repo's copies of `session-job-summary.mjs` and `metrics.mjs`. That job's workspace is the agent's own editing tree, and `classify.mjs` stays here regardless, so splitting it needs a second checkout in that job. It belongs with the reusable-workflow port rather than here.

## Notes for Reviewers

Authored with Claude Code assistance; I have reviewed every hunk. The mechanical transformation asserted an exact match count per site class and aborted on any miss, because a silently-unmatched site is precisely the failure mode this PR is shaped to avoid.

Worth a close look at: the six `path:`-vs-root checkout conversions (that they really are first-in-job), and the `checks.test.mjs` anchor change — the ordering assertion it guards is still the one that matters, but it now keys on the step name rather than a `ref: main` literal that no longer exists.

## Summary by CodeRabbit

- **Chores**
  - Standardized automated workflows on a consistent, pinned pipeline version.
  - Improved consistency and reliability across fixing, review, iteration, routing, summarization, and smoke-test processes.
  - Preserved existing workflow behavior while limiting retrieved pipeline content to what each process requires.

- **Tests**
  - Added validation to ensure workflow pipeline references use immutable commit identifiers.
  - Updated checkout-order checks to reflect the revised workflow process.

### DIFF
--- a/.github/workflows/agent-fix.yml
+++ b/.github/workflows/agent-fix.yml
@@ -47,7 +47,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           repository: wafflebase/agent-pipeline
-          ref: v0.1.1
+          ref: 9b02a392a4f89239f6963ace00e2705d08ad2181 # v0.1.1
           path: .pipeline-src
           sparse-checkout: packages/pipeline/command.mjs
           sparse-checkout-cone-mode: false
@@ -118,7 +118,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: wafflebase/agent-pipeline
-          ref: v0.1.1
+          ref: 9b02a392a4f89239f6963ace00e2705d08ad2181 # v0.1.1
           path: .pipeline-src
           sparse-checkout: packages/pipeline
           sparse-checkout-cone-mode: false
--- a/.github/workflows/agent-fix.yml
+++ b/.github/workflows/agent-fix.yml
@@ -46,9 +46,22 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          sparse-checkout: scripts/agent/command.mjs
+          repository: wafflebase/agent-pipeline
+          ref: v0.1.1
+          path: .pipeline-src
+          sparse-checkout: packages/pipeline/command.mjs
           sparse-checkout-cone-mode: false
           persist-credentials: false
+      # ADAPTER (transitional). The pipeline now lives in its own repo at
+      # packages/pipeline; put it where every step below already looks for it so
+      # this change touches checkout blocks only. Phase D removes this together
+      # with the scripts/agent path itself.
+      - name: Place the trusted pipeline where the steps expect it
+        run: |
+          set -euo pipefail
+          mkdir -p scripts
+          mv .pipeline-src/packages/pipeline scripts/agent
+          rm -rf .pipeline-src
       # `fix` on an ISSUE starts agent-implement.yml; that workflow gates on
       # `!github.event.issue.pull_request` and this one on the presence of it, so
       # the same verb reaches exactly one of them and never both.
@@ -101,11 +114,25 @@ jobs:
       # builder below both run from here; the PR branch is checked out over this
       # tree later, which is why every decision must be made — and captured into
       # step outputs — before that happens.
-      - name: Check out trusted main
+      - name: Check out the trusted pipeline
         uses: actions/checkout@v4
         with:
-          ref: main
+          repository: wafflebase/agent-pipeline
+          ref: v0.1.1
+          path: .pipeline-src
+          sparse-checkout: packages/pipeline
+          sparse-checkout-cone-mode: false
           persist-credentials: false
+      # ADAPTER (transitional). The pipeline now lives in its own repo at
+      # packages/pipeline; put it where every step below already looks for it so
+      # this change touches checkout blocks only. Phase D removes this together
+      # with the scripts/agent path itself.
+      - name: Place the trusted pipeline where the steps expect it
+        run: |
+          set -euo pipefail
+          mkdir -p scripts
+          mv .pipeline-src/packages/pipeline scripts/agent
+          rm -rf .pipeline-src
 
       # Human-facing comments go through the App token so the reply comes from the
       # agent rather than github-actions[bot], matching every other verb.
--- a/.github/workflows/agent-implement.yml
+++ b/.github/workflows/agent-implement.yml
@@ -54,7 +54,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           repository: wafflebase/agent-pipeline
-          ref: v0.1.1
+          ref: 9b02a392a4f89239f6963ace00e2705d08ad2181 # v0.1.1
           path: .pipeline-src
           sparse-checkout: packages/pipeline/command.mjs
           sparse-checkout-cone-mode: false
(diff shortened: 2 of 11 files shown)


hwisoo-kim:
Rebased onto `main` (now includes #767, which touches this same workflow's fixer prompt — no textual conflict, the edits are in different regions) and **re-pinned `v0.1.0` → `v0.1.1`**, all 18 sites.

The bump is not cosmetic: `v0.1.1` syncs the central repo with #759 (`finding-match.mjs` rewrite) and #767. Merging this while pinned at `v0.1.0` would have made wafflebase run *older* pipeline scripts than the copy in its own tree — the exact drift this PR exists to remove.

Re-verified after the rebase:
- `agent:tests` green — **1611 tests, 1605 pass, 0 fail** (6 pre-existing skips), recursive glob with `node_modules` absent.
- All ten workflows parse; 18/18 pins on `v0.1.1`, zero stale `v0.1.0` references.
- #767's "THE ORDER IS LOAD-BEARING" prompt change survived the rebase intact.
- `v0.1.1` re-verified byte-identical to `scripts/agent/` at this branch's base, across the whole moved set rather than just the files known to have changed.

hwisoo-kim:
Verified the finding against current code and **applied it** — all 18 sites now `ref: 9b02a392a4f89239f6963ace00e2705d08ad2181 # v0.1.1`.

**Why it holds even though the tag is protected.** `v0.1.1` *is* covered by a ruleset on `refs/tags/v*` (`deletion`, `non_fast_forward`, `update`) which I mutation-tested — a force-move and a delete were both `remote rejected`. So the tag is not actually movable today. But that guarantee lives in a repository setting: invisible to anyone reading these files, and changeable by an org admin. A 40-hex SHA cannot be re-pointed by anyone.

What made it decisive rather than stylistic: in **this** phase nothing re-verifies that the code fetched is the code intended. The OIDC `job_workflow_ref` assertion only becomes available once these workflows are themselves reusable, so until then the pin is the entire guard. And it is the rule this repo already applies to `create-github-app-token` and `claude-code-action`, with comments saying exactly why.

**One deliberate deviation on the test.** The suggestion was to make `checks.test.mjs` require that specific SHA. I asserted the **shape** of the pin instead — every `wafflebase/agent-pipeline` checkout across every `agent-*.yml` must pin 40 hex characters:

- Pinning the literal SHA needs editing in lockstep with every version bump, and a bumper updating both files learns nothing.
- What actually needs catching is a future bump that quietly reverts to a tag — invisible in a diff, and silent at runtime.
- Mutation-tested both ways: green at 18/18, fails when a single site reverts to `ref: v0.1.1`, green again when restored.

Say the word if you would rather have the literal SHA asserted too and I will add it alongside.

Re-verified: **1612 tests, 1606 pass, 0 fail** (6 pre-existing skips); all ten workflows parse; 18/18 SHA pins with zero bare tag refs.

hwisoo-kim:
Verified both parts of the finding against the code before changing anything. **Both were real; both fixed** — `scripts/agent/checks.test.mjs` only, no workflow touched.

**The silent one, and the more serious.** The repository pattern ended in `\s*$`, so `repository: wafflebase/agent-pipeline # central` did not match — and a line that does not match is not scanned at all, neither counted nor reported. A checkout could be added, or an existing one annotated, and quietly leave the guard's view. A guard whose own evasion is silent is exactly the failure shape it was written to catch, so this mattered more than it looks.

**The other.** `[0-9a-f]{40}\b` accepted anything merely *starting* with 40 hex characters — `<sha>/branch` and `<sha>-evil` both passed. (41+ hex already failed closed: `\b` cannot match between two hex digits.) Both patterns now end in `\s*(#.*)?$` — trailing whitespace and a YAML comment, nothing else. A quoted ref is rejected too; every one of these blocks comes from one template, so pinning one canonical form is the point.

Mutation-tested end to end against the real workflow files:

| case | result |
|---|---|
| baseline, 18/18 SHA-pinned | green |
| one site → `<sha>/branch` | **fails** |
| tag hidden behind a repository comment | **fails** (passed before) |
| restored | green |

**Also, a heads-up on the branch.** `main` was merged into it while I was working (`4058630e5`, 8 commits). I rebased my change on top of that merge rather than force-pushing over it, so nothing was discarded and this was a fast-forward. Two things I checked while there:

- None of those 8 commits touch the moved pipeline set — all `eval/` (measurement), `hunt-ui-tool`, or `ci.yml`, which stays with the consumer by design. So the `v0.1.1` pin is still current and needs no bump.
- **#755 adds a `merge_group` trigger to `ci.yml`.** Worth a thought beyond this PR: a merge queue changes *when* `workflow_run` fires, and the review panel's trigger contract (`requested` + `completed` with `run_attempt > 1`) is built on the current behaviour. Not a blocker here, but it should be checked before a queue is actually enabled.

Suite on the merged base: **1661 tests, 1655 pass, 0 fail** (6 pre-existing skips). 18/18 SHA pins intact.

harrykim8672 (approved):

PR #775, opened 2026-08-11
Strip the variable that is actually set, and retract #772's claim

hwisoo-kim:
## Summary

**#772 was wrong and this retracts it.** It claimed to fix the intermittent
`agent:tests` failure and blamed `NODE_CHANNEL_FD` — but that variable is never set
here, so the strip was inert, and the failure recurred on the very next CI run of the PR
it was meant to unblock (#766).

Measured on Node 22.23.2, a test-file process and everything it spawns carry exactly one
`NODE_*` variable, and it is a different one:

```
PARENT NODE_*: ["NODE_TEST_CONTEXT"]
CHILD  NODE_*: ["NODE_TEST_CONTEXT"]    # spawned with env: process.env
CHILD  fds:    [0,1,2,3,4,5]            # fd 3 — the runner's channel — inherited
```

`NODE_TEST_CONTEXT=child-v8` is how Node tells a process *"you are a test file; report
over the V8-serialized channel."* So every Node process this adapter spawns was told it
is a test file **and** handed the fd that claim refers to. Nothing wants either: no panel
loads `node:test`, and the adapter reads the panel over stdout/stderr pipes.

That is worth closing on its own terms — two processes believing they own one protocol
stream is a defect whether or not it has been caught writing.

## What this does NOT claim

It is **not** claimed to fix the deserialize failure:

```
not ok 19 - eval/run.test.mjs
  error: 'Unable to deserialize cloned data due to invalid or unsupported version.'
  stack: #processRawBuffer (node:internal/test_runner/runner)
```

That is still open, and the docblock now says so instead of asserting a cause. It has not
reproduced locally in **15 runs** across Node 22 and 24, at CI's concurrency, on both
`main` and the branch that hit it twice. This leak is ruled *in* as a hazard, not proven
as the cause — and the comment is the record of that distinction.

## Test plan

The old test could not have caught any of this: it built its own env with
`NODE_CHANNEL_FD` set and asserted the strip removed it, which passes against a strip
that does nothing to the *real* environment. That is the same shape of mistake as the fix
itself — a guard proving something about its own fixture.

It now asserts against `process.env` directly, and first checks the runner is advertising
itself, so it cannot go vacuous unnoticed. Mutation-tested: dropping the
`NODE_TEST_CONTEXT` delete fails with `the panel must not be told it is a test file`
(22 pass / 1 fail); dropping the `NODE_CHANNEL_FD` delete fails too.

Full suite on Node 22, reproduced as CI runs it (recursive glob, `scripts/agent/node_modules`
absent): **1611 pass / 0 fail**. `lint:scripts` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Evaluation panels now launch with a clean test-runner environment, preventing leaked Node.js context variables from affecting execution.
  * Existing environment settings, including `PATH`, remain preserved.
  * The calling process environment is no longer altered.

* **Tests**
  * Added coverage for live test-runner environments, environment cleanup, preservation of required settings, and empty or missing input cases.

### DIFF
--- a/scripts/agent/eval/adapters/reviewer.mjs
+++ b/scripts/agent/eval/adapters/reviewer.mjs
@@ -99,35 +99,46 @@ const INTERRUPTS = Object.freeze(["SIGINT", "SIGTERM"]);
 const IS_POSIX = process.platform !== "win32";
 
 /**
- * The panel's environment, minus the parent's IPC channel.
+ * The panel's environment, minus anything that tells a child it belongs to the test
+ * runner.
  *
- * `NODE_CHANNEL_FD` is how a Node process is told "you have an IPC channel on this
- * fd". Nothing here wants one — no panel calls `process.send`, and the adapter reads
- * the panel over stdout/stderr pipes. But `runAgent` defaults `env` to `process.env`,
- * and under `node --test` the process doing the spawning IS a test-runner child, which
- * has that variable set: the runner talks to each test file over an IPC channel with
- * advanced serialization. So the panel — and the grandchild it spawns with
- * `stdio: "inherit"` — booted believing they shared that channel, and whatever their
- * runtimes wrote into it landed in the middle of the runner's message stream.
+ * WHAT IS MEASURED. Under `node --test`, the process that spawns the panel IS a
+ * test-runner child, and `runAgent` defaults `env` to `process.env`. Probed on Node
+ * 22.23.2, that environment carries exactly one `NODE_*` variable:
  *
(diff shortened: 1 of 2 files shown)


hwisoo-kim:
Closing in favour of a straight revert of #772 (see the revert PR). #772's premise was wrong — `NODE_CHANNEL_FD` is never set in this environment, so it was inert — and since the flake is still undiagnosed, carrying any fix that asserts a cause is worse than carrying none. The `NODE_TEST_CONTEXT` leak this PR found is real, but it should land with a demonstration rather than a hypothesis.

PR #776, opened 2026-08-11
Add inject.mjs's layout half, the first module here that writes

hotsunchip:
## Summary

Ports layout mutation from the prototype: props/class/text edits on one node, subtree insert and remove, import maintenance, and the multi-hunk diff. Both section banners are in place, so the token half (CVA rewrites, token values, `@theme inline` aliases) lands next as a pure fill-in.

**This is the first module here that writes.** `jsx-nodes`, `stamp` and `extract` only read — their worst failure shows the designer something wrong. This one edits a file in someone's working tree.

### The split is not where the plan said

The layout half is the only part that *imports* `jsx-nodes.mjs`, but it also *calls* six helpers living in the token half — `spliceSpan`, `indentOf`, `removeNodeLine`, and `rewriteClassLiteral` + its two class-token helpers. **119 lines this PR has to carry.** They're genuinely shared (`rewriteClassLiteral` serves both `applyLayoutProps` and section A's `applyClassRewrite`), so they sit in a shared section above both banners rather than being duplicated. PR 7 is correspondingly smaller.

### Commit 1 fixes data loss, not a missing capability

`requireStatic` refused three things, but only one depends on the node alone. The other two are about splicing *this node* into or out of a sibling list — and neither hazard exists when the node is merely receiving a child.

Refusing anyway made `applyLayoutInsert` unable to target the root element of any component whose body is `return <div>…</div>`. I found it by probing the involution before writing assertions: **remove `<Beta>` succeeded, and the verbatim re-insert was refused.** The node is gone with no way back through this API.

`role` names what the anchor is to the caller's op:

| guard | `'target'` (default) | `'container'` |
|---|---|---|
| `scope !== 'static'` | refuse | **refuse** — renders N times either way |
| `owner !== node` | refuse | allow |
| is a returned expression | refuse | allow |

Checked against the parser rather than reasoned about: a child spliced into a returned root parses, a child into a conditionally-rendered element parses, and a *sibling* beside a returned root is a syntax error. Every existing caller keeps today's behaviour via the default.

This is the widening #718 deferred with a FOLLOW-UP note, landing with the consumer that justifies it. CodeRabbit flagged the consequence at the time — it was right about the cost, and the fix genuinely needed the op, which only the injector defines.

### A second unapplicable inverse, found by the test table

Writing the involution cases surfaced the same shape from the other direction: removing a direct child of a **returned fragment** succeeded, and the re-insert had no parent to target. The fragment is transparent, so its children sit at depth 1 under the synthetic `#returns` container, which no anchor can name. Now refused. That costs a real capability, which is the right side to err on. The fix — making the returns root addressable as a container — is named in the code and §5.7.

### The re-keying guardrail

`findJsxRoots` was fixed for prototype pollution in #718; `analyzeNodes` then copied its entries into a plain `{}` and undid it one layer down, written by the author who had just added the comment explaining the hazard. That's the argument against a helper you must remember to call.

`name-keyed-maps.test.mjs` enumerates every export returning a map keyed by consumer identifiers and asserts the invariant against hostile component names (`toString`, `valueOf`, `__proto__`). A new export evades it only by omitting its row — visible in review — and the table's own length is asserted so an empty one can't pass vacuously. Scoped entirely to this package's own suite; nothing scans the monorepo.

`analyzeClasses`'s `antiPatterns` is deliberately **exempt and pinned as such**: it is prototyped, and safe, because its keys come from a closed vocabulary in our own source. The rule is about who chooses the key, not which constructor built the object.

## Test plan

- `pnpm --filter @wafflebase/design-editor test` — **250 passing**, 66 new
- `pnpm --filter @wafflebase/design-editor typecheck` — clean
- `pnpm verify:entropy` — 0 dead-code, 0 doc-staleness
- **Involution asserted over six shapes** (first/middle/last/only child, nested, one-line body), each proving `remove → insert verbatim` is byte-identical
- `renderAttribute`'s expression guard pinned against four injection attempts — the value arrives from a browser and is spliced into a file the dev server executes on the next reload
- Each fix proven by reverting it: the role exemption → exactly 2 tests fail; the null-prototype → exactly 6

`inject.d.mts` deliberately not ported; `inject.types.test.mts` replaces it.

## Summary by CodeRabbit

* **New Features**
  * Added safe design-editor updates for JSX properties, classes, text, child insertion/removal, and imports.
  * Added structured change results and unified diffs for clearer edit previews.
  * Enabled inserting content into supported conditionally rendered or returned elements.

* **Bug Fixes**
  * Improved validation for classes, attributes, imports, and inserted JSX content.
  * Strengthened protection against unsafe edits and prototype-pollution risks.
  * Preserved safeguards for ambiguous, unsupported, or structurally unsafe changes.

* **Tests**
  * Expanded coverage for editing behavior, reversibility, typing, security, and conditional rendering.

Part of #700

### DIFF
--- a/docs/design/design-editor/design-editor-engine.md
+++ b/docs/design/design-editor/design-editor-engine.md
@@ -798,13 +798,32 @@ client's `SceneMeta` can be stale — the same reason `/validate` shares
   numbering, so `return <><A/><B/></>` puts A and B at depth 1 where they are
   genuine siblings in a real container, and splicing between them stays legal.
 
-  This guard is deliberately **wider than the rule it enforces**. Refusing the
-  node also refuses inserting a *child* into it, which is valid — and that is
-  the mainline "insert into the page's root `<div>`" case. Narrowing it needs
-  the OP, which `resolveNode` is not given; `inject.mjs` is what defines the
-  ops, so widening `opts` to carry one belongs with that module rather than
-  ahead of it. Until then the refusal is visible and costs a capability, where
-  the alternative writes an unparseable file.
+**`opts.role` — because two of the three guards depend on what the anchor IS.**
+The guards above were written for one op ("splice this node") and applied to
+all, which refused inserting a *child* into any component whose body is
+`return <div>…</div>`. That is not a missing capability but **data loss**:
+`applyLayoutRemove` inside such a root succeeds, and its `verbatim` inverse is
+refused, so the node is gone with no way back.
+
+| guard | `role: 'target'` (default) | `role: 'container'` |
+|---|---|---|
+| `scope !== 'static'` | refuse | **refuse** — renders N times either way |
+| `owner !== node` | refuse | allow |
+| is a returned expression | refuse | allow |
+
+Verified against the parser rather than argued: a child spliced into a returned
+root parses, a child spliced into a conditionally-rendered element parses, and a
+*sibling* beside a returned root is a syntax error. `applyLayoutInsert` passes
+`'container'` for its parent anchor; everything else keeps the default, so every
+pre-existing caller is unchanged.
+
+One shape stays refused for the same data-loss reason, from the other direction:
+removing a **direct child of a returned fragment**. The fragment is transparent,
+so its children sit at depth 1 with the synthetic `#returns` container above
+them — and no anchor can name that, so the re-insert has no parent to target.
+`applyLayoutRemove` refuses rather than outrunning its own inverse. Making the
+returns root addressable as a container (delegating the offset to the wrapped
+fragment's opening token) would restore it.
 
 `walkJsx` numbering: only JSX elements count; `JsxText`/comments do not.
 Fragments are **transparent** (their children number into the parent's list, so
@@ -915,6 +934,34 @@ in step, in the same spirit as the stamper's cross-consumer test.
 treatment of a name two JSX-returning functions claim, and returns those names so
 the UI can say *"two components here are called `Row`"* instead of rendering a
 subtree whose every node rejects its first edit.
+
+### 5.11 Maps keyed by consumer identifiers carry no prototype
+
+Every map keyed by an identifier read out of a consumer's source is a
+prototype-pollution surface. `roots.__proto__ = tree` hits the inherited setter
+rather than defining a key, so that component **disappears** with nothing
+reported; and `roots.valueOf` answers with an inherited method — truthy, so a
+caller testing `if (roots[name])` walks a function instead of reporting "no such
+component". Both are silent.
+
+The rule is about **who chooses the key**, not which constructor was used.
+`analyzeClasses`'s `antiPatterns` is built with `Object.fromEntries` and *does*
+carry `Object.prototype`; it is exempt because every key comes from `ANTI_KEYS`,
+a closed vocabulary in our own source that no consumer identifier can reach.
+
+**Enforced by a table, not by convention.** `findJsxRoots` was fixed for this,
+and `analyzeNodes` then copied its entries into a plain `{}` and undid the fix
+one layer down — written by the author who had just added the comment explaining
+the hazard. A helper you must remember to call fails exactly where memory
+already failed, so `test/server/name-keyed-maps.test.mjs` enumerates every
+export returning such a map and asserts the invariant against deliberately
+hostile component names. A new export evades it only by omitting its row, which
+is visible in review; the table's own length is asserted so an empty one cannot
+pass vacuously.
+
+This is a different failure from the drift §5.10 guards — that one is two files
+re-deriving one rule, this one is one file re-keying a map — and they need
+different guards.
 
 ---
 
--- a/docs/design/design-editor/design-editor-engine.md
+++ b/docs/design/design-editor/design-editor-engine.md
@@ -903,32 +903,41 @@ directions, and §8.1 check 9 proves the round-trip through real writes.
 
 ### 5.10 The outline PREDICTS the resolver (`src/server/extract.mjs`)
 
-`extract.mjs` builds the node tree the outline renders, and each node carries a
-`structuralEditable` flag. That flag is what enables or greys out "insert
-sibling" and "remove", which makes it a **prediction of what `resolveNode(…,
-{requireStatic: true})` will answer** — a third place the node model is
-interpreted, alongside the injector and the stamper.
+`extract.mjs` builds the node tree the outline renders, and each node carries
+**two** prediction flags. They are what enable or grey out the structural
+controls, which makes them predictions of what `resolveNode(…, {requireStatic:
+true})` will answer — a third place the node model is interpreted, alongside the
+injector and the stamper.
 
-A prediction that disagrees with the resolver is worse than a missing one: the
-UI offers a control, the designer uses it, and the server refuses. So the flag
-mirrors the resolver guard for guard, and the *same* distinction applies —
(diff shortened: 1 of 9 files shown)


hotsunchip:
Thanks — six of the seven reproduced, and I've fixed those. One skipped with reasoning, and one fix deviates from the suggested rule. Details below; each fix is proven by reverting it and confirming *exactly* the expected tests fail.

### Fixed

**Injection on the class and text paths.** Both reproduce, and both write *executable* code, not just corrupted text:

```
addition: 'x" onMouseOver={fetch(`//evil`)} y="'
→ <span className="text-sm x" onMouseOver={fetch(`//evil`)} y="">Hello</span>
   0 parse errors · onMouseOver kind=JsxExpression

text: '{fetch(`//evil`)}'
→ <span className="text-sm">{fetch(`//evil`)}</span>
   0 parse errors · children: JsxExpression
```

This is the module's own stated threat model — `renderAttribute` already refuses unguarded values because "whatever lands in the file is executed by the next HMR reload." The class/text paths bypassed it. The create-a-`className` branch was the clearest case: it interpolated `` ` className="${additions.join(' ')}"` `` by hand instead of calling the renderer that exists to do that safely. It now goes through `renderAttribute`.

**But the two splice contexts get different alphabets**, taken from the parser rather than from taste:

| Context | Measured | Rule |
| --- | --- | --- |
| `className="…"` | `className="a{b}c"` → StringLiteral, `.text` = `a{b}c`, 0 errors. `<`, `>`, `{`, `}` inert. Only `"` escapes. | reject whitespace, `"`, `'`, `` ` ``, `\` |
| JSXText | `{e()}` → JsxExpression, `<F/>` → JsxElement (both executable); bare `>` / `}` → parse error | reject `{`, `}`, `<`, `>` |

So the text guard is exactly as suggested. The class guard is deliberately **not**: rejecting `<`/`>`/braces there would reject `[&>svg]:size-4`, `[&:not(:first-child)]:border-t` and `[&::-webkit-scrollbar]:hidden` — all used in this repo — while leaving the quote, the only character that actually escapes, open. A test pins 12 real Tailwind shapes as *accepted* so a later tightening fails loudly instead of quietly disabling the editor's commonest operation.

Validation sits in `rewriteClassLiteral`, the chokepoint both halves share, so the layout and CVA paths cannot diverge.

**`insertImport`, both findings.** `import {}` + `B` emitted the unparseable `import {, B }`; and a request for a default binding got `already imported` when the file had no default at all — leaving the caller without an import it needs. Restructured so the default and named bindings are decided independently. This also fixes an adjacent case neither of us flagged: appending to `import D, * as M` would have produced the illegal `import D, { A }, * as M`.

**Nitpicks.** `$&` in a replacement expanded through `String.prototype.replace` (`text-sm` → `text-sm-text-sm`); now a function replacer, so `to` lands verbatim. An all-whitespace `raw` reported a successful insert after splicing a bare newline; now refused.

### Skipped: `constructor` in the `HOSTILE` fixture

Measured which hazard each name actually exercises:

```
setter hazard (assignment silently loses the key on a plain {}):
  __proto__ → LOST     toString / valueOf / constructor → fine

lookup hazard (absent name answers truthy on a plain {}):
  __proto__, toString, valueOf, constructor, hasOwnProperty → all truthy
```

`constructor` has no setter hazard, so adding it as a *present* name lands in the category a plain `{}` already handles correctly — the new row would pass even against the bug it's meant to catch. It's currently used in the *absent*-name role (`expect(get().constructor).toBeUndefined()`), which is the category a plain `{}` fails. Taking the suggestion moves it from the check that bites to the one that can't, so I've left it.

16 new tests here (266 total in-package), typecheck and `verify:entropy` clean.

hotsunchip:
@claude review this.

hotsunchip:
All 12 reproduced against `13adc50d7` — no false positives in this set. Fixed in three commits, each proven by reverting it and confirming *exactly* the expected tests fail. 293 tests here (was 250), 333 with #777.

Three of these were on code I added in the previous review round, and #2 in particular invalidated reasoning I had written into the design doc. That is corrected, not just patched.

### Critical

**#1 — remove ate its neighbours.** `childrenOf` numbers JSX *elements*, so text and `{expr}` nodes are not children under that numbering, but the span was taken from the previous element. Removing `<B/>` deleted the prose and the expression too:

```
<A/> / Some prose here / {count} / <B/>   →  removing <B/> left only <A/>
```

The offset now anchors immediately *before* the child at `index`. That alone broke the involution — the append position also had to move from "after the last element" to the end of the child region — which is why the involution table gained three shapes rather than one.

**#2 — `${}` in a template literal.** The one I got wrong. `classLiteralOf` returns `NoSubstitutionTemplateLiteral` as well as `StringLiteral`, and `${alert(1)}` carries no quote, no backtick and no whitespace, so my double-quoted alphabet waved it through: `` className={`p-4 ${alert(1)}`} `` — a live `TemplateExpression`, 0 parse errors. `isSafeClassToken` now takes the delimiter as a **required** argument read from `original[0]`, so the context cannot be forgotten. §5.12 said "two contexts" and was wrong; it says three now, and says why the third was missed.

**#3 — `intent.raw` unvalidated.** The widest hole. `</div>); } evil(); function D(){ return (<div>` closed the JSX, ran a call at module scope, reopened the element — and **the result parses cleanly**, so nothing downstream would have caught it either. Fresh snippets must now parse standalone as exactly one JSX element and carry no denylisted attribute. `verbatim` stays exempt with a test saying why: it restores bytes already in the consumer's file, and validating it would make undo refuse to restore what it just removed.

**#4 — import injection.** `./m'; evil(); import './n` wrote a whole extra statement; a named specifier did the same by closing the brace. Bindings go through `isIdent`, module specifiers are rejected for anything that escapes their quotes — scopes, deep relative paths and extensions still work, asserted.

### Major

**#6** was the subtlest: `{flag ? <A/> : <B/>}` yields two children through one owner, and indices 1 and 2 produced **byte-identical** files. Refused now, naming the group. **#7** two writers on one `className` interleaved into `className="REPLACED"gap-2"`; fixed at source plus a general overlap backstop (reachable — two `sets` on one attribute). **#9** `sets` reached the `className` that `classOps` refuses and clobbered a `t("nav.home")` key. **#8** `import type { Foo, bar }` typechecks and leaves `bar` undefined at runtime. **#10** shape-validation accepted `dangerouslySetInnerHTML={x.y}` — a bare dotted reference is exactly what the expression guard allows; denylisted by name on both the attribute path and inside inserted subtrees. **#11/#12** filled, including the end-to-end insert into a conditionally-rendered parent that `role: 'container'` existed to unblock and which had only resolver-level coverage.

**#5** was a genuine contract break and I fixed it in the direction you implied: `role: 'container'` split one predicate into two, so `buildNode` now emits `containerEditable` beside `structuralEditable`, and the agreement test resolves every node under **both** roles. Reverting it fails 6 of the 8 shape fixtures. Worth noting the drift pointed the *opposite* way from the original bug — the outline was withholding "insert child" on the commonest component shape there is, rather than offering something the server refuses. §5.10 now states both directions.

### One thing I did not claim

The post-splice parse check is **redundant**: removing it fails no test, because snippet validation catches everything first. I kept it and said so in the comment rather than implying coverage — the insertion point is the part that has actually shipped bugs (two, both above), and a bad offset with a valid snippet is what every other guard here waves through. The property it protects is asserted separately over a 20+ case matrix.

dlgpdmsly2 (approved):

PR #777, opened 2026-08-11
Add inject.mjs's token half, completing the AST mutator

hotsunchip:
> **Rebased onto `main` now that #776 has merged.** No longer stacked: this is one commit, **+1,180/−18 across 3 files**, and the diff GitHub shows is its own. Nothing blocks it on ordering any more.

## Summary

Fills in the banner #776 left. Sections A–E: CVA class rewrites, token values, the introspection readers, semantic-token creation/removal, the generic const + emitter primitives, and the Tailwind `@theme inline` aliases.

Addressed entirely by names that exist in the source — `cvaName`, `constName`, a CSS custom property — so **this half never touches the child-numbering model.** That is the seam the split followed: #776 is reviewable against the node-model contract alone, this one against the token-file contract alone.

It reuses `rewriteClassLiteral`, `removeNodeLine`, `indentOf` and `tokenRe` from #776's shared section, so `applyClassRewrite` and `applyLayoutProps` cannot drift into treating the same class edit differently — they are the same user-visible operation reached from two panels.

### The round trip is the property

The token pipeline is closed — source const → emitter array → `@theme inline` alias — so every create has an inverse, and an undo that rolls back past a save replays it. All four pairs are asserted to restore the file **exactly**, not merely equivalently:

| pair | round-trips |
|---|---|
| `insertSemanticToken` / `removeSemanticToken` | ✓ (all three edit points) |
| `insertConstMember` / `removeConstMember` | ✓ |
| `insertBlockEmit` / `removeBlockEmit` | ✓ |
| `insertThemeMapping` / `removeThemeMapping` | ✓ |

### Two decisions worth review attention

**`applyTokenValue`'s `valueKind` is an integrity rule, not formatting.** Writing an expression *unquoted* is what keeps a semantic token bound to the palette; always writing a literal severs that link and strands every other palette consumer on the old colour. Its guard has the same shape as `renderAttribute`'s in #776 and is pinned the same way — only a bare dotted `palette.<key>` reference is accepted, because the value arrives from a browser and lands in a file the dev server executes on the next reload. Four injection attempts are asserted to be refused.

**`removeSemanticToken` deliberately tolerates a PARTIAL token.** One present in `light` but missing from `dark` still gets cleaned up. Demanding all three would leave that half-created state permanently unfixable through this API — pinned by test so it reads as a decision rather than an oversight.

### Conformance table grows

`readSemanticBindings` returns maps keyed by token names read out of consumer source, so both inner maps join the §5.11 table (now 5 rows). The **outer** `bindings` map is deliberately excluded: it is keyed by `light`/`dark`, which are our names, not a consumer's — the same distinction that exempts `antiPatterns`. Verified by reverting the null prototype: exactly the 6 new assertions fail.

## Test plan

- `pnpm --filter @wafflebase/design-editor test` — **286 passing**, 36 new
- `pnpm --filter @wafflebase/design-editor typecheck` — clean
- `pnpm verify:entropy` — 0 dead-code, 0 doc-staleness
- Token-boundary semantics asserted shared with the layout path (`text-sm` must not match inside `text-smaller`)
- `applyTokenValue` asserted to target the const and never the type declaration of the same name
- `insertThemeMapping`'s namespace grouping asserted in both directions — after the last same-family declaration, and the fallback when no sibling shares the namespace

The engine doc's §5.1–5.6 already describe these ops; the port is faithful to them, so no doc change was needed beyond the §5.11 table already in #776.

Part of #700

---

### Since the review round on #776

#776 grew the shared helpers while fixing the 12 panel findings, and two of those changes reach this half:

- **`isIdent` moved up into the shared section** (it was needed for `insertImport`'s binding validation). This half's own copy is gone — a second `const isIdent` in the same module scope is a SyntaxError, which is why the fixup could not stay a separate commit through the rebase.
- **`isSafeClassToken` became delimiter-aware.** `applyClassRewrite` inherits it through `rewriteClassLiteral` with no code of its own, so a backtick `cva()` base is protected against `${…}` substitution exactly as a template-literal `className` is. Asserted here rather than assumed, because "one chokepoint covers both halves" is a claim about this file too.

The three commits this PR carried before the rebase were collapsed into one: two of them declared `isIdent` twice and so did not load on their own, and the repo requires every commit to be green.

## Summary by CodeRabbit

* **New Features**
  * Added safe updates for design tokens and component classes.
  * Added semantic and palette token inspection, creation, removal, and partial-state handling.
  * Added theme mapping and generated CSS value management.
  * Added clear validation and failure reporting for invalid or unsupported updates.

* **Bug Fixes**
  * Improved protection against unsafe names and malformed token data.

* **Tests**
  * Added comprehensive coverage for token updates, theme mappings, round trips, validation, and security safeguards.


hotsunchip:
@claude review.

hotsunchip:
@coderabbitai review

dlgpdmsly2 (approved):

PR #779, opened 2026-08-11
Score complementarity between the review panel and CodeRabbit

dlgpdmsly2:
## Summary

Adds `scripts/agent/eval/complementarity.mjs` — a pure `complementarityOf(records, opts)`
plus a CLI that prints. Over defect classes from `finding-match.mjs`'s `groupFindings`, it
reports per item and per arm: classes both reviewers raised, classes each raised alone, and a
severity-agreement census on the shared ones. It reads a store, computes and **writes
nothing**; it spawns nothing and calls no model. No existing file changes.

## Why

`groupFindings` can collapse a pile of findings into defect classes. Nothing reads them, so
the question the classes exist to answer — *of the defects here, which did both reviewers
raise?* — has no code behind it.

The reason this is a module rather than a script is that the obvious way to compute it is
wrong in a way nothing detects. Two reviewers are not symmetric when one can be re-run and the
other cannot: replay our panel three times, pool the results, and our side gets three chances
at every defect while the other gets one. "Unique to us" inflates, "unique to them" deflates,
and every number that comes out looks ordinary. Per-item volume moves by a third between two
replays of the same input, so this is a sum over a wide distribution presented as a property
of the reviewer.

So the default view compares **one draw against one draw** and refuses records from more than
one run; the pooled views must be asked for by name, carry how many draws each side got, and
print as bounds. The range across replicates is reported and the mean is not.

Three other subsets produce a plausible wrong answer the same way, and each gets a guard that
aborts rather than a log line:

- **Zero findings has two causes.** A clean review is a true negative and a data point; an arm
  we failed to load is missing data, and they are the same zero. So every arm/item pair
  carries a declared `present`/`absent` state — reusing the adapters' existing
  `POPULATION_STATES` — and an undeclared pair is refused rather than assumed.
- **A finding written against a commit the other reviewer never saw** compares two snapshots
  and prints one number. Refused, before anything is counted; the census is printed either
  way, by value and by basis.
- **An item whose replay ended in an error** is not a clean review. Refused, and named.

Severity agreement compares our rubric against a **translation** of the other reviewer's
(`trivial` → `nit`, a section heading → `minor`/`nit`, an unstated severity floored at `nit`).
That is not fixable here, so the census is split by how the severity was obtained and the two
halves are never pooled — which is what that field's own documentation requires. Our own floor
is flagged too: `normalizeSeverity` maps an unrecognised severity to `major`, which blocks.

One thing found while building and **not fixed here**: `groupFindings`' same-run gate reads
`finding.lens`, and `finding-record.mjs` deliberately keeps the lens in the arm namespace.
There is no `lensOf` accessor beside `itemOf`/`armOf`/`runOf`, so a caller passing records
unmodified gets `lens: undefined` on both sides of every pair and the (lens, file) gate
silently becomes a file-only gate. On the data available that is worth 20–28 classes per
scoring run. This module widens each record with the field the gate looks for and **refuses**
the run if it is not there, so a rename upstream breaks the scorer instead of loosening a
gate. The gap itself is written up in the task doc for `finding-match.mjs` to close.

**What this changes today: nothing.** No existing file is touched and nothing imports the new
module. It is a library and a CLI, and its value is that a number now exists that did not.

## Linked Issues

None. It follows #759, which added the defect-class grouping this consumes.

## Author checklist

## Verification

Skip reason (if applicable):

Measured locally on the committed tree:

- `cd scripts/agent && node --test-timeout=60000 --test '**/*.test.mjs'` → **1689 tests, 1689
  pass, 0 fail, 0 skipped**, against a freshly measured `main` (`62b5b986b935`, after #773
  landed) at **1661 / 1661 / 0 / 0** (**+28**). Both trees extracted with `git archive` and given identical `node_modules`, so
  the skip count is 0 on both rather than an environment difference.
- `npx eslint scripts` → **exit 0** on the branch tree and on the base tree.
- **Every new test mutation-tested: 38 mutations, 38 caught, 0 survivors**, source restored
  byte-for-byte.
- Run end to end against real data; the shared classes it reported were read by hand against
  their summaries and files.

## Risk Assessment

- **User-facing risk:** none that I can construct. No existing file changes, nothing imports
  the new module, and it is not on any review or gating path — the panel must never import
  benchmark code and this does not invert that. It writes nothing: no store writes, no file
  writes, no `--out`.
- **Data/security risk:** the CLI makes read-only GitHub API calls through the existing
  CodeRabbit adapter (`gh api` on pull-request comments, reviews and commits) and holds no
  credentials of its own. It spends no money — no model calls, no dispatch.
- **Rollback plan:** revert the commit. Nothing reads the module, so nothing degrades.

## Notes for Reviewers

- **AI assistance:** written with Claude Code. Every number in the task doc and in this body
  was produced by running the code and recorded from its real output; the mutation testing
  and the two-way sensitivity measurement of the lens gate were run rather than reasoned
  about.
- **Where to start:** `VIEWS` and `coverageIndex` in `complementarity.mjs`. Those two are the
  whole design — how many draws each side gets, and how an arm's silence is told apart from
  an arm we could not read.
- **The number this produces is a lower bound, and the module says how loose.** A `maybe`
  never merges, so every undecided cross-arm pair currently counts as two unique catches. On
  the data available that ceiling saturates, and the CLI says so in those words rather than
  printing a bound somebody would quote. Resolving those pairs needs adjudication and is not
  a scorer's job — but the queue is reported strongest-first with the count above a stated
  triage score beside the total, because its size and its cost are different numbers: on the
  data available, 412 undecided pairs have 315 below 0.50 and 17 at or above 0.70.
- **The severity census sits on a population that is not itself reproducible**, and the task
  doc says so with its numbers: across three replays of the same input, 34% of
  blocking-severity anchors recur every time and 40% appear only once — so blocking findings
  churn *more* than average, not less. The gate verdict does not move at all. A
  severity-agreement split over four to six shared classes is a sample, not a property of
  either reviewer, which is why the module reports per replay and never a mean.
- **Follow-up:** `finding-match.mjs` wants a `lensOf` accessor beside `itemOf`/`armOf`/`runOf`
  (see *Why*); and the per-item read in `adapters/panel.mjs` does not return the envelope's
  `status`, so a scorer that must exclude a failed item reads it from the store instead. Both
  are edits to merged files this change had no other reason to touch.

## Summary by CodeRabbit

* **New Features**
  * Added a complementarity scorer for comparing panel and automated findings.
  * Supports per-replicate, union, and intersection views with overlap, unique findings, severity agreement, and unresolved-match reporting.
  * Added filtering, JSON output, repeatable run selection, and detailed per-item and per-arm breakdowns.
* **Bug Fixes**
  * Added safeguards that identify incomplete, invalid, or non-comparable evaluation inputs.
* **Documentation**
  * Added guidance covering scoring criteria, validation rules, pilot measurements, and verification expectations.
* **Tests**
  * Added comprehensive coverage for scoring, validation, grouping, malformed inputs, and command-line behavior.


dlgpdmsly2:
All five valid, all fixed in ac58a16. None changed a published number — each fails silently on a path the pilot corpus doesn't reach, which is what made them worth fixing.

- Queue scoping — links span every item, but the classes they're matched against are the scored set. A link on a non-comparable item resolved to no class and was counted anyway with item: null, giving the queue a different denominator from the overlap it qualifies. Both classes must now resolve first.
- item_id trim — defaultItemOf trims, so " pr-1 " groups under "pr-1". A frame keyed on the raw string matched no class, so every class was filtered out and the run reported 0 of 0 with no error. One helper now normalizes both sides; row validation still runs on the original value, so " " is still refused rather than trimmed into acceptance.
- Duplicate --run-id — worse than a double print: union/intersection were fed that leg's records twice while stats.draws still read 1. Now refused as a usage error rather than deduped, since a repeated flag is a typo.
- Malformed inputs — dropped before grouping, so groupFindings' own stats.skipped never saw them either. Now counted in stats.records.malformed and surfaced in concerns.
- Coverage guards — arm and item_id had no test at all; each is now exercised as a single-condition refusal with the rest of the row valid.

harrykim8672 (approved):

PR #784, opened 2026-08-11
Call the auth smoke test from the pipeline repo

hwisoo-kim:
## Summary

First `workflow_call` port. `agent-sdk-smoke-test.yml` goes from **69 lines to 32** — most of the 32 being a comment about the two non-obvious things below. The workflow body now lives in `wafflebase/agent-pipeline`, pinned at `v0.2.1`.

Chosen first deliberately: it is the smallest workflow in the pipeline, it costs one Haiku call, and it exercises every mechanism the larger ports need — `workflow_call` inputs, secrets across a repo boundary, `environment:` resolving in the caller, the pin assertion, and the trusted checkout. Finding a flaw in that design here costs an afternoon; finding it while porting the 2,084-line panel costs a week.

## What the central copy adds

- An isolated **`pin` job** holding `id-token: write` and nothing else, which reads the OIDC claim to establish which pipeline commit is about to be trusted, asserts the **repository** half as well as the ref (the only way to catch a shim aimed at a *fork* — there is no `job_workflow_repo` context field), and **peels the annotated tag to a commit** so the checkout is immutable.
- A fix to the probe's diagnosis. `auth-smoke.mjs` reported *"auth almost certainly failed"* for `You've hit your session limit` — the opposite of the truth, since a quota refusal proves the credential was accepted. It now classifies quota / auth / unclassifiable behind distinct exit codes, and the workflow treats quota as a pass, loudly annotated.

## Two things measured rather than assumed

Both cost a debugging cycle, and neither is visible in any static check:

**`id-token: write` must be granted here.** A called workflow's job cannot request more permission than the caller granted. Without it the run is a `startup_failure` with **zero jobs** and no usable message.

**`secrets: inherit` cannot forward an *environment* secret.** Declaring `CLAUDE_CODE_OAUTH_TOKEN` a required `workflow_call` secret failed every call at startup — for a repo that had done the setup exactly as documented and put the token in the protected `agent` environment. That environment is the point: it is what lets a required reviewer gate spend. So the token stays there and the called job reads it via its own `environment:`, since environments resolve in the caller.

Both are now adoption-kit requirements and preflight checks rather than folklore.

## Verification

Testbed runs [31478400821](https://github.com/harrykim8672/agent-pipeline-testbed/actions/runs/31478400821) (branch pin) and [31478559653](https://github.com/harrykim8672/agent-pipeline-testbed/actions/runs/31478559653) (annotated-tag pin):

- claim read and matched — `wafflebase/agent-pipeline/.github/workflows/auth-smoke.yml@refs/tags/v0.2.1`
- the peel resolves the tag object to its commit; the checkout uses the commit
- the SDK installs from the pinned commit (103 packages)
- an invalid token is classified **`Authentication FAILED`** — that message (`Not logged in · Please run /login`) was returning *"could not be classified"* until the testbed run surfaced it, and it is now pinned as a regression case

`agent:tests` green: **1669 tests, 1663 pass, 0 fail** (6 pre-existing skips). `auth-smoke.mjs` had no test suite at all before this — one of six such scripts — and now has 9 tests.

## Risk Assessment

Lowest blast radius of any port: this workflow is `workflow_dispatch` only, makes one Haiku call, and nothing depends on it. It is deliberately not gated on `AGENT_PIPELINE_ENABLED` because its purpose is to test *before* arming. No test asserted on the old file's contents.

The one thing that cannot be verified from the testbed is a **green** probe, since that needs a real credential. Dispatching this on wafflebase after merge does that — note the account was at its session limit earlier today, in which case the run will report the quota path, which is now a pass rather than a misleading red.

Rollback is `git revert`.

## Notes for Reviewers

Authored with Claude Code assistance; I have reviewed every hunk.

Worth a look at the `Classify the outcome` step in the pipeline's copy: it decides that a quota refusal is a **pass**, on the grounds that the question this workflow answers is "is my credential good?" and a quota refusal answers yes. If you would rather that stayed red, it is a two-line change and I am happy to make it.

## Summary by CodeRabbit

* **Chores**
  * Updated the manual agent SDK smoke-test workflow with a pre-run authentication check.
  * Improved validation of pipeline access and protected credentials before tests begin.
  * Streamlined smoke-test execution through the standardized pipeline workflow.
  * Removed redundant local setup steps to provide a more consistent and reliable test process.


hwisoo-kim:
Verified both parts against the current repo. **One applied, one not** — and checking the second turned up something worth knowing.

## Applied: SHA pin

Both `uses:` and `pipeline-ref` now carry `4c9ee18de95f873ab14f11e98004457a1decd706 # v0.2.3`, matching the convention every other pipeline checkout here already follows. The pipeline's `pin` job asserts the two against each other, so a half-applied bump fails closed. (`v0.2.3` rather than `v0.2.1` — see below.)

## Not applied: removing `secrets: inherit`

It would remove the credential. The premise is that the token lives in the protected `agent` environment — which is what **my own comment claimed**, so the suggestion was entirely reasonable. Both are wrong:

```
wafflebase/wafflebase repo secrets : … CLAUDE_CODE_OAUTH_TOKEN …
wafflebase/wafflebase agent env    : total_count: 0
```

`CLAUDE_CODE_OAUTH_TOKEN` is a **repository** secret. `secrets: inherit` is what delivers it.

## What I measured while checking

Rather than reason about it, I ran both cases on the testbed. Two facts, and they pull in opposite directions:

- **Environment SECRETS do not reach a called workflow's job.** Token in the caller's environment, job declaring that environment, value still arrived **empty** ([run 31558775121](https://github.com/harrykim8672/agent-pipeline-testbed/actions/runs/31558775121)).
- **Environment PROTECTION RULES do apply.** With a required reviewer on the caller's environment, the called job sat in `waiting` with a pending deployment.

So the property the design actually cares about — a human approving before any run spends budget — **survives extraction**. The environment gates the *job*; it just cannot also be the secret store. The comment in this PR now says that, because the previous one asserted the opposite and would have propagated straight into the adoption guide.

I also added a guard upstream (`v0.2.2`) that fails fast when the credential is empty and names the three ways delivery goes wrong. "No credential arrived" and "the credential is wrong" produce the *same* SDK message (`Not logged in`), so the classifier cannot separate them — but delivery is knowable, and that guard is what caught this.

## One thing for a maintainer, out of scope here

`wafflebase/wafflebase`'s `agent` environment currently has **no protection rules and no secrets** — `protection_rules: (none)`, `deployment_branch_policy: none`. So the "approval before any run spends budget" that `agent-implement.yml:88` and `agent-fix.yml:67` describe is **not in force today**. Nothing in this PR changes that either way, but it seemed worth surfacing given how much of the design leans on it.

Re-verified after grafting onto the rebased branch: `agent:tests` **1668 tests, 1662 pass, 0 fail**; both pins match; YAML parses.

dlgpdmsly2 (dismissed):

harrykim8672:
Conflict resolved — now `MERGEABLE` (`BLOCKED` is just the pending review, not the branch).

One conflict, in `agent-sdk-smoke-test.yml`: this branch replaces it with a shim while `main` kept changing the inline version. Resolved in favour of the shim, but **two things from main needed carrying across that the automatic merge would have dropped**.

## The one worth flagging

Main gained a **`model` dispatch input** wired to `SMOKE_MODEL`, making this the pre-flight for a *model* change and not only a credential check. Git resolved the `on:` block in this branch's favour and **silently discarded the input while keeping the step that read it** — `inputs.model` would have been undefined, and nothing would have failed loudly. Restored, and forwarded to the reusable workflow's own `model` input, which already existed.

That is the second time in this PR chain that an automatic resolution quietly dropped something. Worth remembering when reviewing merges here: a clean auto-merge is not evidence that intent survived.

## The other

`GH_REPO` — nothing in a shim needs it today, since it runs no steps of its own and the pin job addresses `gh` at an explicit `repos/<owner>/<repo>/…` path rather than the placeholder that needs a git remote. **Kept anyway**, because the next step added here would need it and its absence fails obscurely — #790 was exactly that failure.

## Details

- Pin moved to **v0.3.0** so it agrees with every other workflow on main; `uses:` and `pipeline-ref` carry the same commit, which the pin job asserts against the OIDC claim.
- **Merged rather than rebased**: the branch already carried someone else's merge of main, and a rebase would have rewritten their history.
- Final diff against main is **only** `agent-sdk-smoke-test.yml` (+48 / −58) — nothing else rode along.

Verified: `agent:tests` **1693 pass / 0 fail** plus `eval/run.test.mjs` 56/0; YAML parses; all three pins agree.

One pre-existing warning, not this PR's: the drift lane flags `review-panel.mjs`, because #797 landed a review-panel change on main without bumping the pipeline pin. #800 does that bump. Left alone deliberately rather than widened into this PR.

harrykim8672 (approved):

harrykim8672 (dismissed):

hackerwins (approved):

PR #787, opened 2026-08-12
Surface the refused className expression as classNameExpr

hotsunchip:
## Summary

The UX follow-up left open by #777. `classLiteralOf` was narrowed to a joiner
allowlist, so `className={t("nav.home")}` is correctly **refused** by every write
path — but the refusal reached the UI as `className: null`, indistinguishable
from a node with no class attribute at all. The editor therefore showed an empty
class list and offered an edit `applyClassRewrite` then rejected.

`attrsOf` now returns an additive sibling field `classNameExpr` — the expression
**as written** — and `buildNode` carries it onto every node, so the UI can render
a locked, read-only token instead.

The two fields are siblings, **not a union**:

| `className` | `classNameExpr` | the class value is |
|---|---|---|
| string | `null` | a plain literal — fully editable (`"p-2"`, `{"p-2"}`, `` {`p-2`} ``, `{("p-2")}` are one case) |
| string | string | a joiner call with an authored blob — `cn("p-2", x)`. The blob is editable |
| `null` | string | **locked** — an expression with no attributable blob |
| `null` | `null` | no `className` attribute |

So `classNameExpr !== null` means "an expression exists", **not** "locked" — row
2 is editable, and a UI keying off the single field would grey out the commonest
shadcn shape there is.

### The invariant that carried the risk

`fp`/`fpx` are a **stored** wire format: they live in pending edits, transaction
logs, and clients that have not reloaded. Any change to the hashed bytes silently
invalidates every anchor already written down, and surfaces much later as "your
edit could not be located". The new field is read out of the same `attrsOf` call
that feeds `fpOf`, so "additive" needed proving rather than asserting:

- `fp`'s payload and `fpx`'s inputs are untouched.
- A new test freezes **both hashes for 16 fixtures** against values captured by
  running the fixture set against `4ccbd8656`, *before* the field existed.
- That guard is **not** redundant with the §5.10 outline/resolver drift check: a
  payload change moves `walkJsx` and `extract.mjs` together, so they keep
  agreeing while every stored anchor goes stale. Measured — leaking the field
  into `fpOf` fails only the 16-fixture golden test, not the drift check.

`classLiteralOf`'s return type is unchanged (it is the splice target for
`applyClassRewrite`/`applyLayoutProps`), and `className`'s type is unchanged.

### Not covered, deliberately

A **valueless** attribute (`<div className/>`, `<div className={}/>`) has no
expression to show, so both fields are null and it reads as row 4 — while
`applyClassRewrite` still refuses it, since its test is `findJsxAttribute &&
!classLiteralOf`. A UI mirroring the refusal exactly must check
`attrs.includes('className')`; `classNameExpr` supplies the text to *show*, not
the decision. Both shapes are a type error in any real consumer file. This is
asserted as a known shape, documented in `attrsOf`, and flagged in the
`applyLayoutProps` comment that previously promised this PR.

The expression is **verbatim source text**, newlines included, so a single-line
token is the caller's collapse to make.

## Test plan

- `pnpm --filter @wafflebase/design-editor test` — **369 pass** (343 + 26).
- `pnpm --filter @wafflebase/design-editor typecheck` — clean. `classNameExpr` is
  **required** on `SceneNodeMeta`, so the client fixture in `test/types.test.ts`
  failed until updated (TS2741); a UI cannot forget the field.
- `pnpm verify:entropy` — clean (knip, doc staleness, dep freshness).
- `pnpm lint:scripts`, `pnpm verify:architecture` — clean.

Every change was proved by **reverting it** and confirming exactly the expected
tests fail:

| reverted | tests that fail |
|---|---|
| `attrsOf` drops the field | 10 — the 8 new pair tests, the whole-object `attrsOf` assertion, the extract test |
| `buildNode` stops surfacing it | 1 — the extract test only |
| field leaks into `fpOf` | **6** — exactly the fixtures with a non-null expression; nothing else in the suite |
| field leaks into `fpxOf` | the same 6 |
| first-`className` guard removed | 1 — the duplicate-attribute test |
| braced-literal exemption removed | 1 — the braced-literal test |
| paren-unwrap removed from the exemption | 1 — the braced-literal test |

One assertion was caught **vacuous** by this method: the extract test originally
used `t("nav.home")`, whose text contains no tokens, so it passed whether or not
the expression reached `analyzeClasses`. The fixture is now a ternary
(`isActive ? "bg-primary" : "bg-muted"`), whose text yields
`['muted', 'primary']` if leaked — measured, and the test now fails under that
mutation.

Docs: `design-editor-engine.md` §5.7 gains the contract table, the two documented
limits, and the anchor-stability argument; §5.10 gains a cross-reference.

## Summary by CodeRabbit

* **New Features**
  * JSX class names now include their original expressions.
  * Editable class literals remain editable, while complex or unsupported expressions are read-only.
  * The outline displays resolved class names and their source expressions.

* **Bug Fixes**
  * Class-name expressions no longer affect node identity or fingerprint values.

* **Documentation**
  * Clarified class-name editing and outline display behavior.

* **Tests**
  * Added coverage for literals, conditionals, duplicates, missing values, spreads, nesting, and multiline attributes.

### DIFF
--- a/docs/design/design-editor/design-editor-engine.md
+++ b/docs/design/design-editor/design-editor-engine.md
@@ -770,6 +770,41 @@ exactly the moment it is needed, and resolution falls through to `fp` when it is
 not. No new failure mode, and most post-external-edit relocations become silent
 recoveries instead of "discard your edit".
 
+**`className` reaches the UI as TWO fields, and neither one is in `fp`.**
+`classLiteralOf` names a rewrite TARGET, so it refuses anything it cannot
+attribute to an authored blob — and that refusal was invisible:
+`className={t("nav.home")}` arrived as `className: null`, which the editor could
+not tell from a node with no class attribute, so it offered an edit
+`applyClassRewrite` then refused. `attrsOf` therefore also returns
+`classNameExpr`, the expression **as written**, for the UI to render read-only:
+
+| `className` | `classNameExpr` | the class value is |
+|---|---|---|
+| string | `null` | a plain literal — fully editable (`"p-2"`, `{"p-2"}`, `` {`p-2`} ``, `{("p-2")}` are one case) |
+| string | string | a joiner call with an authored blob — `cn("p-2", x)`. The blob is editable, the rest is the author's |
+| `null` | string | **locked** — an expression with no attributable blob (`t("nav.home")`, `styles.row`, a ternary) |
+| `null` | `null` | no `className` attribute |
+
+`classNameExpr !== null` means "an expression exists", **not** "locked" — row 2
+is editable, and a UI keying off the single field greys out the commonest shadcn
+shape there is. Locked is `className === null && classNameExpr !== null`.
+
+Two limits, both deliberate. A **valueless** attribute (`<div className/>`,
+`<div className={}/>`) has no expression to show, so it reads as row 4 while
+`applyClassRewrite` still refuses it — its test is `findJsxAttribute &&
+!classLiteralOf`, so a UI mirroring the refusal exactly must check
+`attrs.includes('className')`; `classNameExpr` supplies the text to *show*, not
+the decision. And the text is **verbatim source**, newlines included, so a
+single-line token is the caller's collapse to make.
+
+The field is additive in the strict sense: it enters neither `fp` nor `fpx`, and
+`test/server/jsx-nodes.test.mjs` pins both hashes for sixteen fixtures against
+values recorded *before* it existed. That guard is not redundant with the drift
+check in §5.10 — a payload change moves `walkJsx` and `extract.mjs` together, so
+they keep agreeing while every anchor already written down goes stale. Measured:
+routing the new field into `fpOf` fails exactly the six fixtures that have an
+expression, and nothing else in the suite.
+
 `resolveNode` tries **path → unique `fpx` → unique `fp` → refuse**, and *every
 search step resolves only on exactly one match*. Ambiguity is treated as absence:
 picking the first of two identical spans would write to the wrong node silently,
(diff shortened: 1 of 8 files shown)


hotsunchip:
> **Cover parenthesized non-literal expressions.** Line 860 verifies a parenthesized direct literal only. `unwrapParens` also changes joiner and refused-expression classification.

Valid, and applied in d42e88339 — thanks. Verified by probe before writing the test, so recording what it is and isn't:

**Not a bug.** The behaviour is already correct, and your predicted values are exact:

```
{(cn("p-2", x))}     → ['p-2', '(cn("p-2", x))']
{(t("nav.home"))}    → [null,  '(t("nav.home"))']
{((cn("p-2", x)))}   → ['p-2', '((cn("p-2", x)))']   # the unwrap recurses
```

**But not redundant either**, which is the part worth stating. The gap is narrower than "parenthesized expressions" and also realer:

- `{("p-2")}` (my braced-literal test) exercises `unwrapParens` on the path to the **literal** check.
- `cn(("a b"))` (pre-existing `classLiteralOf` test) exercises it on an **argument**.
- Parens around the **joiner call itself** was covered by neither — and it is the only shape that reaches `unwrapParens` on the path to the **call** check.

So it guards a specific refactor slip: unwrap for `asLiteral`, then test `isCallExpression` on the raw expression. Mutating `classLiteralOf` that way fails the new test and **nothing else in the 370-test suite** — measured, not assumed.

Third case added beyond your two: `classNameExpr` keeps the parens (`'(cn("p-2", x))'`, not `'cn("p-2", x)'`), because it is verbatim source text rather than a normal form. Pinning that stops a later "tidy the display text" change from silently becoming a normalizer.

dlgpdmsly2 (approved):

PR #788, opened 2026-08-12
Re-plan PR 8: split the plugin host from the token seam

hotsunchip:
## Summary

Docs-only. Scoping PR 8 (the §6 generalization) required re-verifying the
coupling table against `feat/design-system`, and **the table was wrong in two
ways** — so this lands the corrections and the resulting PR split before any code
is written.

### Two rows named one construct and pointed at another

| row | said | actually |
|---|---|---|
| ``FRONTEND_SRC` + the `@` alias`` @ `350` | one site | **two.** `FRONTEND_SRC` is at `350` (used only by `sceneSpecifier` at `364`); the `@` alias is at **`2441`**, resolved from `__dirname` rather than `REPO_ROOT` |
| ``packages/{sheets,docs,slides,notes}/src` aliases` @ `624-627` | aliases | `624-627` is **`ENGINE_SRC_ROOTS`**, a frame-propagation *performance* boundary. The aliases are at **`2483-2486`** |

`ENGINE_SRC_ROOTS` is not a move at all: it marks trees resolved and served but
never re-queried per frame side, because doing so doubles every canvas scene mount
for byte-identical content. Every consumer with a large non-JSX subtree wants
that, so the *option* is generic (`opaqueRoots`) and only its value is ours.

### Five couplings were missing outright

`antlr4tsAssertShim()` / `yorkieOffline()` + `YORKIE_OFFLINE_SHIM` (`51,75,119`) ·
app libs aliased into `packages/frontend/node_modules` (`2527-2535`) ·
`optimizeDeps.include` (`2387-2436`) · the `define` block —
`__APP_VERSION__`/`rootVersion()`, `process.env`, `VITE_BACKEND_API_URL` →
`SCENE_API_ORIGIN` (`2330-2341,2378-2386`) · and that the `plugins` array supplies
`react()` and `tailwindcss()` **itself** (`2356-2365`), which the plugin must stop
doing for the same reason it must stop owning the Vite root.

Both kinds of error are exactly what §2 says a boundary resting on inspection will
suffer — and this table *is* the inspection. The table now carries a `PR` column
so it drives the rollout instead of being re-read each time.

### The load-bearing finding: the seam is not on the file boundary

The obvious cut is one PR per file — host is `vite.config.ts`, adapter is
`edits.ts`. It fails: **seven token rows are inside `vite.config.ts`**, threaded
through the machinery the host PR must port anyway.

| site | what the token pipeline does there |
|---|---|
| `mutationBridge` `1798-1803`, `1899` | the observe/watch list *is* the four token files + theme CSS + `TOKENS_CSS` |
| `mutationBridge` `1909-1915`, `2000-2015` | two endpoints read all four; `/preview-tokens` diffs `renderTokenVars(patched)` vs base |
| `filesForIntent` `1354` | a token intent's file set is `FAMILY[…].file + BUILD_CSS_FILE + THEME_CSS_FILE` |
| `applyIntentToCache` `1465-1467` | the three-point write |
| `restoreTransaction` `1564-1565`, commit `2084`, `2263` | `isTokenSourcePath` gates the CSS regen |

`edits.ts:116-119` is only the **client's duplicate** of the four paths. A
file-shaped 8a therefore carries the entire token pipeline, still weighs ~2,500
lines, and its halves fail the *same* way rather than differently — which is none
of what splitting is for.

So PR 8 splits **by pipeline**, the cut that already worked for the module
underneath it (`inject.mjs` shipped layout-half #776 then token-half #777):

- **8a — plugin host, layout only.** `options.root` threaded through all **50**
  `REPO_ROOT` sites, shell served from prebuilt assets via `configureServer`,
  scene registry, `?wbFrame=` + stamping, safelist as a virtual `@source`,
  `resolveSafe` + backups, transactions, layout endpoints. *Declares*
  `TokenAdapter`, ships no implementation; token intents answer "no token adapter
  configured" — which §3 already promises as the steady state for any project
  outside the support matrix.
- **8b — the token seam.** `cssVariables` default impl, token endpoints routed
  through the adapter, `edits.ts`'s four paths → server-supplied metadata.
- **8c — `packages/design-sandbox`.** Finally gives population C a destination.

### Two things the doc assumed were already done

- **`cssVariables` is net-new code.** §4 presented two adapters as if both
  existed; only `wafflebaseCore` does. Nothing on the prototype reads or writes a
  single-stylesheet token layer — so the *default* adapter, the one every
  non-wafflebase consumer gets, is the one piece of the token half with no
  reference behaviour to diff against.
- **`packages/design-sandbox` does not exist** — not on `main`, not on
  `feat/design-system`. Every "moves to design-sandbox" is a destination 8c must
  create, and until it lands, population C has nowhere to go and wafflebase's own
  scenes stop rendering in the interval. Now stated as a scheduled cost.

### Verification for 8a, stated as a cost

The prototype verifies the bridge with **scripts against a live dev server**, not
unit tests: `verify-bridge.mjs` (18.9 KB), `smoke-scene.ts` (13.1),
`smoke-canvas.ts` (12.1), `smoke-layout.ts` (6.8). The engine modules' measured
3.2–4.8× test multiplier does not transfer to plugin code. They port across
unchanged in kind and stay out of `verify:fast`/`verify:self` — so **8a lands with
no automated gate**, and a regression in it is caught by hand or not at all. That
is written down rather than omitted, and it is the argument for a fixture-project
integration lane later, which would prove the pivot's central claim (the plugin
works in a *foreign* project).

Also folds in the deferred §8 staleness: the ledger said #718/#738 were open and
#758 was "next" when all of PRs 1–7 are merged.

## Test plan

Docs-only — no source, no tests, no behaviour change.

- `pnpm verify:entropy` — clean, **including the doc-staleness check**.
- Every line number in the new table verified against
  `git show feat/design-system:packages/design-sdk/vite.config.ts`, not carried
  over from the previous revision. The `REPO_ROOT` count (50) and the token-site
  list are `grep -n` output.
- Markdown table arity checked column-by-column; the one 5-field row is the
  pre-existing escaped `\|` inside `WATCHED_RE`, unchanged.

Independent of #787 (that one edits `design-editor-engine.md`), so the two review
in parallel and merge in either order.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the design editor rollout plan and sequencing.
  * Documented the upcoming design sandbox and temporary handling of Wafflebase scenes.
  * Expanded the coupling inventory with configuration dependencies, ownership, and behavioral considerations.
  * Added details on shell serving, Tailwind safelisting, opaque roots, token handling, and verification status.

### DIFF
--- a/docs/design/design-editor/design-editor-local-plugin.md
+++ b/docs/design/design-editor/design-editor-local-plugin.md
@@ -108,6 +108,14 @@ Every file on `feat/design-system` falls into exactly one of three populations.
 `src/scenes/fixtures/**`, `data/mock-metadata.ts`, `scenes.config.json`,
 `src/scenes/providers.tsx`.
 
+> **`packages/design-sandbox` does not exist yet** — not on `main`, and not on
+> `feat/design-system` either. Every "moves to `design-sandbox`" in this document
+> is a destination that PR **8c** creates, so until it lands, population C has
+> nowhere to go: the plugin PRs *delete* those couplings and wafflebase's own
+> scenes stop rendering in the interval. That is the cost of the boundary being
+> structural, and it is the reason 8c is scheduled rather than assumed — the
+> dogfood is also the only proof the split holds.
+
 **Target layout**
 
 ```text
@@ -183,11 +191,16 @@ interface TokenAdapter {
 }
 ```
 
-Two implementations:
+Two implementations — **one of which does not exist yet:**
 
 - **`cssVariables` (default, ships in the package)** — one stylesheet, `:root` /
   `.dark` blocks, `@theme inline`. `plan()` returns a single write. Covers the
-  shadcn population.
+  shadcn population. **This is net-new code.** Nothing on `feat/design-system`
+  reads or writes a single-stylesheet token layer; the prototype has only the
+  four-file pipeline below. So the default adapter — the one every consumer
+  outside wafflebase gets — is the only part of the token half with no reference
+  implementation to diff against, and the only part whose first version cannot be
+  verified by comparing it to the prototype's behaviour.
 - **`wafflebaseCore` (lives in `design-sandbox`)** — the existing four-file
   pipeline, the `build-css.ts` worker, and the three-point `token-add`. It stays
   exactly as built; it just stops being the assumption.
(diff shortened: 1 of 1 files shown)


hotsunchip:
@coderabbitai review this.

dlgpdmsly2 (approved):

hackerwins:
@coderabbitai review this.

PR #596, opened 2026-07-30
Post an agent-effort summary after an on-demand @claude review

hwisoo-kim:
## Summary

`@claude review` (the on-demand review panel) ran the lens panel and posted findings, but — unlike the autonomous panel — it never recorded the review's cost or posted the `🤖 Agent effort` summary. A maintainer saw the findings with no idea what the review cost. This wires up the missing metrics.

The on-demand `review` job already had everything it needed: the panel writes `.agent-review/review-execution.json` + `review-lens-stats.json`. This adds the two steps that consume them.

## Changes

- **`agent-review-on-demand.yml`** — after the findings comment, two new steps:
  - **record** the review into the PR metric ledger (`metrics.mjs record --kind review`)
  - **summarize** (`metrics.mjs summarize --pr N`) → posts the `🤖 Agent effort` comment
  
  Both run the **trusted `.trusted/` copy** of `metrics.mjs` (never the branch's), gate on `steps.panel.outcome == 'success'`, use the **App token** (a fork-PR `GITHUB_TOKEN` is read-only), and are `continue-on-error` on top of `metrics.mjs` already exiting 0 on any error — metrics can never fail a review. Deliberately **not `--final`**: that sweeps every hidden metric record on the PR and would delete the autonomous pipeline's ledger mid-flight on an agent-managed PR.
- **App token** gains `permission-pull-requests: read` — `summarize` calls `gh pr view --json additions,deletions` to size scope.
- **`metrics.mjs renderSummary`** — omits the "Code-fix agent" section when there are no code-fix sessions (`aggregate([])`), so a review-only summary no longer renders a block of $0.00/0 and the total collapses to the review figure. The autonomous path always has a code-fix record, so it's unchanged.

## Test plan

- New `renderSummary: review-only (no code-fix)` test; full agent suite `cd scripts/agent && node --test *.test.mjs` → **344/344 pass**.
- Workflow YAML validated.

Note: this is a pipeline-internal change and ships no product code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * On-demand review results now record execution and lens metrics automatically.
  * Successful reviews now include an aggregated agent-effort summary.

* **Bug Fixes**
  * Review-only summaries no longer display empty code-fix sections or misleading cost breakdowns.
  * Review-only totals and token counts now accurately reflect review activity.

* **Tests**
  * Added coverage validating review-only summary formatting and totals.

PR #790, opened 2026-08-12
Set GH_REPO so gh stops needing a git remote

hwisoo-kim:
## Summary

**Fixes a live outage, and I caused it.** The autonomous loop has been dead repo-wide since #769 merged (08:48 UTC, 2026-08-11). #786 is the first agent PR after it and the first casualty.

One line per workflow: `GH_REPO: ${{ github.repository }}` at workflow `env:` level.

## What broke

`gh` expands `{owner}/{repo}` by shelling out to git, so it needs a git remote in the working directory.

Before #769, the fix job's first step was `actions/checkout` of **this** repo at `ref: main`. That gave the job two things: the trusted scripts, and a `.git` at the workspace root. #769 replaced it with a checkout of the pipeline repo into a scratch path that the adapter deletes after moving the scripts out:

```yaml
- uses: actions/checkout@v4          # repository: wafflebase/agent-pipeline
  with: { path: .pipeline-src }
- run: |
    mv .pipeline-src/packages/pipeline scripts/agent
    rm -rf .pipeline-src              # ← the only .git in the workspace
```

The scripts still arrive. The git context no longer does. Every step running before the PR-branch checkout fails one second in:

```
unable to expand placeholder in path: failed to run git:
fatal: not a git repository (or any of the parent directories): .git
```

Confirmed on run [31501221344](https://github.com/wafflebase/wafflebase/actions/runs/31501221344) — `fix` red, error verbatim.

## Wider than the fix job

`mark-ready.mjs` uses the same placeholder, so **`promote` would fail identically** the first time a PR was clean enough to reach it. It has been `skipped` since the merge, which is why only the fixer had surfaced.

The `stalled` job's mislabelling has the same root: it pages via `github-script` (works) but calls `set-state.mjs`, which is fail-safe and exits 0 — so #786 still reads `agent:reviewing` while paged.

`agent-review-on-demand.yml` is the one workflow that kept working, for two reasons: its review job checks out the PR at `head_sha`, and it has carried `GH_REPO` since #596 for exactly this reason. **The fix is this repo's own existing idiom, not a new one.**

## Why workflow level, and all ten

Five were broken; all ten now carry it. A new step that shells out to `gh` is exactly how this comes back, and only a workflow-level default covers steps nobody has written yet. Pre-existing `env:` keys (`MAX_ATTEMPTS`, `MAX_REVIEW_ROUNDS`) are preserved — asserted, not eyeballed.

## Why the tests did not catch it

Worth stating plainly, because it is the more useful lesson. The scripts were never wrong — the missing thing was the **environment they run in**, which no unit test observes. And the smoke test I verified #769 with does not exercise it either: `command.mjs` and `auth-smoke.mjs` never call `gh`. So my verification proved the scripts were present and executable, which was true, and said nothing about whether they could talk to the API.

`checks.test.mjs` now asserts the invariant. Mutation-tested: removing `GH_REPO` from any one workflow fails the guard.

## Verification

- `agent:tests` green: **1742 tests, 1736 pass, 0 fail** (6 pre-existing skips).
- All ten workflows parse; `GH_REPO` present at workflow level in all ten.
- Pre-existing `env:` keys preserved.

I cannot dispatch the affected workflows to prove the runtime fix (they are `workflow_run`/`issue_comment` triggered), so the real confirmation is the next agent PR — or `@claude rerun` on #786 once this lands.

## Risk Assessment

Additive: one environment variable, no logic touched. `GH_REPO` only affects how `gh` resolves a repo it was already targeting. Rollback is `git revert`.

Given the loop is currently dead, this is worth landing ahead of the other open pipeline PRs.

## Notes for Reviewers

Authored with Claude Code assistance; I have reviewed every hunk. Credit to whoever ran the diagnosis on #786 — the analysis was exactly right, including the `agent-review-on-demand` observation that pointed at the idiom. The one thing I would add is the `promote` blast radius above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub Actions reliability when running commands before repository checkout or without local Git metadata.
  * Ensured automated workflows consistently identify the current repository.

* **Tests**
  * Added regression coverage to verify applicable automation workflows define the required repository context.

### DIFF
--- a/.github/workflows/agent-fix.yml
+++ b/.github/workflows/agent-fix.yml
@@ -30,6 +30,16 @@ on:
 
 permissions: {}
 
+env:
+  # gh expands `{owner}/{repo}` by shelling out to git, so it needs a git remote
+  # in the working directory. Since the pipeline moved to its own repo, the
+  # trusted checkout lands in a scratch path that is then deleted — which removed
+  # the only .git in the workspace, and every step that runs before the PR-branch
+  # checkout began failing with "not a git repository". GH_REPO makes gh skip the
+  # git lookup entirely. agent-review-on-demand.yml has carried it since #596 for
+  # this exact reason; it was the one workflow the regression did not reach.
+  GH_REPO: ${{ github.repository }}
+
 jobs:
   route:
     if: >-
--- a/.github/workflows/agent-implement.yml
+++ b/.github/workflows/agent-implement.yml
@@ -30,6 +30,16 @@ permissions:
   pull-requests: write # open the draft PR + comment
   issues: write # comment back on the issue
 
+env:
+  # gh expands `{owner}/{repo}` by shelling out to git, so it needs a git remote
+  # in the working directory. Since the pipeline moved to its own repo, the
+  # trusted checkout lands in a scratch path that is then deleted — which removed
+  # the only .git in the workspace, and every step that runs before the PR-branch
+  # checkout began failing with "not a git repository". GH_REPO makes gh skip the
+  # git lookup entirely. agent-review-on-demand.yml has carried it since #596 for
+  # this exact reason; it was the one workflow the regression did not reach.
+  GH_REPO: ${{ github.repository }}
+
 jobs:
   # Parse the "@claude <verb>" command from the comment via the shared
   # scripts/agent/command.mjs (flexible/containment matching: case-insensitive,
--- a/.github/workflows/agent-iterate-ci.yml
+++ b/.github/workflows/agent-iterate-ci.yml
@@ -29,6 +29,14 @@ permissions:
   checks: read # loop-status.mjs reads per-commit check runs for the round table
 
 env:
+  # gh expands `{owner}/{repo}` by shelling out to git, so it needs a git remote
+  # in the working directory. Since the pipeline moved to its own repo, the
+  # trusted checkout lands in a scratch path that is then deleted — which removed
+  # the only .git in the workspace, and every step that runs before the PR-branch
+  # checkout began failing with "not a git repository". GH_REPO makes gh skip the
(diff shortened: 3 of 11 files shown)


harrykim8672:
CI was red on the first attempt. **It was the known `agent:tests` deserialize flake, not this change** — green on re-run, no code touched.

Evidence:

- The failure was `scripts/agent/eval/run.test.mjs` dying with `Unable to deserialize cloned data due to invalid or unsupported version` inside `node:internal/test_runner/runner #processRawBuffer`. This branch changes ten workflow `env:` blocks and one test file; it touches nothing under `eval/`.
- **`main` fails the same way.** Run [31554461954](https://github.com/wafflebase/wafflebase/actions/runs/31554461954) on `main` produced the identical `run.test.mjs` / `Unable to deserialize cloned data` pair, and `main` has been intermittently red today (02:49 and 01:41 failures against green runs either side).
- This branch already contains #774's mitigation (running `eval/run.test.mjs` in its own `node --test` invocation), so the remaining flake is what #774 reduced without eliminating.
- Attempt 2: `verify-self (22.x)` **success**, no changes pushed.

I deliberately did **not** attach a flake fix to this PR. #781 reverted #772 for exactly the right reason — *"a fix carrying a cause nobody has demonstrated is worse than no fix"* — and #772's `NODE_CHANNEL_FD` hypothesis was measurably wrong. Bolting a speculative fix onto an urgent outage PR would repeat that mistake and expand a change whose whole value is being one additive line per workflow.

Happy to take the flake on separately, starting from evidence rather than a hypothesis. The `NODE_TEST_CONTEXT`-reaching-spawned-children thread that #781's message leaves open looks like the place to start.

harrykim8672 (approved):

PR #791, opened 2026-08-12
Score cost and wall-clock latency over stored replay envelopes

dlgpdmsly2:
## Summary

Adds `scripts/agent/eval/cost-latency.mjs`: a pure `costLatencyOf(runs, opts)` plus a CLI that prints. Per item and per replicate it reports cost, wall clock, turns and calls; across replicates it reports a range rather than a mean; and it segments by diff-size bucket. It reads a store, computes and writes nothing — no model call, no network call, no spend.

## Why

Every replayed item stores what it cost and how long it took, and nothing reads any of it.

The reason the shape matters more than the arithmetic: **each envelope carries two time fields and only one of them is elapsed time.** `duration_ms` is the panel's own wall clock from `review-timing.json`. Beside it sits the flat sum over every SDK call, which is larger because the panel runs its lenses — and each lens's samples and verifier calls — concurrently. That field is not a defect; its docblock in `run.mjs` says outright that it is not a duration. The hazard is that it is adjacent, plausible and bigger. **Over the 21 stored replays it exceeds true wall clock by 1.36×–3.27× per item and 2.40× at run level.** So one function reads a duration here, it reads the provenance beside the number, it refuses when the two disagree, and the summed field is not named in the module at all — a test greps for it.

Three numbers are absent rather than approximated, and each absence is the point:

- **No average cost per line.** Cost is dominated by a per-item floor — a fitted `$2.20 per item + $5.20 per 1000 lines`, so **47% of a run is fixed overhead**, and the 1004-line item costs 3.1× the 22-line one for 46× the lines. The average therefore inverts: the biggest item is the cheapest per line. A marginal slope is reported instead, because it cannot be read as the cost of a typical line.
- **No cost per real finding.** It needs confirmed-real findings and none are adjudicated. Cost ÷ all findings is the worst substitute available precisely because it looks like the real metric — a reviewer that raised twice as many false findings would score twice as cheap.
- **No ratio against the other reviewer.** A flat subscription has no per-review price to divide by, so that block carries a different key name, a different unit and its basis; an amortised per-PR figure appears only when a caller supplies both a list price and the volume it is spread over.

Two money questions are kept apart and neither is called "cost": what a run spent, failures included, and what one successful review costs. A failed item is not a cheap review, and one stored replay spent real money before it failed.

`duration_source` turns out to have three values rather than two — the runner also writes `not-run` for an item where no panel ran, which is a different absence from a missing timing file. The census keeps all three apart, prints every one including the zeros, and gives an unrecognised value its own bucket and a reason.

The reviewer-identity and corpus-label guards are `reliability.mjs`'s rather than a second pair here. Two exported guards sharing a name in one directory while drifting apart is a failure this directory has documented before, and the borrowed pair is the stricter — it refuses an input where no run states an identity. It diverges from that scorer in one place, deliberately: reliability *aborts* on a failed item because a truncated finding set reads as disagreement, while cost *reports* it, because a failed replay still spent money and refusing the run would make that spend invisible.

## Linked Issues

None. It follows the eval store, the replay runner and the existing scorers in `scripts/agent/eval/`; it closes no issue.

## Author checklist

## Verification

Skip reason (if applicable): none.

Measured locally from the committed tree, both trees extracted with `git archive` and given identical `node_modules`:

- `agent:tests` — **1770 (1715 + 55), 0 fail, 0 skipped**, against a freshly measured `main` (`c615b8c`) at **1741 (1686 + 55)**. **+29.** Two invocations, matching the lane.
- `eslint scripts` — exit 0 at the lockfile-pinned 9.24.0.
- **36 mutations, 36 caught, 0 survivors.** One survivor of the first pass was a real defect and is fixed in the code: a `variance === 0` guard against "every item is the same size" never fires, because `100/1000` is not exact in binary and three copies of it give a variance of `2e-33`. It now counts distinct integer line counts.
- Run over three stored replicates it reproduces every per-item cost and wall figure and the `$2.20 + $5.20/1000 lines · 47% fixed` fit to the cent.

## Risk Assessment

- **User-facing risk:** none. Three new files; nothing imports them, no workflow invokes them, and nothing in the gating path is touched. `review-panel.mjs`, `severity.mjs`, the gate and every workflow are untouched, and the whole existing suite passes unmodified.
- **Data/security risk:** none. The module writes no file and makes no network call; there is no default `--root`, so it cannot fall back to a path inside this repository. It spends nothing — no model is invoked on any path.
- **Rollback plan:** revert. Nothing reads it.

## Notes for Reviewers

- **AI assistance:** written with Claude Code — the module, its tests and the task doc. The measured figures come from running the code against stored data, not from the model.
- **Follow-up work:**
  - `duration_source`'s three values are inline literals in `run.mjs` with no exported constant, so a consumer cannot pin against the vocabulary; this module's unrecognised-value bucket is a local workaround, not a fix.
  - The other arm's latency is emitted as an explicit null whose reason says **measurable, and not from anything this scorer reads** — its status comment carries a "review in progress" marker and is edited in place, so its own revision timestamps bracket the work. That is a GraphQL read of one comment's edit history and belongs to the arm's adapter, not to a scorer that otherwise touches no network. A follow-up should wire it up; measured by hand over the same pull requests it lands at 2.5–14.4 min.
  - `scripts/agent/eval/README.md`'s module table is two scorers stale — worth one PR that catches it up rather than an edit here.

## Summary by CodeRabbit

* **New Features**
  * Added cost and latency reporting for stored evaluation runs.
  * Reports spending, timing, failed-item costs, replicate ranges, and size-based trends.
  * Supports text and optional JSON output, with validation for incomplete or inconsistent data.
  * Includes optional subscription-cost comparisons and clearly identifies unavailable metrics.

* **Documentation**
  * Added design guidance, pilot results, safety considerations, and known limitations.

* **Tests**
  * Added comprehensive coverage for timing, spending, validation, aggregation, reporting, and comparisons.


dlgpdmsly2:
All six addressed in `de3287c1`.

**`cost-latency.mjs:789` — unsized items only logged.** The substantive one: the result could read `COMPLETE` and exit 0 while a size bucket said `(unknown)`, so every per-size figure was quietly about less than the corpus. The CLI now records an unreadable item as unsized instead of omitting it from the map, and the scorer raises a named completeness reason — making the verdict `partial` and the exit non-zero. Unknown-bucket reporting is unchanged. One deviation from the suggestion: the reason is derived from the scored rows rather than from ids threaded into `costLatencyOf` — same outcome with no new parameter, and it also catches a caller that built its size map wrongly, not just one that could not read an item.

**`cost-latency.mjs:725` — size-bucket line.** `b.diff_lines` now goes through `rangeStr` like every other series, so an empty series prints `n/a (n=0)` rather than `null–null`. Normal ranges are unchanged apart from gaining the median every other range already prints. New test: an item with no frozen metadata, asserting `diff_lines.n === 0`, that its dollars are still counted, and that the rendered line contains no `null`.

**`cost-latency.mjs:648` — `replicate_spend_usd` comment.** It said ~6%, which is half-range over the mean; the field is `spread_over_min` and prints ~11%. Both are real with different denominators, so the comment now states ~11% and names which. Code unchanged.

**`cost-latency.mjs:611` — precision.** Both dollar figures now print at four decimals; the test asserts the formatted stored value.

**`cost-latency.test.mjs:281` — lower-bound discordant count.** Now exactly 3 discordant, 18 concordant. A lower bound passes just as happily when a change makes latency look *less* ordered than it is, which is the direction that number gets misread in. Also added a value-tie case — `tied` is reachable from either side (equal sizes, or equal values) and only the first was covered.

**Task doc fence** — `text` tag added.

Verification on the committed tree: **1771 tests (1716 + 55), 0 fail, 0 skipped**, against `main` at **1741 (1686 + 55)**; `eslint scripts` exit 0 at the pinned 9.24.0; **40 mutations, 40 caught**, including one restoring each fix above. Output over the three stored replicates is unchanged apart from the bucket line.

dlgpdmsly2:
Pushed `0a4c16d2`, correcting the `coderabbit_latency_ms` declared gap.

The gap said the metric was measurable and then proposed the wrong way to measure it — the status comment's `created_at` → `updated_at` pair, described as "a duration it timed itself". That pair reads **53.7 min against a true 6.6** on one pilot item, because the last edit is the last edit of *anything*: the comment is re-edited long after the review lands. The same rule is right to within four seconds on the per-invocation acknowledgement, and a rule that is right on one comment kind and 8× wrong on another is not a rule.

The gap is rewritten rather than deleted, and now names the interval — `coderabbit-start-marker-to-first-finding`, `t0` being the latest CodeRabbit start marker before the finding, `t1` the earliest CodeRabbit artefact on the frozen commit — along with all three rejected starts and their measured errors. A deleted wrong answer gets reinvented; a named one does not.

The half that matters more: **a number will not make the two arms comparable, and the bias runs the opposite way from the long-standing assumption.** Ours is a panel process's elapsed time on an offline replay that queued for nothing; theirs is a production reviewer measured end to end. Where both were measured from a single trigger in production, ours ran about **2.2× longer**. So the two stay in separate blocks with separate units and no ratio — minutes need that discipline more than dollars, because they look commensurable.

The latency block itself is deliberately still not here. It needs the arm's adapter to supply the figures and will take them as an injected option, so this module keeps making no network call and its tests stay fixture-only.

No behaviour change — two strings, a docblock, the assertions that pin them, and an appended task-doc section. Verification from the committed tree: **1772 tests (1717 + 55), 0 fail, 0 skipped**, against a base of **1742 (1687 + 55)**; `eslint scripts` exit 0 at the pinned 9.24.0; store output unchanged apart from that one reason line. The base is this branch's own tip with only its three files removed rather than a `main` sha, since `main` moved during the work and the branch has been brought up to it.

hotsunchip (dismissed):
Approved but need to check again for sure.

dlgpdmsly2:
Pushed `43acbddc`. All three addressed.

**`cost-latency.mjs` — the unsized reason named a cause it cannot know.** The message said size was unknown *"(no frozen corpus item under this root)"*. This function receives a sizes map and has no idea where the entries came from: an item the caller could not read, one it never looked up, or a `meta.json` whose line counts would not parse are three different things to go and check. It now states the consequence and names no cause. Item identification and the per-item de-duplication are unchanged, and no size provenance is inferred from the map.

The existing test caught the change, which is what it was for. It is widened to pin both halves: that the message carries no cause claim, and that **one unsized item across three replicates produces one reason rather than three** — the de-duplication behaviour nothing covered before.

**Task doc — the stale verification bullet.** It still cited the by-hand comment-edit-history method and the 2.5–14.4 range it produced. Both are marked superseded and point at §7's settled `coderabbit-start-marker-to-first-finding` interval — **2.6–14.4 min, median 6.8, n=7/7** — with §4 retained as history, since the mechanism it proposed is wrong by 8× and a deleted wrong answer gets reinvented.

**"undefensible" → "indefensible"** in both places.

Verification from the committed tree: **rest lane 1730 pass, 0 fail, 0 skipped** against a base of 1700 (**+30**); **iso lane 56 pass, 0 fail**; `eslint scripts` exit 0 at the pinned 9.24.0; **40 mutations, 40 caught**, two of them new — one restoring the cause claim, one dropping the de-duplication. The base is this branch's own tip with only its three files removed rather than a `main` sha, since `main` has moved five times during this work.

⚠ **One thing not fixed, and not mine.** The iso lane is flaky on my machine: across 13 runs it gave 0–3 failures, always one of two tests in `eval/run.test.mjs` that snapshot `os.tmpdir()` and assert no new directory appeared — *"a failed item KEEPS its raw panel output"* and *"a throw inside the item loop still deregisters the worktree"*. **Base and this branch behave identically and both reach 56/56**, so this diff is not the cause, and neither test is in a file it touches. The mechanism looks local rather than code: `eval-item-*` debris in my temp directory grew from 720 to 986 over the session, ~13 per run, which is what an intermittently denied `rmSync` leaves behind — and a cleanup that fails is exactly what breaks a "nothing new appeared" assertion. Flagging it rather than repairing it here, since fixing another change's flake inside this diff would make the diff unreviewable.

hackerwins (approved):
Thanks for your contribution.

PR #794, opened 2026-08-12
Choose the cross-source token metric by arm rather than by run

dlgpdmsly2:
## Summary

`tokenOverlap` divides shared tokens by the **smaller** token set. Upstream chose that
deliberately, for a reason it states: *"the panel restates one defect at different levels of
detail, and Jaccard penalises the extra specifics in the longer wording."*

That reason is about **one reviewer restating itself** — but `gateFor` routes
same-arm-different-run down the cross-source path, so the rationale governed a population it was
never written for. The metric is now selected by **arm**:

- **same-arm** pairs keep `tokenOverlap` and `DEFAULT_SIMILARITY`, unchanged;
- **cross-arm** pairs get `crossArmTokenOverlap` (Dice) at `CROSS_ARM_SIMILARITY = 0.22`.

**The gate structure is untouched in both cases.** Same-arm-cross-run keeps L2's location and
anchor rules exactly, which is why `eval/reliability.mjs` reproduces byte for byte.

Three files: the matcher, its tests, and a task doc.

## Why

`harvest.mjs` uses this verdict to decide when one reviewer's finding restates another's. A match
**suppresses** the candidate — a false one deletes a real finding from the corpus with no later
pass able to recover it, while a false non-match files a duplicate a curator strikes out in one
line.

Three callers reach this code with three different shapes, and containment cannot see two of them
because it simply divides by whichever side is shorter:

| caller | pairs | shape |
|---|---|---|
| `eval/reliability.mjs` | panel across replicates | one reviewer — containment is right |
| `harvest.mjs` | panel vs CodeRabbit **prose** | 14 tokens vs 32 |
| `eval/complementarity.mjs` | panel vs CodeRabbit **title** | 14 tokens vs 7 |

Measured over **3608 pairs drawn from different pull requests**, so every pair is a known
non-match, plus the 6 pairs the incumbent metric calls `match`:

| shape | metric | false matches | true matches kept |
|---|---|---|---|
| PROSE | containment @ 0.30 | 16 / 3608 | 4 / 6 |
| PROSE | **dice @ 0.22** | **1 / 3608** | **3 / 6** |
| TITLE | containment @ 0.30 | 2 / 3608 | 6 / 6 |
| TITLE | **dice @ 0.22** | **2 / 3608** | **6 / 6** |

**On the threshold, which is the weakest part of this PR.** Dice rescales by `2·min/(|a|+|b|)`, so
the faithful translation of 0.3 depends on operand shape — 0.20 for the title-shaped caller, 0.18
for the prose-shaped one. 0.22 is the value in that region which holds the title-shaped caller
exactly while taking the prose-shaped caller's gain.

**There are no labels behind either number.** Recall is verified against the only true matches that
exist, and those were selected by the metric being replaced — that shows recall is not silently
destroyed and is not a recall measurement. The incumbent 0.3 had four hand-read rephrasings of one
defect behind it; this constant has less. This PR does not claim the new metric is more accurate.

## Linked Issues

None. This follows from measurement on `finding-match.mjs`'s own behaviour, not from a filed issue.

## Author checklist

## Verification

Measured from the **committed tree**, against a freshly measured baseline at the parent commit with
both trees set up identically:

| | tests (`rest + iso`) | fail | skipped |
|---|---|---|---|
| parent `a6f25053e` | 1684 + 55 = **1739** | 0 | 0 |
| this branch | 1691 + 55 = **1746** | 0 | 0 |

`eslint scripts` exits **0** on both (eslint 9.24.0, the lockfile pin).

**`eval/reliability.mjs`'s entire output is byte-identical** on the sample corpus — `diff` clean —
so 245 defect classes, 56/245 = 22.9% present in all three replicates and Jaccard 0.434 all stand.
That is the proof that only the metric changed and not the gate structure.

`eval/complementarity.mjs` moves on exactly one pair: overlap 6/166 = 3.6% → 7/165 = 4.2%, with the
21.1% ceiling unchanged.

**Seven mutations, each caught by the intended test** — `sameArm` defaulting to true (3 red),
`gateFor`'s unreadable-arm case (1), `gateFor` never same-arm (6), the cross-arm metric reverted to
containment (5), the bar at 0.20 and at 0.25 (1 each), and dropping the floor (1). The second of
those **survived the first version of its test**, which asserted `stats.gate.defaulted > 0` — true
under either routing; it now uses a pair whose verdict differs between the metrics.

⚠ `eval/run.test.mjs` failed 2 on **both** trees when the two lanes were run back to back — the
shared-state flake the repo already documents. Run serially, both are 55/55. It reproduces on the
untouched parent commit.

## Risk Assessment

- **User-facing risk:** none directly — no lens, no gate and no panel path reads this;
  `clusterFindings` goes through `findingSimilarity`, untouched. The real exposure is
  `harvest.mjs`'s corpus curation, which will suppress fewer CodeRabbit candidates as duplicates,
  so more get filed for a human to read. That is the recoverable direction; the failure it removes
  is not. It also loses one true suppression in six on that shape.
- **Data/security risk:** none. No I/O, no network, no persisted format changes.
- **Rollback plan:** revert. Nothing has persisted a score computed with either metric.

## Notes for Reviewers

- **AI assistance:** this PR was written with Claude Code — the measurement, the implementation,
  the tests and this body. Every number quoted was produced by running the code against stored
  review data, not estimated.
- **The trap worth reviewing closely:** `sameArm` is read from `opts`, never off the findings.
  `harvest.mjs` compares through `attributeToPanel`, whose `neutral()` helper carries no `arm` and
  no `run`, so `a.arm === b.arm` there would be `undefined === undefined` — `true` — and would hand
  the one production cross-arm caller the same-arm metric with every test still green. Absent
  provenance therefore defaults to cross-arm in both `matchAnchored` and `gateFor`.
- **Follow-up work:** one CodeRabbit finding can still match several panel findings. That is an
  assignment problem rather than a scoring one, and it is deliberately not in this PR.

### DIFF
--- a/docs/tasks/active/20260812-token-overlap-arm-aware-todo.md
+++ b/docs/tasks/active/20260812-token-overlap-arm-aware-todo.md
@@ -1,0 +1,170 @@
+# Choose the token metric by ARM, not by run
+
+A **new** document rather than a section on
+`archive/2026/08/20260803-finding-level-matching-todo.md`, which owns the matcher and is
+archived, or on `20260810-defect-class-grouping-todo.md`, which owns `groupFindings` and is a
+*consumer*. This one is about which similarity a pair gets and why.
+
+## The problem
+
+`tokenOverlap` scores a cross-source pair as **shared tokens over the smaller token set**. With
+the numerator fixed, that is *constant in the length of the longer operand*:
+
+| 3 shared, short side 6 tokens | other side 6 | 12 | 24 | 48 |
+|---|---|---|---|---|
+| containment | 0.50 | 0.50 | 0.50 | 0.50 |
+| dice | 0.50 | 0.33 | 0.20 | 0.11 |
+
+Upstream chose containment deliberately, and the reason is written down: *"the panel restates one
+defect at different levels of detail, and Jaccard penalises the extra specifics in the longer
+wording, which is precisely the wrong behaviour here."*
+
+**That is an argument about one reviewer restating itself.** It is not an argument about two
+different reviewers, where a long text covering a short one is the arithmetic of shared file
+vocabulary rather than evidence of agreement. And `gateFor` routes **same-arm-different-run down
+the cross-source path**, so the argument's own population and the population it governs had come
+apart. Three live callers, three shapes, one metric:
+
+| caller | pairs | shape | wants |
+|---|---|---|---|
+| `eval/reliability.mjs:535` | panel k1 vs k2 vs k3 | one reviewer, two replicates | containment |
+| `harvest.mjs:1642` | panel vs CodeRabbit **prose** | 14 tokens vs 32 | dice |
+| `eval/complementarity.mjs:638` | panel vs CodeRabbit **title** | 14 tokens vs 7 | dice |
+
+⚠ **The prose row is easy to get backwards.** `attributeToPanel`'s `neutral()` reads `f.summary`,
+but the call site one frame up has already put `finding.detail` into that field — *"Prose for the
+token comparison, the whole body for the anchor layer."* The field is named `summary` and holds
+the prose.
+
+**Why the cross-arm half is worth changing.** `harvest.mjs` uses the verdict to decide when one
+reviewer's finding restates another's, and a match **suppresses** the candidate: a false one
+deletes a real finding from the corpus with no later pass able to recover it, while a false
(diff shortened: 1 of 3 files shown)


dlgpdmsly2:
@claude review

hotsunchip (approved):

PR #799, opened 2026-08-12
Measure CodeRabbit's review latency from its own start marker

dlgpdmsly2:
## Summary

`eval/adapters/coderabbit.mjs` now reports how long a CodeRabbit review took, beside what it
found. Two intervals are emitted under separate keys, each naming what it is measured from:
`self_timed` (CodeRabbit's own marker comment → its first in-window finding) and `push_proxy`
(the earliest check run on the frozen commit → the same finding). The trigger that caused the
review is read from CodeRabbit's own marker, twelve reasons a duration can be missing are
declared and printed at zero, and a `gh` failure to expand `{owner}/{repo}` now aborts instead
of degrading to an empty result. `commitCheckRuns` is composed from `gh-checks.mjs`; the record
shape, the window rule and the severity translation are unchanged.

## Why

**A duration needs a start, and the two places previously searched do not hold one** —
`buildItemMeta` records `merged_at`, and the commit's *author* date is not a push time. Two
starts are available: the check runs on the frozen commit, and the HTML marker CodeRabbit
stamps into a comment when it takes a job (per invocation for a review command, once per pull
request for its status comment). Both are read; neither is called "the latency", because they
are not the same interval.

**A push-anchored figure measures the wrong thing whenever a human asked for the review, and
that is 2 of 7 corpus items.** PR #549 reads **183.7 min** from its earliest check run — 26× the
median of the other six — and **7.8 min** from CodeRabbit's own acknowledgement, because a
maintainer wrote `@coderabbitai review this.` at 03:58:27, CodeRabbit acked ten seconds later
and posted at 04:06:25. PR #605 has the same defect behind an unremarkable **9.8 min**, so any
outlier rule that catches the first keeps the second. Where the push is not what the review
answered, the proxy figure is reported and marked not poolable rather than dropped.

**`updated_at` is not read anywhere.** `(created_at, updated_at)` on the status comment looks
like a duration CodeRabbit timed itself and is not one — the last edit is the last edit of
anything. On PR #415 it reads 53.7 min against a true 6.6.

**A missing duration read as zero makes the other reviewer look instantaneous**, so twelve
absences are declared and every one is printed at zero. Three are one sentence in English and
different facts: a clean review, output we could not read, and a review of a later snapshot.
The one that has never occurred — a frozen commit with no check runs — is the one whose
mishandling nobody would notice.

**One failure is total rather than per-endpoint.** `gh` expands `{owner}/{repo}` from the
working directory's git remote; with none and no `GH_REPO`, every call fails identically, each
degrades to "absent", and the output is a complete and entirely empty census. #790 closed the CI
half with a workflow-level `GH_REPO` and an assertion in `checks.test.mjs`; no workflow
assertion can see a local invocation, which is where this CLI runs, so the message `gh` emits is
turned into a refusal.

This changes no reported number today: nothing consumes the latency yet.

## Linked Issues

None. Follows #739 (the adapter), #771 (in-window identity) and #790 (`GH_REPO`).

## Author checklist

## Verification

Local, from the **committed tree**, base `f4d0d65d6`, the same `node_modules` linked into both
trees before either was measured, `agent:tests` as its two invocations (`rest + iso`):

| | rest | iso (`eval/run.test.mjs`) | total |
|---|---|---|---|
| base `f4d0d65d6` | 1687 / 0 fail / 0 skip | 55 / 0 / 0 | **1742** |
| this branch | 1702 / 0 fail / 0 skip | 55 / 0 / 0 | **1757** |

**+15 tests, 0 fail, 0 skip on either tree.** `eslint scripts` exits 0 on the lockfile's pinned
9.24.0 on both. **16 mutations, 16 caught, 0 survived**, each by a test naming the right thing.

Over the frozen corpus, from the branch tree: 7 of 7 items time a review — self-timed
**2.6–14.4 min, median 6.8**; push-proxy 2.8–183.7, poolable on the 5 automatic items
(2.8–14.8, median 6.7); `trigger automatic=5 on-demand=2 unknown=0`; every absence at 0. The
record census is byte-identical to `main`'s (30 records, `in-window=30`, `severity major=3
minor=13 nit=14`). With `GH_REPO` unset in a directory with no remote, the CLI exits 1 with the
refusal instead of printing an empty result.

## Risk Assessment

- **User-facing risk:** none — nothing in the gating path is touched, no lens, gate or
  `review-panel.mjs`. This module is an offline analysis CLI that reads and prints.
- **Data/security risk:** two additional read-only `gh api` GETs per item (`issues/{n}/comments`,
  `commits/{sha}/check-runs`), no writes, no new scope. The CodeRabbit author gate is
  `CODERABBIT_LOGINS`, exact, and it now guards the clock as well as the count: the markers are
  public strings on public pull requests, so an ungated read would let anyone move the start of
  the interval by pasting one into a comment. A test asserts that a human, and a
  `coderabbitai-x` lookalike, cannot.
- **Rollback plan:** revert. Nothing consumes the new fields, no stored data changes, and the
  existing return shape is widened rather than altered.

## Notes for Reviewers

- **AI assistance:** written with Claude Code — the module, its tests, the task doc and this
  body. Every figure quoted was printed by the code or by a throwaway script against the GitHub
  API; the fixtures are verbatim API responses with their comment and check-run ids, re-fetchable
  from the header of the test file.
- **Where to start:** `SELF_TIMED_INTERVAL` and `TRIGGERS` in
  `scripts/agent/eval/adapters/coderabbit.mjs` are the two docblocks that carry the argument;
  `LATENCY_ABSENT` is the vocabulary. The task doc is
  `docs/tasks/active/20260812-coderabbit-latency-todo.md`.
- **Follow-up work:** a scorer consuming this owes the reader one thing this PR deliberately does
  not do — a statement of what the two arms' latencies are *not* measuring. Ours is a panel
  process on an offline replay; these are a production reviewer end to end. No ratio between the
  arms is computed here and none should be.
- **Deliberately not done:** no central figure in the adapter (a median belongs to one scorer,
  not two); nothing added to a frozen `meta.json`; `sdk_duration_ms_sum` is not read and is not
  named anywhere in the file.

## Summary by CodeRabbit

* **New Features**
  * Added CodeRabbit review latency tracking, including self-timed and push-proxy intervals.
  * Added trigger classification, review-start detection, absence reasons, and poolability reporting.
  * Added latency summaries to command-line output and evaluation reports.
  * Added support for retrieving review comments and check-run information.

* **Documentation**
  * Documented latency measurement rules, reporting caveats, validation requirements, and known limitations.

* **Tests**
  * Added comprehensive coverage for latency calculations, reporting, marker selection, endpoint failures, and end-to-end data propagation.

### DIFF
--- a/docs/tasks/active/20260812-coderabbit-latency-todo.md
+++ b/docs/tasks/active/20260812-coderabbit-latency-todo.md
@@ -1,0 +1,257 @@
+# Measure how long a CodeRabbit review takes, and say what the clock is measured from
+
+*Extends `archive/2026/08/20260807-coderabbit-adapter-todo.md` (#739) and the
+follow-up `20260811-coderabbit-window-identity-todo.md` (#771). The record shape,
+the window rule and the severity translation are unchanged; this adds a duration
+beside them and one refusal to the fetch.*
+
+The adapter could say **what** CodeRabbit found on the frozen snapshot and not **when**
+it landed. It now emits two named intervals, the trigger that caused the review, and
+twelve reasons a duration can be missing — none of which is zero.
+
+## The problem
+
+Three separate defects, and only the first is the obvious one.
+
+**1. The start instant was looked for in the wrong two places and declared absent.**
+A duration needs a start, and neither `buildItemMeta` (which records `merged_at`) nor
+`fetchCodeRabbitPr` (which reads commit **author** dates) holds one. Concluding from
+those two that no start exists anywhere is an inference from two call sites. Two are
+available: **the check runs on the frozen commit** (`GET
+/repos/{owner}/{repo}/commits/{sha}/check-runs`, which resolves on 7 of 7 pilot items)
+and **CodeRabbit's own marker comments**, which it stamps when it takes a job.
+
+**2. A push-anchored figure measures the wrong interval whenever a human asked for the
+review.** Anchored to the earliest check run on its frozen commit, `pr-549` reads
+**183.7 min** — 26× the median of the other six, enough to move the mean from 7.0 to
+32.6. It was never a slow review. The pull request's timeline:
+
+```
+2026-07-26T01:02:46Z  verify-self starts on 158c6faa…   ← the anchor
+2026-07-26T03:58:27Z  hackerwins: "@coderabbitai review this."
+2026-07-26T03:58:37Z  CodeRabbit acknowledges (invocation ad4e5877…)
+2026-07-26T04:06:25Z  CodeRabbit's finding on 158c6faa…
+```
+
+**8.0 min from the ask, 7.8 min from CodeRabbit's own acknowledgement**, and the other
+176 minutes are a human deciding to ask. `pr-605` is the same shape — two asks, the
+review 7.6 min after the second acknowledgement — and it is the dangerous one, because
+its push-anchored figure is an entirely ordinary **9.8 min**. Nothing about the number
+says it is timing the wrong thing. Any outlier rule that caught pr-549 would have kept
+pr-605.
+
+**3. `gh` reports an unexpandable `{owner}/{repo}` as an empty result.** Run from a
+directory with no git remote and no `GH_REPO`, every call fails identically with
+*"unable to expand placeholder in path"*, each one degrades to `absent`, and the output
+is a complete, plausible, entirely empty census — previously measured as 0 CodeRabbit
+records across a whole corpus version. #790 closed the CI half of this by setting a
+workflow-level `GH_REPO` and asserting it in `checks.test.mjs`; no workflow assertion
+can see a local invocation, which is where this CLI runs.
+
+## The change
+
+**Two intervals, each naming what it is measured from**, because "CodeRabbit's latency"
+is not one number:
+
+| key | from → to | pilot, n=7 |
+|---|---|---|
+| `self_timed` · `coderabbit-start-marker-to-first-finding` | CodeRabbit's own marker comment → its first in-window finding | 2.6–14.4 min, **median 6.8** |
+| `push_proxy` · `earliest-check-run-start-to-first-finding` | earliest check run on the frozen commit → the same finding | 2.8–183.7, poolable on 5: 2.8–14.8, median 6.7 |
+
+`self_timed` is the one to prefer, for two measured reasons: it needs no guess about
+what triggered the review, and it is the only one of the two that survives an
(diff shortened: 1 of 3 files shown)


dlgpdmsly2:
Fixed the three inline findings; skipped one nitpick with a reason. Pushed as `4e80ecaf3a29`.

**The census bucket was the real one.** `latencyCensus` opens a key for an absence nobody declared, so a new flavour shows up as an unexplained bucket rather than disappearing into a known one — and the formatter iterated `LATENCY_ABSENT` alone, which left that bucket counted and invisible. It is now the exported `latencyAbsentLine`: declared rows in declared order, then any unrecognised ones marked, matching the wording `cost-latency.mjs` uses for an unrecognised `duration_source`. Two mutations pin it (hide the extra bucket; drop the declared zeros).

**Both off-by-one statements were wrong.** Six absences end the interval, not five — `posted-at-absent` is the sixth and `endOfReview` returns all six — corrected in the docblock, the task doc and the test's grouping comment. And the production figure is **18.7 min, not 18.6**: 18.6 is when the first lens check run appeared, 18.7 is when the last one completed, which is what "took" means there and what the task doc already said.

**The null `t0` I left alone, and stated the invariant instead.** It cannot fire: both producers set `started_at` from the very string whose `at()` they already found finite, and both report a start only in that case — which the preceding line checks. `span` is not exported, so no test could prove such a branch fires, and this file's own rule for that (see `checkArmFields`) is that a guard nothing can prove fires is decoration. The invariant is now written where the subtraction happens. Worth flagging that you read the shape correctly: had it been reachable, `t1 - null` would have produced a ~57-year duration rather than an absence.

**The fence label I skipped.** There is no markdownlint anywhere in this repository and no docs lint script, and the sibling task docs run 23 bare fences to 3 labelled — so labelling this one would depart from the local convention to satisfy a linter that does not run here.

Verification, from the committed tree: **1703 + 55 = 1758, 0 fail, 0 skip** against a base of 1687 + 55 = 1742 (**+16**), `eslint scripts` 0, **18 mutations 18 caught**. One thing to know if you re-run the lane locally: `eval/run.test.mjs` fails 2 of its 55 when `os.tmpdir()` already holds `eval-item-*` directories from earlier runs of the same file — the base tree fails the identical two under the same dirty `TMPDIR`, and both trees pass 55/55 under a fresh one. Pre-existing, from #682.

hotsunchip (approved):

PR #805, opened 2026-08-12
Write metrics into scores/ and render a comparison report into reports/

dlgpdmsly2:
## Summary

Adds the first writer for the two directories `eval/store.mjs` has documented since
#677 and never written to:

- `store.mjs` gains `putScore` / `getScore` / `putReport`, plus `configHashSegment` and
  `byConfigSegment` to turn a config hash into a directory name. **Scores and reports
  are re-writable**, unlike everything else on that surface.
- `eval/report.mjs` (new) files a scorer's `--json` under `scores/` and renders a
  markdown comparison out of `scores/` into `reports/<comparison id>.md`. The render
  path makes no network calls and invokes no scorer.
- `eval/report.test.mjs` (new), fixtures only — no store, no filesystem, no network.

Three new files, one existing file extended. No scorer, adapter, matcher, lens or gate
is touched. `--root` is required with no default, so the module cannot write inside
this repository.

## Why

**`scores/` had no writer, so no metric could be re-scored.** Every scorer in
`eval/` says the same thing in its own words — *"reads only; writes nothing into the
store"* — so a computed metric exists only as terminal output. The store README's
invariant is that metrics live in `scores/` precisely so a matcher or metric change can
be re-scored against cached artifacts without re-invoking the model. That is the cheap
half of this pipeline and it has already mattered: #780 moved a defect-class count from
219 to 245 with no model call, and there was nowhere to file either number.

**`scores/` is therefore re-writable, and `putRun`'s discipline is deliberately not
copied.** `putRun` refuses a differing `config_hash` because two reviewers' items must
never be filed under one run id — a statement about an *observation* of a
non-deterministic judge. A score is a *derivation* over observations that remain
immutable underneath it, so recomputing from them is safe by construction. The docblock
argues this at length because two of the module's three write paths are write-once and
the next reader will reasonably assume the third is.

**The renderer reads committed data and never calls a scorer.** Two scorers reach their
second arm through an adapter that makes live API calls, and with no repository context
that adapter returns zero records and a plausible empty result rather than an error — so
a report assembled by invoking them could print a one-armed comparison indistinguishable
from a two-armed one, and would not reproduce offline. Persisting first makes a missing
arm a missing *file*.

**Absence needs four words, not a blank cell.** An empty table cell can mean the scorer
is unbuilt, or that it ran and measured a real zero, or that the quantity cannot be
measured on this data at all, or that a real measurement was withheld for a thin
denominator. A reader acts differently on each, so each renders as its own sentence, and
a figure cannot be constructed at all without its `n` and its unit. Two of the five
sections have no scorer built today and say so by name; both states occur in the same
document, which is why the distinction is in the types rather than in a comment.

**Two path components become directory names**, so both go through the existing
`requireSegment` rather than being sanitised — a scorer id arrives from a command line,
and a report path built from an unvalidated one writes outside the store. The config
hash is checked against its format *before* the colon is replaced, so unlike the
sanitising this module already condemns the mapping is injective over its domain.

**This changes no existing behaviour and no existing number.** It adds a surface; nothing
in the product reads it.

## Linked Issues

None. Follows #677, which defined the layout these methods write to.

## Author checklist

## Verification

Measured locally, from the committed tree, with the two `node --test` invocations the
`agent:tests` lane uses since #774:

```
this branch    1737 (1736 pass · 0 fail · 1 skipped)  +  55 (55 · 0 · 0)
main d327ee410 1696 (1695 pass · 0 fail · 1 skipped)  +  55 (55 · 0 · 0)
               +41 tests · 0 new failures · skip count unchanged
eslint scripts (9.24.0, lockfile pin) — exit 0 on both trees
```

⚠ `eval/run.test.mjs` flakes on the machine this was measured on — two END-TO-END cases that
snapshot `os.tmpdir()`, **2 of 4 runs red on `main` alone** against 1 of 3 here, and this diff
touches neither `run.mjs` nor `run.test.mjs`. Pre-existing (#682); reported rather than repaired
inside an unrelated PR.

Both trees were extracted separately and had the same `node_modules` symlinked into them
before either was measured, so the skip count is comparable rather than an environment
artefact.

**51 mutations, 51 caught**, re-run against the final bytes of all four files. Two
initially survived and neither was dismissed as ineffective: deleting `_scorePath`'s scorer-id
check left every test green because `putScore` validates twice while `getScore` does not — the
read path was genuinely unguarded; and a hard-coded replicate count is invisible at K=1 (different
branch) and at K=3 (literal equals derived), so only a K=2 case discriminates. Both got the
missing assertion.

**The third commit answers four findings from the segmentation scorer's session**, which
round-tripped 149 real cells through this renderer without editing it: §5's caption predicted a
blank grid where 76 of 149 cells actually report (true per PR, false per finding — it needed its
unit); `renderCell`'s formatter parameter was dead at all nine call sites, so a real proportion
printed as `0.6944444444444444`; two cross-arm rows are 1.000 by construction because an inline
comment is anchored to a diff line, which is a fact about the comment API rather than a reviewer;
and an axis nobody can compute had no cell to live in, which inverted this module's own argument
about declared expectations. Validated against that real payload, not only fixtures.

**The second commit fixes four defects review found**, all live behind a green suite and three of
them printing a plausible number rather than failing: a partial per-run score set was silently
dropped so the section rendered fewer replicates than the header claimed; the bold total row
ignored a consistency guard every severity row honoured; a *present but hollow* proportion reached
the page as `(undefined/7)` and a missing nested block threw a `TypeError` that aborted the whole
render; and two caveats hard-coded facts about one corpus. See the task doc's ⑨–⑫.

End-to-end against the real pilot data, rendered under `env -u GH_REPO -u GITHUB_TOKEN
-u GH_TOKEN` with `PATH` stripped to `node` plus `/usr/bin:/bin`: every figure reproduces
the numbers already recorded for that dataset, and the committed code reproduces the
committed report byte for byte.

## Risk Assessment

- **User-facing risk:** none — nothing in the product imports `eval/`, and no gating path
  is touched. The panel cannot reach this code.
- **Data/security risk:** this is the risk line. Two things are new: a **write path** into
  a data root, and **two path components that become directory names**. Both are
  addressed by validation rather than sanitising, on the read path as well as the write
  path, with tests that assert nothing is created when a component is refused. `--root`
  has no default, so the module cannot write into this repository even if the flag is
  forgotten. Writes are atomic (temp file plus rename), reusing the helper already in the
  module. **`scores/` and `reports/` are intentionally overwritable** — a reviewer should
  read `SCORES_DIR`'s docblock and disagree there if anywhere. A score computed under a
  different reviewer lands in a different path by construction, so "re-score" and
  "different reviewer" are never the same write.
- **Rollback plan:** revert. Nothing reads these methods except the new module, and
  nothing in the repository invokes that module — no workflow, no script, no lane.

## Notes for Reviewers

- **AI assistance:** written with Claude Code. The design decisions, the mutation testing
  and the corrections in the task doc are recorded there; the task doc
  (`docs/tasks/active/20260812-eval-report-renderer-todo.md`) is the long form of this
  body and worth reading for the *Corrected while building* section in particular — the
  first entry is a bias the first draft introduced into a published ratio.
- **The two halves are one PR on purpose.** Persisting and rendering are the same job —
  the documented paths getting a writer — and the renderer is untestable end to end
  without something committed to read. Happy to split them if you would rather review
  them separately.
- **Follow-up work:** the module lists five report sections and two have no scorer yet
  (`cost-latency-v1`, `segmentation-v1`); both render as "not computed" and grow a
  section when their scorers land. No CI lane invokes any of this — that is deliberate
  and separate.

### DIFF
--- a/docs/tasks/active/20260812-eval-report-renderer-todo.md
+++ b/docs/tasks/active/20260812-eval-report-renderer-todo.md
@@ -1,0 +1,214 @@
+# Persist scores and render a comparison report — `eval/report.mjs`
+
+## The problem
+
+`scripts/agent/eval/store.mjs` has documented two directories since #677 and **nothing
+has ever written to either of them**:
+
+```
+scores/                 metrics — RE-SCOREABLE, never written under runs/
+  per-run/<run_id>/<scorer_id>.json
+  by-config/<config_hash>__<corpus_version>/<scorer_id>.json
+reports/<comparison_id>.md   static A-vs-B comparisons
+```
+
+`EvalStore` exposes eleven methods and not one of them is a score or a report method:
+`putCorpusItem`, `putCorpusManifest`, `getCorpusItemInput`, `getCorpusManifest`,
+`getCorpus`, `listCorpusItems`, `putRun`, `getRun`, `putItem`, `getItem`, `hasItem`,
+`listItems`, `listRuns`. The three merged scorers (`volume-mix.mjs`,
+`complementarity.mjs`, `reliability.mjs`) each say the same thing in their own words —
+*"Reads only; writes nothing into the store"* — so **every metric this project has
+computed exists only as terminal output.**
+
+Two consequences, and the second is the one that matters:
+
+1. **A metric cannot be re-scored against its predecessor.** #780 moved the pilot's
+   defect-class count from 219 to 245 with no model call. That is the cheap half of
+   this benchmark working exactly as designed — and there is nowhere to file either
+   number, so the delta lives in a chat log.
+2. **A report assembled by invoking the scorers would not be reproducible, and would
+   fail silently rather than loudly.** `complementarity.mjs` and `volume-mix.mjs` reach
+   the CodeRabbit arm through `adapters/coderabbit.mjs`, whose `corpusRecords` calls
+   `gh api`. With no repository context that adapter yields **zero CodeRabbit records
+   and a plausible empty result** — not an error. So a renderer that fetched would
+   print a one-armed comparison that looks exactly like a two-armed one.
+
+## The change
+
+Three files. Nothing existing is modified except `store.mjs`, which gains methods.
+
+**`store.mjs` — `putScore` / `getScore` / `putReport`.**
+
+- 🔴 **`scores/` and `reports/` are RE-WRITABLE**, deliberately and against the grain
+  of the rest of the module. `putRun`'s refuse-on-conflict discipline is not copied,
+  and its reasoning does not transfer: `putRun` refuses a differing `config_hash`
+  because two reviewers' items must never be filed under one run id, which is a
+  statement about an *observation* of a non-deterministic judge. A score is a
+  *derivation* over observations that remain immutable underneath it. Argued at
+  `SCORES_DIR`, because the next reader will assume the store is uniformly write-once
+  — two of its three write paths are.
+- **Every path component is validated by the existing `requireSegment`**, not
+  sanitised, and not by a second validator. `<scorer_id>` and `<comparison_id>` arrive
+  from a command line, and `reports/../../README.md` built from an unvalidated one
+  writes outside the store.
+- **`configHashSegment` derives `sha256-<hex>` from `sha256:<hex>`.** This is *not*
+  the fork's `[:/\\] → -` mangle that `_runDir` condemns: the input is refused unless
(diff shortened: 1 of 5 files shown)


dlgpdmsly2:
@claude review

hotsunchip (dismissed):

harrykim8672 (approved):

Issue #793, opened 2026-08-12
Docs: Highlight "None" stores backgroundColor:"" instead of clearing it


hwisoo-kim:
## What happens

Clearing a highlight with **None** does not clear it. It writes `backgroundColor: ""`
into the run, so the run keeps a dead value and never merges back into its
neighbours.

1. Select part of a paragraph and apply a highlight colour
2. Select the same range, open **Highlight color**, choose **None**

The highlight stops rendering, so it looks fixed. The stored document does not go
back to what it was.

## Repro

Driving the real store helper with exactly what the toolbar calls:

```ts
const block = { id: 'b1', type: 'paragraph', style: {},
                inlines: [{ text: 'abcdef', style: {} }] };

const on  = applyInlineStyle(block, 2, 4, { backgroundColor: '#FFF176' });
const off = applyInlineStyle(on,    2, 4, { backgroundColor: '' });   // "None"
```

```
AFTER ON : [{"text":"ab",...{}}, {"text":"cd", "backgroundColor":"#FFF176"}, {"text":"ef",...{}}]
AFTER OFF: [{"text":"ab",...{}}, {"text":"cd", "backgroundColor":""},        {"text":"ef",...{}}]
```

Still three runs. `inlineStylesEqual({ backgroundColor: '' }, {})` is `false`.

## Why

**None** passes an empty string rather than clearing the property
(`packages/frontend/src/components/text-formatting/text-format-group.tsx:344`):

```tsx
onReset={() => handleColor("")}
```

which reaches `editor.applyStyle({ backgroundColor: "" })`, and the store merges the
incoming style over the existing one without treating `""` as "remove"
(`packages/docs/src/store/block-helpers.ts:346`):

```ts
newInlines.push({
  text: inline.text.slice(overlapStart, overlapEnd),
  style: { ...inline.style, ...resolvedStyle },
});
```

`normalizeInlines` then cannot re-merge the run, because `inlineStylesEqual` compares
colours with `storedColorsEqual`, and `'' === undefined` is false — an unset
background and one explicitly set to `""` are different values.

## Why it matters beyond the extra run

- The document keeps accumulating runs that differ only by a meaningless field, and
  the split is permanent — no later edit re-merges them.
- `""` is not a colour. Anything reading `backgroundColor` and treating a present
  value as "there is a highlight" sees one, and export paths may emit it.

## Related

Same shape as #749 (turning an inline style off stores an explicit `false`): the
[...]

Issue #792, opened 2026-08-12
Docs: applying a block type leaves the Text style control showing the old style


hwisoo-kim:
## What happens

Applying a block type leaves the **Text style** control showing the old value. Put
the caret in a paragraph, apply `Heading 2`, and the toolbar button still reads
`Normal text` — and re-opening the menu still shows `Normal text` checked. The block
really did change; only the control is stale.

It corrects itself on the next caret move or any *other* formatting action, which is
what makes it easy to miss: click Bold, or move the caret, and the label jumps to
`Heading 2` with no further block-type change.

## Why

The toolbar re-derives its state from cursor-move callbacks
(`packages/frontend/src/app/docs/docs-formatting-toolbar.tsx:322`):

```ts
refresh();
const unsubscribe = editor.onCursorMove(refresh);
```

Style operations that do not move the caret synthesise one by calling
`notifyStyleApplied()`, which exists for exactly this
(`packages/docs/src/view/editor.ts:1237`, and the note at `:972` — "so toolbar
pickers re-derive"):

```ts
function notifyStyleApplied(): void {
  ...
  fireCursorMoveCallbacks(cursor.position, selRange);
}
```

`setBlockType` is the one that does not (`packages/docs/src/view/editor.ts:3034`):

```ts
setBlockType(type: BlockType, opts?: {...}) {
  docStore.snapshot();
  doc.setBlockType(cursor.position.blockId, type, opts);
  invalidateLayout();
  render();
},
```

Its siblings all do — `applyBlockStyle` (`:2980`), and the calls at `:3047`,
`:3080`, `:3088`, `:3096`, `:3176`. `setBlockType` is the outlier.

## Both entry points are affected

The keyboard shortcut is **not** a workaround. It does not go through the method
above, but it does not notify either
(`packages/docs/src/view/text-editor.ts:1028`):

```ts
this.saveSnapshot();
...
this.doc.setBlockType(block.id, 'heading', { headingLevel: level });
this.invalidateLayout();
this.requestRender();
```

`text-editor.ts` has no style-notification path at all, so `⌘⌥2` leaves the same
control equally stale. Menu and shortcut behave identically here.

## Suggested fix

Call `notifyStyleApplied()` at the end of `EditorAPI.setBlockType`, matching its
siblings. The shortcut path needs its own answer, since it bypasses that method.

## Provenance

Found by the UI hunter (run `colour1`) as a `dom.snapshot not-equals` prediction on
[...]

Issue #783, opened 2026-08-11
Docs: un-listing a heading downgrades it to a paragraph, losing the heading level


hwisoo-kim:
## What happens

Applying a list to a heading and then removing the list leaves a plain
paragraph. The heading level is destroyed, silently, by an action that looks
reversible.

1. Put the caret in a `Heading 2` block
2. Click **Bulleted list** — the block becomes a list item
3. Click **Bulleted list** again to turn it off

Expected: a `Heading 2` again. Actual: body text.

Google Docs and Word both keep the heading through this round trip (the bullet
is applied *to* the heading), so this is a deviation rather than a house style.

## Why

Two independent pieces, both required.

`setBlockType` clears the heading level unconditionally and only ever restores
it when the new type is `heading` — so the level is gone the moment the list is
applied (`packages/docs/src/store/memory.ts:249`):

```ts
block.type = type;
delete block.headingLevel;
delete block.listKind;
delete block.listLevel;
if (type === 'heading') {
  block.headingLevel = opts?.headingLevel ?? 1;
}
```

`toggleList` then hardcodes `'paragraph'` on the way out, so there is nothing to
restore from even if the level had survived
(`packages/docs/src/view/editor.ts:3100`):

```ts
toggleList(kind: 'ordered' | 'unordered') {
  docStore.snapshot();
  forEachBlockInSelection((block) => {
    if (block.type === 'list-item' && block.listKind === kind) {
      doc.setBlockType(block.id, 'paragraph');
```

Nothing anywhere records the previous block type — there is no `prevType` /
`previousType` / `restoreType` in either file.

## Repro at the model level

Driving the store directly with exactly the two calls `toggleList` makes:

```ts
const block = { id: 'b1', type: 'heading', headingLevel: 2,
                inlines: [{ text: 'Quarterly results', style: {} }], style: {} };
const store = new MemDocStore({ blocks: [block] });

store.setBlockType('b1', 'list-item', { listKind: 'unordered', listLevel: 0 });
store.setBlockType('b1', 'paragraph');

store.findBlock('b1');   // { type: 'paragraph' }  — headingLevel is gone
```

## Scope

This is about the **round trip**, not about how a list item is styled while it
is a list item. `blockStyleId` mapping `list-item` to `normal`
(`packages/docs/src/model/named-styles.ts:97`) is a styling derivation for the
block's current state and is working as intended; it does not speak to what the
block should return to when the list is removed.

[...]

Issue #749, opened 2026-08-10
Docs: turning an inline style off stores an explicit false, so the run never re-merges


hwisoo-kim:
## Summary

Toggling a boolean inline style **on and then off** on the same selection does not restore the document. The run stays permanently split, and the fragment keeps an explicit `italic: false` (or `bold: false` / `underline: false`) rather than having the key removed.

Rendering is unaffected today — this is a model/structure issue, with one user-visible consequence noted at the bottom.

## Minimal reproduction

1. In a doc, type `abcdef`.
2. Read the runs — the paragraph is a **single run**.
3. Put the caret before `c` and Shift+ArrowRight twice to select `cd`.
4. Click **Italic** → `cd` splits out as `italic: true`. ✅ correct
5. Click **Italic** again → expected: back to one run. Actual:

```jsonc
// baseline (step 2)
[{ "text": "abcdefSmall ", "fontSize": 11 }]

// after italic ON (step 4)
[{ "text": "ab", "fontSize": 11 },
 { "text": "cd", "fontSize": 11, "italic": true },
 { "text": "efSmall ", "fontSize": 11 }]

// after italic OFF (step 5) — still three runs, and a dead flag
[{ "text": "ab", "fontSize": 11 },
 { "text": "cd", "fontSize": 11, "italic": false },   // ← was absent before
 { "text": "efSmall ", "fontSize": 11 }]
```

Every on/off cycle on a fresh sub-range adds another permanent fragment.

## Why the runs never re-merge

`normalizeInlines` merges adjacent runs only when `inlineStylesEqual` says their styles match, and that comparison is strict:

```ts
// packages/docs/src/model/types.ts
export function inlineStylesEqual(a: InlineStyle, b: InlineStyle): boolean {
  return (
    a.bold === b.bold &&
    a.italic === b.italic &&
    ...
```

```ts
// packages/docs/src/store/block-helpers.ts
if (last && inlineStylesEqual(last.style, inline.style)) {
  last.text += inline.text;
} else {
  merged.push({ text: inline.text, style: { ...inline.style } });
}
```

`false === undefined` is `false`, so a run carrying `italic: false` can never merge with an identical-looking neighbour that simply lacks the key. Clearing the key instead of writing `false` would let `normalizeInlines` do its job.

## The consequence beyond tidiness

[...]

PR #806, opened 2026-08-12
Give the sheet surface the toolbar its defects live behind

hwisoo-kim:
## Summary

Every defect this hunter has filed came from a **toolbar control** — five for five
(#715 B/I/U/S, #749 style toggle, #783 list button, #792 style menu, #793 highlight
palette). The sheet surface had no toolbar at all: the harness gated it to `doc`, so the
sheet persona was asked to run the round-trip shape with nothing to run it through. Its
entire live record is one false finding and one empty run.

The real `FormattingToolbar` is now mounted there. Only `spreadsheet` is required — the
harness already holds it — and `MemStore` genuinely implements the style API, so these
controls really mutate state rather than no-op the way undo does.

Two readers make the result observable:

| reader | answers |
|---|---|
| `sheet.rangeStyles` | the `{range, style}` patches the toolbar appends |
| `sheet.activeCellStyle` | what the app COMPUTES for the active cell (sheet→col→row→cell) |

Each description says which side it is on, because a pair like this without that is how
`doc.styleSummary` invited false findings for months.

## The per-cell reader this started as was wrong

I scoped `sheet.cellStyle(sref)` reading `cell.s`, and **the oracle check killed it
before it shipped.** `setRangeStyle` appends a patch and then calls
`applyStylePatchToExistingCells`, which skips any cell not already carrying its own
style:

```js
const existing = await this.store.get(anchor);
if (!existing?.s) continue;
```

Measured: with B1 selected and holding `"Label"`, clicking Bold gave an effective style
of `{b:true}` and a per-cell reading of `null`. **A reader whose common case is a silent
null is a trap, not a sensor** — it would have looked exactly like a broken toolbar. It
was replaced with the one that points where the data actually is.

## Mounting a toolbar creates traps; the brief names them

A trap the brief does not name is a false finding. `sheet-author.md` now states:

- **styling does not land on the cell** — read `sheet.rangeStyles`, not a per-cell style
- **patches ACCUMULATE** — toggling Bold on/off appends `{b:true}` then `{b:false}`
  rather than removing the first, so a round trip is *not* `equals` on that list and
  should be predicted on the computed style instead
- **four buttons are unwired here** — `Insert chart`, `Insert image`, `Data validation`,
  `Conditional formatting` open panels this harness does not mount, so their handlers
  are absent and the click is swallowed
- **`Undo`/`Redo` are now VISIBLE and still cannot work** — the same `MemStore` no-op as
  `sheet.canUndo`. This is the single most likely false finding the change introduces

The optional handlers are deliberately not stubbed: inventing behaviour the product does
not have would be worse than naming the dead ends.

## The rubric test expired, again

It asserted `no formatting toolbar` — correct when written, wrong on merge. It now pins
the new facts including both traps. **That assertion has now expired twice**: a brief
telling the doc explorer colour was out of bounds outlived the reader that made colour
observable, and the capability sat unused until someone noticed. The test carries the
note that a rubric stating a surface fact is a claim with an expiry date.

## Test plan

`verify:hunt:oracles` — new `checkSheetToolbar` selects B1 through `sheet.cellCenter`
(the same path the hunter must use), clicks the real Bold button, and asserts a bold
patch appears in `sheet.rangeStyles` **and** that the computed reader agrees. Both new
readers are registered in `checkReaderRegistry`, so neither can inherit zero coverage.

It deliberately does **not** assert what the second Bold click leaves behind.
`toggleRangeStyle` computes `!effective[prop]` and writes it, so bold-off stores
`b: false`. Whether that is a defect is a real question — an explicit `false` is how a
cell overrides a `true` inherited from a row or column style — and pinning either answer
here would bless a defect or fail the lane over one. That judgement belongs to the panel.

Mutation-tested:

| mutation | result |
|---|---|
| unmount the sheet toolbar | `no toolbar is mounted on the sheet surface` |
| make `sheet.rangeStyles` always report empty | `clicking Bold appended no range style` |

Also: `agent:tests` on Node 22 CI-faithful **1755 pass / 0 fail**, frontend `tsc` clean,
`lint:scripts` and `frontend lint` clean, `personas.json` parses.

## Summary by CodeRabbit

* **New Features**
  * Added a formatting toolbar to the spreadsheet view.
  * Added support for applying bold formatting to selected cells.
  * Added visibility into stored range styles and the active cell’s computed formatting.

* **Documentation**
  * Updated spreadsheet guidance to describe toolbar behavior, supported actions, and formatting workflows.

* **Tests**
  * Added end-to-end coverage for applying and verifying formatting on spreadsheet cells, including selected-cell styling and computed results.

### DIFF
--- a/packages/frontend/scripts/verify-hunt-oracles.mjs
+++ b/packages/frontend/scripts/verify-hunt-oracles.mjs
@@ -28,6 +28,7 @@ const HOST = "127.0.0.1";
 const PORT = Number(process.env.HUNT_ORACLES_PORT || 4178);
 const READY_SELECTOR = "[data-testid='hunt-harness-root'][data-hunt-harness-ready='true']";
 const HOST_TESTID = "hunt-harness-host";
+const TOOLBAR_TESTID = "hunt-harness-toolbar";
 
 /** How long silence must hold before a negative control counts as quiet. */
 const QUIET_WINDOW_MS = 250;
@@ -243,6 +244,12 @@ const READER_EXPECTATIONS = [
     v === null || v === undefined ? null : `a literal cell must have no formula, got ${JSON.stringify(v)}`],
   ["sheet", "sheet.activeCell", [], (v) => (typeof v === "string" && /^[A-Z]+\d+$/.test(v) ? null : "expected a cell ref")],
   ["sheet", "sheet.canUndo", [], (v) => (typeof v === "boolean" ? null : "expected a boolean")],
+  // A1 carries no style in the seed, so `null` is the right answer and an OBJECT would
+  // mean the reader had started resolving inherited values — which is the other
+  // reader's job, and the confusion this pair is named to avoid.
+  ["sheet", "sheet.rangeStyles", [], (v) => (Array.isArray(v) ? null : "expected an array of range-style patches")],
+  ["sheet", "sheet.activeCellStyle", [], (v) =>
+    v === null || (v && typeof v === "object" && !Array.isArray(v)) ? null : "expected a style object or null"],
   ["sheet", "sheet.cellCenter", ["B2"], (v) =>
     v && Number.isFinite(v.x) && Number.isFinite(v.y) ? null : "expected a finite point"],
 ];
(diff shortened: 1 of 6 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #807, opened 2026-08-12
Gate the heavy jobs' steps: skipping a matrix job renames it

hotsunchip:
## Summary

**#803 is stranding the merge queue right now.** Its two heavy jobs carry a job-level `if:` on the path filter. A job with a `strategy.matrix` that is skipped by its own `if:` **never expands the matrix**, so it files its check run under the *bare* job name. Measured on the `merge_group` SHA for #799:

```
verify-self (22.x)      success     ← ran, so the matrix expanded
verify-browser          skipped     ← BARE NAME
verify-integration      skipped     ← BARE NAME
```

`main` requires `verify-browser (22.x)` and `verify-integration (22.x)`. Neither name exists on that SHA, so the entry sits on *"Expected — Waiting for status to be reported"* until the status-check timeout dequeues it.

This fix makes both jobs always run and always expand, gating their **steps** instead.

## Why

The premise in #803 was half right. A skipped job *does* file a `skipped` check run, and that *does* satisfy a required check — but only if the name matches. Skipping a matrix job renames its check. MAINTAINING.md warns that "required check names must keep matching" about *renaming a job*; it turns out **skipping one renames it too**.

Not a merge-queue-only problem: a filtered ordinary PR's `verify-browser (22.x)` also never reports, so branch protection can never be satisfied there either. The queue is just where it costs an hour instead of a puzzled maintainer.

With the gate on the steps, a filtered run no-ops and reports success in seconds — the semantics the skip was reaching for, under a name that exists. Costs: one runner for a few seconds, plus Postgres starting for `verify-integration`, because `services:` come up before the first step and cannot be made conditional. The Yorkie container is started by a step, so it is gated with everything else.

The one remaining job-level `if:` is `!cancelled() && needs.verify-self.result == 'success'`, and it is deliberate: it can only be false when `verify-self (22.x)` has already reported a failure, so the entry is dequeued rather than stranded. Removing it entirely would be worse — a `changes` failure alongside a passing `verify-self` would skip these jobs via `needs:` and strand the queue again.

## Linked Issues

Fixes the regression introduced by #803.

## Author checklist

## Verification

**This PR touches `.github/workflows/**`, which is in `ci.ciConfig`, so it forces `full: true` on itself — every lane and both heavy jobs must run here.** That is the self-guard working, and it also means this PR cannot demonstrate the *filtered* path. See *Follow-up* for how to confirm the fix.

Locally: 77 scripts tests pass, `pnpm lint:scripts` clean, `pnpm verify:entropy` 0 issues, all four workflows parse and every embedded `github-script` body parses.

**The new guard was mutation-tested, not just written.** `scripts/test/ci-workflow.test.mjs` was run against the exact broken form this commit replaces:

- reintroducing the job-level `if: … outputs.heavy …` → fails with *"a skipped matrix job files its check under the BARE name, so the `(22.x)` context required by branch protection would never report. Gate the steps instead."*
- removing one step's `RUN_HEAVY` condition → fails with *"step `Build Docker image` has no RUN_HEAVY condition — it would run even when the filter says there is nothing to test."*

Both pass again once restored.

## Risk Assessment

- **User-facing risk:** none. CI-only.
- **Data/security risk:** none. No permission or token change; `ci.yml` stays `contents: read`.
- **Rollback plan:** revert this commit and the queue returns to the current broken state, so the safer rollback is reverting #803's two `if:` blocks instead (which restores unconditional heavy jobs — slower, but never stranded).

## Notes for Reviewers

- **AI assistance:** authored with Claude Code (Opus 5), including the root-cause diagnosis from the `merge_group` check-run names and the mutation testing of the new guard.
- **Read `scripts/test/ci-workflow.test.mjs` first.** It encodes the lesson mechanically, which is the only reason this cannot silently come back. It is deliberately narrow: it bans the *filter* decision from gating a matrix job, while still permitting an `if:` that can only be false on an already-doomed run.
- **Immediate unblock while this is in review:** the `full-ci` label will *not* help, because `merge_group` payloads carry no PR labels. Either dequeue #799 and merge it outside the queue, or temporarily un-require the two `(22.x)` contexts.
- **Follow-up:** confirming the filtered path needs a PR that touches only inert paths (e.g. `scripts/agent/**`). On that PR, `verify-browser (22.x)` should report **success in seconds** with a "had nothing to test" notice, rather than reporting `skipped` under a bare name.

## Summary by CodeRabbit

* **CI Improvements**
  * Merge-queue verification checks now remain available consistently, including for filtered changes.
  * Unaffected heavy browser and integration tests complete quickly without running unnecessary work.
  * Browser screenshots and diagnostic logs are collected only when relevant tests fail.
  * CI avoids running additional verification after cancellation or failed core checks.

* **Documentation**
  * Updated contributor and maintainer guidance for path-aware CI, full CI runs, merge queues, required checks, and workflow permissions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,10 +56,16 @@ jobs:
   # WHY THIS IS A JOB AND NOT `on: paths:`. A workflow filtered out at the trigger
   # produces no check run at all. `main`'s required contexts then never report,
   # and a merge-queue entry waits on them until the status-check timeout dequeues
-  # it. A job skipped by `if:` DOES report — as `skipped`, which GitHub counts as
-  # passing for both required checks and the queue. That difference is the whole
-  # reason this file is shaped this way; see
-  # docs/design/harness-engineering.md#path-aware-ci.
+  # it. A job that runs DOES report, under a name the queue recognises.
+  #
+  # AND NOT A SKIPPED JOB EITHER, which is what #803 tried. A skipped job files a
+  # `skipped` check run that would satisfy a required check — but a job with a
+  # `strategy.matrix` skipped by its own `if:` never expands the matrix, so the
+  # check is filed as `verify-browser` and not `verify-browser (22.x)`, which is
+  # the name actually required. The required context then never reports and the
+  # queue hangs exactly as if the check had been filtered away at the trigger.
+  # So the two heavy jobs below always run and gate their STEPS. See the long note
+  # on verify-browser and docs/design/harness-engineering.md#path-aware-ci.
   #
   # `fetch-depth: 0` is load-bearing. `actions/checkout` defaults to a depth-1
   # clone in which the base commit does not exist, so `git diff` fails — and the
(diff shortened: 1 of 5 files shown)


hotsunchip:
## The filtered path is now verified end to end

This PR cannot exercise the reduced path itself (it touches `.github/workflows/**`, so `ci.ciConfig` forces `full: true` on it). So I ran a disposable PR in a fork against a base carrying this fix, with a diff of exactly one comment line under `scripts/agent/` — the resolver returning `full=false heavy=false tags=[agent]`, which is the case that stranded #799.

**Every check run on that head SHA:**

```
Pipeline copy matches the pinned commit   success
What changed                              success
verify-browser (22.x)                     success     <- REQUIRED NAME
verify-integration (22.x)                 success     <- REQUIRED NAME
verify-self (22.x)                         success
```

There is **no bare `verify-browser` or `verify-integration` check run at all.** Compare the same query on #799's `merge_group` SHA before the fix:

```
verify-self (22.x)      success
verify-browser          skipped     <- BARE NAME, and the (22.x) context reported nothing
verify-integration      skipped     <- BARE NAME
```

**And the speed win survived the fix:**

| Check | Full run (#807) | Filtered run |
| --- | --- | --- |
| `verify-self (22.x)` | 9m21s | **49s** |
| `verify-browser (22.x)` | 8m44s | **4s** |
| `verify-integration (22.x)` | 2m42s | **27s** |

~1.5 minutes wall clock against ~15. The 27s on `verify-integration` is the Postgres `services:` container starting — the one cost that cannot be gated away, exactly as the code comment predicts; every step in that job no-ops.

Both throwaway branches and the fork PR are deleted.

### On the framing in the original description

A maintainer correctly pointed out that "the queue hangs" overstates it, and the distinction is worth recording because the two paths differ:

| | Behaviour | Bound |
| --- | --- | --- |
| Merge queue | the skip succeeds; `verify-browser (22.x)` never reports, so the entry sits on "Expected — Waiting for status to be reported" | **bounded** — dequeued at the status-check timeout |
| Ordinary PR | identical, but branch protection has no timeout | **unbounded** — stays unmergeable until someone intervenes |

The job was always skipping correctly. What never arrived was a report under the *required name*. And the ordinary-PR case is the worse of the two, not the queue.

dlgpdmsly2 (approved):

hackerwins (approved):
Code-owner approval (`.github/workflows/` + `docs/design/harness-engineering.md`).

Verified the load-bearing claim against the branch rather than the prose:

- `verify-self` carries `if: ${{ !cancelled() }}` with `needs: changes`, so it runs and expands its matrix even when `changes` fails. `RUN_HEAVY`'s leading `needs.changes.result != 'success'` then forces the full suite — the fail-safe direction is preserved.
- The heavy jobs' only skip path is `verify-self` *failing*, where `verify-self (22.x)` reports that failure and the entry is dequeued rather than left waiting on a name that never arrives. That is the one case the bare-name rename is harmless in, and the comment says so.
- `env` is a valid context in step-level `if:`, and job-level `env:` may read `needs` — so `RUN_HEAVY` resolves where it is used.
- `scripts/test/ci-workflow.test.mjs` pins the invariant, so a revert to the shorter job-level `if:` fails a test instead of stranding a queue entry an hour later.

On the two open threads, neither blocks:

- The `full-ci` label cannot fire on `merge_group` — correct, but pre-existing and already written down in the comment this PR retains. On a queue run `needs.changes` still governs, so the escape hatch is inert there rather than wrong.
- The MD040 fence language is a docs nit; worth folding into whatever next touches that file.

PR #808, opened 2026-08-12
Segment the volume and mix metrics by file class, severity and diff size

dlgpdmsly2:
## Summary

Adds `scripts/agent/eval/segmentation.mjs`: the five metrics `eval/volume-mix.mjs` already computes, recomputed over subsets of the same records — by file class, severity, diff size, authorship, code novelty and (for CodeRabbit) its own category. Pure functions plus a CLI that reads a store, computes and prints; two new files plus a task doc, nothing existing touched.

No new metric is defined. What is new is the slicing, plus the two rules a slice needs: a cell whose denominator is below `min_n` is **withheld** with the `n` and `min_n` that decided it and carries no value at all, and every proportion carries a **Wilson 95% interval** — the first confidence interval anywhere under `scripts/agent/`.

## Why

`volume-mix.mjs` reports one row per arm, averaged over every finding at once. *"This reviewer localises 80% of its findings"* is a weaker claim than *"82% in `code` files and 40% in `docs/design`"*, and only the second tells anyone where to look. Cutting the existing numbers is the cheapest way to get there, and it needs no new data.

**The reason this is a whole module rather than a `groupBy` is that a segmented average fails silently.** Cutting a small population six ways produces cells holding one or two observations, and a ratio over n=1 prints exactly like a ratio over n=100. Two guards carry that:

- **Suppression is policy, not formatting.** A withheld cell has no `value`, no numerator and no interval — the keys are absent, so the number is not in the output for anyone to quote. On the data this was measured against, **73 of 149 cells are withheld**, and that is the correct answer rather than a shortfall.
- **The textbook confidence interval is wrong in a specific direction at exactly the cells a small grid is made of.** At `k=0` it collapses to `[0,0]` — "we saw none, therefore there are none", from a sample of any size — and at `k=n` it claims certainty the same way. Wilson gives `[0, z²/(n+z²)]` and `[n/(n+z²), 1]`. All four edges (`k=0`, `k=n`, `n=0`, `n=1`) are pinned to their exact closed forms by tests, because a test that only exercises `k=3, n=10` passes against the formula this module exists not to use. `n=0` returns no interval at all, with the reason.

**Two things it deliberately refuses to average.** A metric counted in pull requests is only ever cut by an axis that segments pull requests — *"median minor findings per PR"* keeps every item in its denominator whatever the severity bucket says, which is a numerator filter wearing a segment's label. And a metric its axis determines is skipped as a tautology: the nit ratio within `severity=nit` is 1.000 by construction, and so is the nit ratio within `novelty=not-annotated`, because `annotateFindings` stamps novelty only on critical and major findings. The second one is the dangerous shape — it reported `1.000 (112/112)` with a tight interval, a tautology with a fat denominator — and it was found by running the grid and reading it.

**Three absences are kept out of the zeros they resemble.** A finding citing no file goes to `no-file` rather than to `code`, which is what `classifyFile("")` answers and is the right fail-safe for the panel and the wrong bucket here. A finding never given a novelty origin goes to `not-annotated` rather than `unknown`, which is a real origin meaning git looked and could not tell. And a bucket where one observation set found nothing is described as that, with the per-set counts in the message, rather than as empty.

Two of the axis vocabularies are read from the code that owns them rather than assumed: file class is the five `FILE_CLASSES`, and authorship is the three values `classifyProvenance` returns. Diff size has two colliding scales, so the one used — `scopeSize`, cross-checked against the `meta.scope` frozen beside the numbers it derives from — is part of the axis id and therefore of every segment label.

Defect type is **declared and not computed**, with the reason: it needs labels that do not exist, and an omitted axis and an unbuildable one look identical in a table.

## Linked Issues

None. It follows `eval/volume-mix.mjs` (#773) and sits beside `eval/reliability.mjs` (#782) as a third read-only scorer over the same records.

## Author checklist

## Verification

Measured locally, from the committed tree rather than the working copy:

- **`agent:tests`, both invocations: 1792 tests (1736 + 56), 1791 pass, 0 fail, 1 skipped**, against a freshly measured `main` at **1755 tests (1699 + 56), 1754 pass, 0 fail, 1 skipped**. **+37**, which is exactly this PR's test count. Both trees were extracted with `git archive` and given identical `node_modules` before either was measured, so the one skip is a property of both trees rather than an environment artefact.
- **`eslint scripts` exits 0** at the lockfile-pinned 9.24.0, on both trees.
- **Mutation testing: 54 mutations, 54 caught, 0 survivors** — each by the test that claims to prevent it, checked by test name rather than by the suite going red. Three survived a first pass; all three turned out to be ineffective for the inputs the test used rather than uncaught, and all three are written up in the task doc. One is worth the line: removing the `[0,1]` clamp from the Wilson bounds changes nothing at n ∈ {1,5,10,30,142} and lands 7e-18 below zero at **n=27**, so 27 is now in the loop.
- **Run end to end against a real store**: 149 cells over three observation sets, 76 reported and 73 withheld, in about 4 seconds. The command is in the task doc and costs nothing.

Skip reason (if applicable): none — both lanes apply.

## Risk Assessment

- **User-facing risk: none, and this is the case where "nothing reads it" is the honest answer rather than a dodge.** Two new files, no existing file touched, no export changed, nothing imports them. Not in the gating path: `review-panel.mjs`, `severity.mjs` and every lens are untouched, and this module only ever reads. It does import `classifyFile` and `FILE_CLASSES` from `review-panel.mjs` and `scopeSize` from `metrics.mjs` — reads of exported pure functions, with no call in the other direction.
- **Data/security risk: none. It writes nothing** — no `--out`, no store write, no network beyond the read-only GitHub calls the CodeRabbit reader already makes, no model call and no spend. `--root` is required with no default anywhere, so it cannot fall back to a path inside this repository.
- **Rollback plan:** revert. Nothing consumes it, so nothing else moves.

## Notes for Reviewers

- UI changes (screenshots/gifs if applicable): none — this is a CLI that prints to stderr.
- **AI assistance:** written with Claude Code — the module, its tests, the task doc and this body. Every upstream claim it rests on was re-checked at `file:line` on `main` before use, and the measured numbers above come from runs on this branch rather than from a document.
- **Where to look first:** `cellFrom` and `wilson`. Between them they decide whether a number is published at all, and the invariant in `cellFrom` — that a cell's value, numerator and denominator all come from one observation set — exists because the first draft published `0.833 (25/17)`, a ratio larger than one, with nothing red.
- **The console report prints every cell including the withheld ones**, deliberately. On a small population the withheld rows are the result, and their count is the headline.
- **A caveat the output makes visible and this module does not resolve:** two of the five metrics — the localisation rate and the in-diff rate — come out at exactly 1.000 in every reported CodeRabbit cell, because an inline review comment is anchored to a diff line by construction. So on those two metrics a cross-arm reading compares posting mechanisms rather than reviewer discipline, and whoever writes prose from this grid needs to say so. Deciding how to present that is not a scorer's call, which is why it is a note here rather than a filter in the code.
- **Follow-up work:** nothing reads a `segmentation-v1` payload yet, so the payload shape is asserted against its own tests here. Filing it into a store and rendering it are separate changes.

## Summary by CodeRabbit

- **New Features**
  - Added segmented evaluation across file class, severity, size, provenance, novelty, category, and review window.
  - Added configurable scoring and reporting with per-segment comparisons, completeness details, and suppression metadata.
  - Added support for minimum-sample suppression, replicate aggregation, and statistically appropriate confidence intervals.
  - Added validation for incompatible metrics, incomplete data, and inconsistent evaluation configurations.

- **Documentation**
  - Documented the evaluation approach, validation rules, pilot measurements, known corrections, and verification results.

- **Tests**
  - Added comprehensive coverage for scoring, segmentation, suppression, validation, reporting, and edge cases.

### DIFF
--- a/docs/tasks/active/20260812-eval-segmentation-todo.md
+++ b/docs/tasks/active/20260812-eval-segmentation-todo.md
@@ -150,7 +150,7 @@ intended guard produces; nothing was weakened.
 
 ## The numbers — the pilot's grid, measured once with the code frozen
 
-```
+```sh
 node eval/segmentation.mjs --root <store> --corpus-version 2026-08-10-pilot-reviewed \
   --run-id pilot-01__k1 --run-id pilot-01__k2 --run-id pilot-01__k3
 ```
@@ -222,13 +222,13 @@ A sample of what survives both arms' suppression tests:
 
 ## Verification
 
-- [x] `agent:tests`, the real CI lane (two invocations since #774) — **1792 tests (1736
-      + 56), 1791 pass, 0 fail, 1 skipped**, against a freshly measured `main`
-      (`49b51885a`) at **1755 (1699 + 56) tests, 1754 pass, 0 fail, 1 skipped**. **+37**,
+- [x] `agent:tests`, the real CI lane (two invocations since #774) — **1797 tests (1741
+      + 56), 1796 pass, 0 fail, 1 skipped**, against a freshly measured `main`
+      (`49b51885a`) at **1755 (1699 + 56) tests, 1754 pass, 0 fail, 1 skipped**. **+42**,
       which is exactly this file's test count. Both trees were extracted with `git
       archive` and given identical `node_modules` symlinks **before either was measured**,
       so the single skip (the Agent SDK) is a property of both trees rather than an
-      environment artefact.
+      environment artefact. *(Was 1792 / +37 before the review round below.)*
 
       ⚠ **`eval/run.test.mjs` failed 2–3 rotating tests mid-session on BOTH trees, and it
       is not this change.** Another session was running its own mutation harness over the
@@ -239,10 +239,12 @@ A sample of what survives both arms' suppression tests:
       deregisters the worktree`, `the reaper is WIRED UP`), and clean again — 56/56 — once
       the other harness stopped. Not investigated further and not touched.
 - [x] `eslint scripts` exits **0** at the lockfile-pinned 9.24.0, on both trees.
-- [x] **Mutation testing: 54 mutations, 54 caught, 0 survivors** — and each by the test
-      that *claims* to prevent it, checked by test name rather than by the suite going
-      red. Three survived the first pass; all three were ineffective rather than
-      uncaught, and all three are written up above.
+- [x] **Mutation testing: 67 mutations, 66 caught, 1 ineffective** — 54 in the first
+      pass and 13 more over the review round's code, each caught by the test that
+      *claims* to prevent it, checked by test name rather than by the suite going red.
+      All 54 originals were re-run against the reworked module and still apply. Four
+      survived a first pass across the two rounds; all four were **ineffective rather
+      than uncaught**, and all four are written up above and below.
 - [x] Verified from the **committed tree** (`git archive <branch> | tar -x`), not the
       working copy; both new files are byte-identical between the two.
 - [x] Measured against the real store **once, with the code frozen**, and the whole run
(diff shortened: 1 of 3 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #817, opened 2026-08-13
Add the label record and an adjudication CLI for eval findings

dlgpdmsly2:
## Summary

Two new files under `scripts/agent/eval/`, and nothing existing is touched.

- **`labels.mjs`** — the record for one human reading of a diff: a pinned
  `SCHEMA_VERSION`, frozen vocabularies, `buildFindingLabel` / `buildItemLabel`, a strict
  `validateLabel`, `labelPathFor` and `labelCensus`. Same shape as `finding-record.mjs`.
- **`adjudicate.mjs`** — a CLI that presents a finding for judgement, collects the answer
  and writes it. `--root` is required with no default; it writes nothing without
  `--write`; it is resumable from the records already on disk.

Nothing consumes these records yet, and no reading is produced here.

## Why

`eval/` can replay a frozen pull request through the review panel and score what each
reviewer *produced* — volume, severity mix, complementarity, agreement across replicates,
cost. **None of those numbers says whether a finding is correct**, and none of them can:
correctness needs an independent reading of the diff, and the tree has nowhere to record
one.

**The thing this is shaped around is not the schema, it is a failure mode.** The cheap way
to collect these readings is to show the reader what the reviewer concluded and ask them
to agree — and that produces a correctness figure that is higher, publishable, and wrong.
Every other defect this subsystem has had printed evidence of itself somewhere; this one
does not. A scorer that disagrees with the panel's own artifact turns something red. A
reading taken from the panel's verdict agrees with itself.

So the guard is structural. `admitRecord` projects a finding record onto **nine named
fields** at the boundary: the reviewer's severity, its verifier outcome, its lane and its
gating verdict are not withheld further down — they never enter the pipeline, so nothing
downstream can print what nothing downstream has. An allowlist and not a denylist, because
a denylist has to be updated by whoever adds the next field to a record and they will not
know to; a test constructs a record carrying all four signals plus a field nothing has
added yet and requires that none reaches the presenter. `labels.mjs` shuts a second door:
what was on screen is a required field, and a record admitting the reviewer's severity was
on it is refused at write time. Two doors, because a validator only guards the door it
stands in.

⚠ **This deliberately narrows, against the convention the adapters follow.** That
convention protects the road to a scorer, where a dropped field is a dropped fact; this
projection is the road to a person, where a carried field is the circularity above.
Nothing is lost — the record stays in the store and a reading joins back by `finding_key`.

**The unit is where the design came from.** `findingKey` is `(file, lowercased summary)`,
and it is identity rather than similarity by design, so exact keys barely repeat across
replicates: over 428 stored records, **426 distinct keys, none in all three replicates,
two in exactly two**. The reuse is one unit up — `groupFindings` collapses the same 428
into **245 defect classes**, 127 of which hold more than one key. So a person judges a
class and the tool writes one record per member key, each naming the class it came from,
and the census reports **how many readings produced how many records**: 428 records from
245 readings are not 428 independent judgements, and a later agreement pass that assumed
otherwise would overstate its own ceiling.

Bundles never span reviewers, and that is part of the blinding rather than tidiness — a
bundle holding both reviewers' claims would tell the reader the other one agreed, which is
the prior that looks most like evidence and is also exactly what a shared hallucination
looks like. The triage orders are `coverage` (most records settled per reading: the first
twenty cover 60 of the pilot's records against 33 unordered), `locality` (the dominant
cost is reading the diff, not typing) and `none`. Five more are refused **by name** with
the leak each would produce — unreachable anyway, since a card has no severity to sort by,
but the list sits where someone would come to add one.

The item-level verdict rule is `BLOCKING` from `severity.mjs`, imported rather than
restated: the gate's own definition of what stops a merge is the one a recorded verdict
has to agree with, so `block` is refused with no blocking defect named and `approve` is
refused beside one.

## Linked Issues

**None.** It follows from the eval subsystem's own gap — every scorer under `eval/`
measures what a reviewer produced and none of them can say whether it was right.

## Author checklist

- [x] Changes follow the conventions in the touched packages; any deviations are explained
      in *Why* above — specifically the narrowing projection, which inverts the adapters'
      widening rule on purpose.

## Verification

Measured locally on both trees extracted separately with the same `node_modules`, each
measured once, the way the `agent:tests` lane runs it (two invocations since #774):

```
                       rest    iso   total
main                   1700  +  56  = 1756     0 fail · 1 skip
this branch            1765  +  56  = 1821     0 fail · 1 skip
                                       +65     = exactly this change's test count
```

The single skip is the Agent SDK case and is present in **both** trees. `eslint@9.24.0`
(the lockfile pin) exits **0 on both**. Verified from the committed tree rather than the
working copy: all five files byte-identical, same numbers.

**62 mutations · 61 caught by the test that names the guard** — matched by test name, not
by the suite reddening. The three that changed no behaviour are named in the task doc with
the proof; two were replaced with effective ones, and the third documents a tiebreak that
is unreachable today and kept anyway, so ordering stays a property of this module rather
than one inherited from another module's internal sort.

Exercised end to end against the real replay store, read-only: 428 records → 245 defect
classes, which reproduces the partition an existing scorer publishes (118 · 71 · 56 by
replicate span) down a separate code path.

Two unchecked boxes in the task doc, deliberately: the atomic-write property (a crash
between write and rename is not something `node --test` can stage honestly) and that the
projection holds for arbitrary reviewer *prose* — the word scan proves the four signals of
a constructed record do not survive, and cannot prove a reviewer never writes "major" in
its own evidence, which would be their word rather than a leak.

## Risk Assessment

- **User-facing risk: none, and rollback is a revert.** Two new files that nothing imports.
  No workflow, no check run, no gating path, no lens and no gate decision is touched — the
  gate's severity rules are *imported* (`KNOWN`, `BLOCKING`) and not redefined, so nothing
  here can change what blocks a pull request. `severity.test.mjs` and
  `finding-record.test.mjs` pass untouched inside the 1765.
- **Data/security risk: the write path, and it is the reason `--root` is required with no
  default.** git history is permanent, so one forgotten flag falling back to a path inside
  this repository would commit data here for good and no later `git rm` shrinks anyone's
  clone — `capture-store.mjs` established the rule. On top of that: the CLI writes nothing
  at all without `--write`; every path is built from validated path segments and asserted
  to be inside the root; writes go through a temp file and a rename, so a destination is
  only ever absent or complete; an existing record is refused rather than overwritten
  unless asked; and a record whose diff hash disagrees with the item's current one is
  refused rather than stored, because a reading of a diff that no longer exists describes
  the wrong code silently.
- **Rollback plan:** revert. Nothing reads these modules, and anything they have written
  lives outside this repository.

## Notes for Reviewers

- **AI assistance:** written with Claude Code (Claude Opus 5) — the modules, the tests, the
  mutation testing and the task doc. Every measured figure in this body was produced by
  running the command, and the two defects the task doc records under *Corrected while
  building* were both found by measurement rather than by review; one of them silently
  disabled all bundling and reported a number indistinguishable from a correct one.
- **Follow-up work.** Nothing consumes these records. The scorer that computes precision,
  relative recall and the false-positive profile from them is a later change and must not
  be written before records exist — one built against an empty set would be tested only on
  fixtures. The first real readings will be produced by a person using this CLI.
- **The governance doc for these records lives outside this repository** and names two
  store methods that were never built; the schema and the drift check therefore live in
  `labels.mjs`, on `finding-record.mjs`'s precedent, and `store.mjs` is untouched.
- **`eval/README.md` gets no row on purpose** — several changes in flight would each want
  to edit that one file. Happy to add it if you would rather have it here.

## Summary by CodeRabbit

- **New Features**
  - Added a blinded, resumable command-line workflow for reviewing evaluation findings.
  - Supports grouped review queues, optional diff inspection, preview mode, relabel protection, and JSON output.
  - Added structured label creation, validation, secure storage, drift detection, and review census reporting.

- **Documentation**
  - Documented the workflow, label schema, operational rules, verification results, and limitations.

- **Tests**
  - Added comprehensive coverage for blinding, queueing, resumability, validation, persistence, prompting, and census behavior.


dlgpdmsly2:
All five fixed in `637226ef`, plus the nitpick.

- **`fileDiffSection`** — both `diff --git` paths are now compared whole, so `a.ts` no longer matches `data.ts`. Renames resolve from either side. A C-quoted header (the form `extract-corpus.mjs` documents for non-ASCII paths) no longer matches and falls back to "open `diff.patch` directly", which is the right direction here: showing the wrong file is worse than showing none.
- **Line range** — re-asked until it parses, reversed ranges refused too. This was the most valuable of the five: an unparsable range and a blank one both produced `line_range: null`, and the annotation guide treats "not line-localizable" as a real statement about the defect — so a location the annotator typed silently became an affirmative claim that there wasn't one.
- **Malformed label** — a file that parses but carries no `finding_key`/`arm` now joins the truncated-JSON case in `unreadable`. It previously counted as a label while settling nothing, so the bundle returned to the queue, got answered a second time, and the write path then refused because a file already existed there — a dead end with no line explaining it.
- **`--annotator`** — now required for any session that will ask a question, not just a writing one. `--json` is exempt since it prints the queue and returns.
- **`withRoot`** — awaits the body before cleanup.
- **Nitpick** — `ADJUDICATION_MODES` and `LABEL_SOURCES` are imported from the schema module, and the tiers this CLI can produce are derived by excluding `distant` rather than by listing the other two.

One correction on `withRoot`. Consequence (1) is right, and I measured it: **3 leaked directories per run with label files in them, 0 after the fix.** Consequence (2) is not. `writeFileAtomic` calls `mkdirSync(..., { recursive: true })`, so a stray write *recreates* the tree and the `existsSync` check still catches it — forcing a write in preview mode turns that test red under both the old and the new helper. The absence assertions were sound; the leak was the whole cost. I'd assumed you were right about this one until I tried to prove it.

Verification: **1768 + 56 tests, 0 fail, 1 skip** against a freshly re-measured 1700 + 56 on `main` (+68); eslint 0; **68 mutations, 67 caught by the test that names the guard** — the survivor is a tiebreak that is unreachable today and documented as kept deliberately.

dlgpdmsly2:
Both fixed in `dd7df198`, and the first one is worth more than the flag it names.

**`--mode model --label-source gold`** was accepted at startup and refused per judgement — measured on one bundle: **7 questions asked, 0 labels built**. That is the same defect as the `--annotator` one from the last round, one flag over, and I fixed that instance without asking what else had the shape. The cause is that `resolveOptions` was restating rules `validateLabel` owns, so any rule it missed cost a reader their answers.

So rather than adding the one condition, the preflight now **asks the schema**: it builds a throwaway label from the resolved mode/tier/annotator with known-good constants elsewhere, and reports whatever `buildFindingLabel` says. Never written. That immediately caught two combinations nobody had reported — a human id under `--mode model`, and a model id under `--mode human`, which are the laundering the annotation guide warns about in both directions. It runs last and only on an otherwise-clean invocation, so flag-level messages still name the flag and one mistake prints once. It fails closed if the schema later demands a field the probe lacks; loudly unusable beats quietly discarding an evening.

**The readFindingLabels one** needed a code change, not just a test — the behaviour it asks about didn't exist. `readFindingLabels` now runs `validateLabel`, so a label that parses with a real `finding_key` and `arm` but that the schema refuses stops counting as a resume key. Previously resume called it judged, a validating scorer dropped it, and nothing asked again. An item label misfiled under `findings/` is rejected there too. Two boundaries are deliberate and written at the call site: **drift is not checked here** (a stale label is a well-formed judgement about a re-extracted diff — a different fact, already refused at the write path, and checking it here would return a whole corpus to the queue on re-freeze), and **a `SCHEMA_VERSION` bump is a migration, never a re-adjudication** — human reading is the scarcest thing this module spends.

hotsunchip (approved):

PR #824, opened 2026-08-13
Name the test that actually failed in a lane's failure summary

dlgpdmsly2:
## Summary

`extractFailureSummary` now reads the structure the tool already emits — TAP,
then `node --test`'s spec reporter — instead of grepping the whole lane output
for the first line containing "error". Moved to `scripts/failure-summary.mjs` so
it can be tested, and covered by 16 tests in `scripts/test/`.

The non-TAP fallback keeps its heuristic, with two corrections: log-level lines
are skipped, and the `✗`/`✘`/`✖` marks now match — inside `\b` they never could,
so a tick-marked failure had always fallen through to the lane's last line.

## Why

The old rule was: return the first line matching
`/\b(FAIL|ERROR|error|Error|✗|✘|FAILED)\b/`. That is right only by accident, and
two real runs show both failure modes:

- **#758's `agent:tests`** reported `classifyResult: distinguishes verdict /
  api-error / no-output at its new home` — a test that **passed** (`ok 29`),
  matched on `api-error` inside its own name, 2,600 lines before the real
  failure. A suite that tests error handling always has "error" in some passing
  test's name, so **every `agent:tests` failure since #578 carried a wrong
  summary**.
- **Run 31658161586's `verify:entropy`** reported a NestJS `WARN` line — a
  fixture deliberately logging a handled error, ~700 lines before anything
  failed.

This matters more since #803: `ci-report.yml` renders `failureSummary` into the
job summary, so the wrong name is now read by everyone who opens a failing run,
not just whoever downloads the artifact.

**Both reporters are parsed, because node picks between them itself** — `tap`
when stdout is a pipe, `spec` on a TTY, and the default has moved between
majors. CI captures lanes through a pipe and sees TAP; the same command on a
laptop prints spec. Not hypothetical: the end-to-end check for this change was
written against CI's format, passed, then printed the *passing* test's name when
run locally, because the spec marker is `✖` U+2716 and the old pattern only knew
`✗` U+2717 and `✘` U+2718.

The fallback stays a heuristic on purpose. The honest fix for vitest, tsc and
knip is to read each tool's own format, and inventing that without a real failing
sample per tool is how a reporting change starts reporting something new and
equally wrong.

## Linked Issues

None. Follows from #803, which added the `scripts:tests` lane this is tested by
and the job summary that surfaces the value.

## Author checklist

## Verification

Skip reason (if applicable): n/a — no integration surface is touched.

- **16 tests**, every fixture lifted from a real run rather than invented.
  **7 of them fail against the previous implementation**; the 2 that pass are the
  pins that non-TAP behaviour is unchanged.
- **14 of 14 mutations caught.** Three survived a first pass and are now covered:
  an unbounded YAML search (which would attach a later failure's reason to an
  earlier name — reachable here, since the deserialize failure truncates TAP
  mid-block), an unanchored `not ok` that would match the phrase inside a test's
  *name*, and a spec pattern missing its duration requirement, which would report
  the literal string `failing tests`. Four more cover the marker alternation,
  including reverting it to the original `\b(…|✗|✘|✖|…)\b` form.

- **A review finding corrected two things here.** `\b` cannot hold before `✗`
  (U+2717 is not a word character), so all three marks were dead — two of them in
  the original pattern, meaning a vitest failure line has never been selected by
  this scan. And the test meant to cover it passed for the wrong reason: its
  failure line was also the LAST line, so the assertion held via the fallback
  rather than the scan. It now carries a non-matching footer, which is what tells
  the two apart.
- **End to end through `verify-self.mjs`**, reading the value back out of
  `.harness-reports/`, under both reporters:

  ```
  spec -> the genuinely broken thing — AssertionError [ERR_ASSERTION]: the real reason
  tap  -> the genuinely broken thing — the real reason
  ```

- `npx eslint scripts` exits 0. `scripts:tests` is 129 tests, 0 failures.

## Risk Assessment

- **User-facing risk:** none. Reporting only — the value is written to
  `.harness-reports/<lane>.json` and rendered in the job summary. No lane's exit
  status, selection or output changes.
- **Data/security risk:** the summary is truncated to 500 characters as before,
  so nothing new is exposed. The one behavioural risk is misreporting, which is
  the defect being fixed; the fallback path is byte-identical to today except for
  skipping lines a log level marks, and that is pinned by a test.
- **Rollback plan:** revert. The module is imported from exactly one place and
  nothing else consumes it.

## Notes for Reviewers

- **AI assistance:** written with Claude Code (Claude Opus 5), including the
  mutation testing and the end-to-end runs. Every quoted output came from a
  command that was run.
- **Deliberately not done:** per-tool parsing for vitest / tsc / knip. Those
  lanes still use the heuristic, and `verify:entropy`'s WARN case above is the
  evidence that it remains imprecise for them — worth its own PR once there are
  real failing samples to write against.


dlgpdmsly2:
@coderabbit review

harrykim8672 (approved):

PR #827, opened 2026-08-13
Consume the CodeRabbit adapter's latency read in the cost scorer

dlgpdmsly2:
## Summary

`eval/cost-latency.mjs` now consumes the timing records `eval/adapters/coderabbit.mjs`
already produces, instead of declaring CodeRabbit's latency as a gap.

- `coderabbitLatency()` takes the adapter's per-item records and emits two figures, each
  naming its own interval and carrying its own `n`.
- The `coderabbit_latency_ms` declared gap is retired when records are supplied and kept,
  with a reason naming the flag, when they are not. `cost_per_real_finding` is unchanged.
- `PANEL_INTERVAL` names our own interval in a field rather than a comment.
- `panel.review_wall_ms` / `panel.review_cost_usd` pool over observations (n=21) beside
  the existing per-replicate spend series (n=3).
- `fitCostToSize` returns its own minimum as `min_n` on every path.
- `--coderabbit-latency` is a new, opt-in CLI flag; it is the only network read in the file.
- `SCHEMA_VERSION` → 2.

## Why

The scorer declared `coderabbit_latency_ms` as *measurable, and not from anything this
scorer reads*, and named the shape the measurement would arrive in. The adapter then built
that read — with the author gate, the in-window rule and twelve declared absences — and
nothing consumed it. **A gap that outlives the read it asked for is a gap nobody believes:**
the next reader either re-derives the number or substitutes a worse one, and the substitute
this field used to propose is wrong by 8×.

Passing the adapter's **records** rather than a number is what keeps those guards attached
to the figure. Two intervals come out, because they answer different questions:

| | interval | pooled over | pilot |
|---|---|---|---|
| primary | `coderabbit-start-marker-to-first-finding` | every item | **2.6–14.4 min, median 6.8, n=7/7** |
| secondary | `earliest-check-run-start-to-first-finding` | automatic triggers only | 2.8–14.8 min, median 6.7, n=5/7 |

The secondary comes off *our* CI's clock rather than CodeRabbit's, so agreement between
them is evidence and disagreement is a question. It is pooled over automatic triggers only
because where a human asked for the review it times the human's delay in asking: one pilot
item reads **183.7 min** from the push and **7.8 min** from CodeRabbit's own
acknowledgement. A second item is the same event reading an entirely ordinary 9.8 min, so
magnitude cannot separate them — the trigger is read, and the excluded number is kept in
the record rather than dropped.

**The harder half is what happens once both arms carry minutes.** Ours is a panel
*process's* elapsed time on an offline replay: it starts after the lane materialised a
worktree and it queues for nothing. Theirs is a production reviewer measured end to end.
Where both ran in production from one trigger, ours took about **2.2× longer** — the
opposite of the direction this was assumed to err in. So our interval is now a field, the
two arms stay under separate keys in separate blocks, and no ratio is computed anywhere. A
test asserts the payload holds no quotient key and that the report prints the warning on
the same lines as the numbers.

The two smaller additions serve the same discipline. The same corpus reads three
denominators — 3 replicates, 7 items, 21 observations — and a consumer that wanted one
pooled figure with an `n` previously had to build it and pick one itself. And a consumer
rendering "withheld for a thin denominator" needs both the `n` there was and the `n`
wanted; parsing the second out of an English `reason` is how a caption comes to contradict
the refusal it captions.

The flag is opt-in because the default invocation must stay offline and reproducible
against a committed store, and because a scorer that touched the network on every call
could not be re-run over stored data. With the flag absent the figure is a declared gap
naming the flag, never a blank.

## Linked Issues

None. It completes the interlock #791 set up and #799 built; neither has an open issue.

## Author checklist

## Verification

Measured locally from the committed tree, against a freshly measured baseline on the same
`main` (`1902133bf`), both trees set up identically:

| | rest | `eval/run.test.mjs` | total | fail | skipped |
|---|---|---|---|---|---|
| `main` `1902133bf` | 1857 | 56 | 1913 | 0 | 0 |
| this branch | 1865 | 56 | **1921** | 0 | 0 |

`npx eslint scripts` exits 0 (lockfile-pinned `eslint@9.24.0`).

Against the real store, all three replicates, `--coderabbit-latency`:

```
latency [coderabbit-start-marker-to-first-finding]: 2.6m–14.4m (median 6.8m, n=7)
  — 7 of 7 item(s) poolable, 7 measured
latency [earliest-check-run-start-to-first-finding]: 2.8m–14.8m (median 6.7m, n=5)
  — 5 of 7 item(s) poolable, 7 measured
triggers: automatic=5 · on-demand=2 · unknown=0
declared gaps: cost_per_real_finding only
```

Exit 0, `COMPLETE`. Our own pooled figure reads median 9.3 min over n=21.

**11 mutations, 11 caught**, each by the test that should catch it — including pooling the
proxy regardless of trigger, counting a missing latency as zero, retiring the wrong gap,
and emitting a cross-arm ratio.

⚠ `eval/run.test.mjs` is flaky in this local environment when another session runs
concurrently, on the two `END TO END` tests that snapshot `os.tmpdir()`. It reproduces on
`main` **more** often than on this branch (4/5 runs against 3/5), and this diff touches
neither `run.mjs` nor `run.test.mjs`. Pre-existing; not fixed here.

## Risk Assessment

- **User-facing risk:** none. Nothing in the gating path is touched — this module is a
  read-only scorer that no workflow invokes and the panel never imports. The one new
  behaviour a user can reach is a CLI flag that is off by default.
- **Data/security risk:** the new flag makes `gh api` reads through the arm's adapter. They
  are reads, they spend nothing, and the adapter already refuses rather than degrading when
  `gh` cannot resolve the repository — a corpus-wide absence would otherwise be
  indistinguishable from a clean review. Nothing is written and no model is called.
- **Rollback plan:** revert. `SCHEMA_VERSION` moves 1 → 2 because
  `coderabbit.latency.wall_ms` changed from a null scalar to a block; a score file written
  by this version renders correctly on the previous renderer, which reads neither field.

## Notes for Reviewers

- **AI assistance:** written with Claude Code. Every figure quoted above was printed by the
  code against the real store rather than transcribed, and the tests were mutation-tested.
- **Follow-up:** this changes a scorer, not a stored score. The published report does not
  move until somebody re-runs the scorers and re-renders.
- Deliberately **not** included: an agreement statistic between the two anchors (they agree
  to within 0.1–0.4 min on the automatic items, and that is a new number).


dlgpdmsly2:
@claude review

dlgpdmsly2:
Both defects in this PR's own code are fixed in `d596ea4d`. The third is real and I've left it — reasoning below.

**`requested` read as evidence of a measurement.** Confirmed by reproduction: with `latency: []`, or entries carrying no `latency`, the gap was retired with no figure behind it, and the completeness guard compared `0 < 0` so pushed no reason either — the run exited 0 reporting `complete`. `measured` is now its own field meaning a pooled figure exists (`self_timed.ms.n > 0`), the completeness guard checks it before the count comparison, and the absence carries three reasons instead of one (not requested / records carried nothing / nothing poolable), because they point at different fixes and the not-requested wording would tell someone to pass a flag they already passed.

**Duplicated `latencyAbsentLine`.** Correct — it was the function body inline, while importing `LATENCY_ABSENT` from the module that exports the function. Now imported and called. The test asserts through an unrecognised flavour, which is what a copy loses first.

**Unvalidated `source_pr` / `review_commit` in the API paths.** Real, and deliberately not fixed here. This PR touches neither `adapters/coderabbit.mjs` nor `gh-checks.mjs`, and `corpusRecords` is already reached on `main` by `segmentation.mjs:1352`, `complementarity.mjs:939` and the adapter's own CLI (fetches at `:1267`, validates at `:1273`) — so this adds a fourth caller of an existing API rather than new exposure. Fixing it here would cover one of four callers and put an unrelated repair inside this diff. It wants its own PR validating inside `fetchCodeRabbitPr` before the first API call: `^\d+$` on the pull request number, `^[0-9a-f]{7,40}$` on the sha.

Verification: **1868 + 56 = 1924** tests, 0 fail, 0 skipped, against **1913** freshly measured on `main` `1902133bf`; `eslint scripts` exits 0. **17 mutations, 17 caught** — two of them re-introduce the defects above and both go red. The real-store run is unchanged: `median 6.8m, n=7`, exit 0, and the census line is byte-identical now that it comes from the adapter's formatter.

harrykim8672 (approved):

PR #828, opened 2026-08-13
Render the cost and latency section of the comparison report

dlgpdmsly2:
## Summary

`eval/report.mjs` now renders section 4 from the cost/latency score payload instead of
printing a single "not computed" line over it.

- Section 4 becomes two blocks — our panel, then CodeRabbit — that share no table, plus a
  cross-arm row that is `not measurable` **permanently**.
- All four availability states appear where they belong, including a measured **zero**
  rendered as `present` and a scorer-withheld fit rendered as `suppressed` with both `n`s.
- Section 6 gains the n=2 production-latency pair that bounds section 4's minutes.
- Section 2's adjudication-budget sentence is corrected: it conflated the band's two bounds.

## Why

**Section 4 was the report's one blind section.** `costLatencyFigures` returned the
cross-arm cell and a sorted list of the payload's key names; its docblock said the section
would grow when the field shapes settled. They settled and it did not, so the report has
been printing *"nobody computed it"* over a 32 KB score file sitting unread in `scores/`.

**The layout is the argument here, more than anywhere else in this report.** Every instinct
a reader brings to this section is one the data cannot support: they want a cost per review
to compare, and there is no per-review price for a flat subscription; they want a latency
ratio, and the two clocks time different things. Put the two latencies in one table and the
reader does the division themselves — so they are never in one table, never on one axis,
and every figure names the interval it was measured over rather than carrying the bare word
"latency". The cross-arm cell is `not measurable` rather than `not computed` because no
re-run closes it: it is a result, not a hole waiting on a scorer.

**Two states that are easy to conflate now render distinctly.** A measured zero is
`present` — *"0 of 21 replays had no wall clock"* is a fact about the store, and a blank
cell in that position says nothing about how many were looked at. A fit the scorer withheld
for too few points is `suppressed`, carrying both the `n` it had and the minimum it wanted;
both come from the payload, because a default here would caption the scorer's refusal with
a threshold it no longer uses.

**Section 6 carries the measurement that does not flatter us.** Two corpus items hold our
own review check runs on the frozen commit, so on those the panel ran in production from
the same trigger the other arm's second anchor uses: **ours 18.7 and 19.0 min against
theirs 8.0 and 8.6.** Section 4's replay figures therefore understate us by roughly that
factor. It stays in the limits with its `n` rather than becoming a headline over two data
points, and — like the self-review caveat — it is intersected with the corpus being
rendered rather than asserted over it.

**Section 2 carried a sentence that understated the adjudication work by an order of
magnitude:** *"the queue is 409 undecided pairs, of which 15 score ≥ 0.7 — so adjudicating
this costs tens of pairs rather than hundreds."* That conflates the band's two bounds. The
threshold head moves the **floor**, one pair at a time. The **ceiling** does not move until
a finding has every one of its pairs decided, because until the last one is settled it
could still turn out shared — so its budget is the queue attached to the undecided
findings, which is hundreds of decisions. Both numbers come from the payload and each is
now labelled with the bound it buys. The exact ceiling budget is smaller than the whole
queue, because pairs owned by an already-shared finding cannot move either bound; that
deduction needs a per-finding pair count no scorer emits, so the section names it as **not
stated** rather than approximating it.

Two tests exist because mutation testing found the code passing for the wrong reason. A
per-line check that no rendered line carried both intervals stayed green when the two were
put in adjacent **rows of one table**, which is exactly the layout that invites the
division; it now works per table and rejects a shared pair of rendered minutes as well.
And the suppressed threshold was asserted against a fixture whose value matched the
hard-coded one, so hard-coding it passed.

## Linked Issues

None. It fills in a section the renderer shipped as deliberately empty.

## Author checklist

## Verification

Measured locally from the committed tree, against a freshly measured baseline on the same
`main` (`1902133bf`), both trees set up identically:

| | rest | `eval/run.test.mjs` | total | fail | skipped |
|---|---|---|---|---|---|
| `main` `1902133bf` | 1857 | 56 | 1913 | 0 | 0 |
| this branch | 1866 | 56 | **1922** | 0 | 0 |

`npx eslint scripts` exits 0 (lockfile-pinned `eslint@9.24.0`).

Rendered against the real store end to end, over all three replicates. Section 4 prints the
panel's `$30.49 ($29.53–$32.91)` spend across 3 replicates, `$4.17` per review and
`9.3 min` per review across 21 observations, `0 of 21` replays untimed, and CodeRabbit's
`6.8 min (2.6–14.4)` over 7 items — each with its `n` and its interval, in separate tables.
Also rendered against a payload written before the arm's timing read existed, which is what
is committed in the store today: the panel's figures still render and the latency cell
prints the scorer's own reason. **This PR does not depend on the scorer-side change.**

**18 mutations, 18 caught**, including the two that were the point: a cross-arm ratio
printed as a sentence, and the two arms' minutes placed in one table.

The byte-identical re-render test passes untouched — the report still has no clock.

⚠ `eval/run.test.mjs` is flaky in this local environment when another session runs
concurrently, on the two `END TO END` tests that snapshot `os.tmpdir()`. It reproduces on
`main` at 4 of 5 runs, and this diff touches neither `run.mjs` nor `run.test.mjs`.
Pre-existing; not fixed here.

## Risk Assessment

- **User-facing risk:** none in the product. Nothing in the gating path is touched; this
  module is a read-only renderer that no workflow invokes and the panel never imports. The
  real risk is *editorial*: this section is the one artifact in the benchmark with an
  audience, and a figure that looks comparable and is not would be worse than the blank it
  replaces. That is what the two-block layout, the per-table ratio test and the section 6
  limit are for.
- **Data/security risk:** none. No network, no store writes beyond the report file the
  renderer already wrote, no clock. A render is reproducible from committed data.
- **Rollback plan:** revert; the section returns to its single line. The renderer reads a
  payload it does not write, and it degrades to `not computed` — rather than throwing — on
  a score file written before the latency read existed, so an old store still renders.

## Notes for Reviewers

- **AI assistance:** written with Claude Code. Every figure quoted above was rendered by the
  code against the real store rather than transcribed, and the tests were mutation-tested.
- **Follow-up:** this changes the renderer, not the store. The published report under
  `reports/` does not move until somebody re-runs the scorers and renders again.
- **For whoever owns the complementarity scorer:** emitting a per-finding count of undecided
  pairs would let section 2 print the exact ceiling budget instead of bounding it by the
  whole queue. That file is not touched here.


dlgpdmsly2:
@claude review

dlgpdmsly2:
@coderabbitai review

dlgpdmsly2:
Thirteen findings, six of which were three restatements of two root causes. Five defects fixed in `b2d1254b`; three declined, and one of those has since expired on its own.

**Fixed.**

- **A refused cost-vs-size fit was labelled `not-measurable`** — which claims no such quantity can exist however long anyone runs anything, and a third priced item disproves that. The fallback is `not-computed` now; `suppressed` is kept for the case where the payload states the threshold it failed.
- **§6's latency caveat was gated on CodeRabbit's latency alone**, so a score file carrying our wall clock and not theirs lost the caveat entirely, fallback included. That is the payload where it matters most. The gate is now "either arm rendered minutes"; with neither it stays silent, which is correct rather than arbitrary.
- **A latency cell took its `n` from the parent of the series `series()` had just validated.** Equal in today's producer, so invisible — and a payload carrying `ms` without the sibling count made `figure` refuse and aborted the whole render.
- **Interval enforcement was asymmetric**: CodeRabbit's minutes refused without an interval name, ours fell back to the literal `unnamed`. Ours is the figure likeliest to be quoted against theirs. Both refuse.
- **The production ratio was hard-coded beside the numbers it came from.** Recomputed from those four values it is **2.3×, not 2.2×** — the 2.2× traces to a different figure, the production median in `agent-review-panel.yml`'s header, not to this pair. It is derived from the constant now, so it cannot disagree with the pair it summarises. The test recomputes it rather than pinning a literal, because a literal happily agreed with the wrong number.

Seven tests, covering the four paths flagged as uncovered: the second-anchor row, the trigger paragraph, a present fit row, and a priced CodeRabbit cost.

**Declined — and now moot.** Three findings said §4's pooled keys, the latency block and `min_n` are absent from the merged scorer. That was true of `main` `1902133bf`; #827 merged at 08:05Z and `main` `bb07acd4` now exports `MIN_FIT_ITEMS` and emits `panel.review_wall_ms`, `panel.review_cost_usd`, `panel.latency_interval` and `coderabbit.latency.self_timed`. Rendered against that scorer, §4 prints every headline — panel `$30.49 ($29.53–$32.91)` / `$4.17` / `9.3 min` / `0 of 21`, CodeRabbit `6.8 min (2.6–14.4)` and the second anchor `6.7 min (2.8–14.8)`, each with its `n` and interval — and §6 shows the derived `2.3x`. Every path called unreachable is reachable.

Before that landed they were declined on the merits: §4 rendered each missing field as an absence **with the scorer's own reason**, which is the required behaviour. Reading the per-replicate `review.wall_ms` instead was rejected because that payload names no interval for those minutes, and publishing our minutes without their interval is the one thing the recorded interval decision forbids.

**Verification.** 1959 + 56 = **2015**, 0 fail, 0 skipped, against **2008** measured on this branch's own head before the review round — +7, both trees set up identically; `eslint scripts` exits 0. Re-measured against the branch tip rather than `main`, because the *Update branch* merge at 08:00Z brought 53 `scripts/` files and the earlier 1857 baseline describes a tree this branch no longer has; neither `report.mjs` nor `report.test.mjs` was among the 53, checked before rebuilding on the merge rather than over it. **29 mutations, 29 caught.** Two survived the first pass and both were real test gaps rather than ineffective mutations — the `n`-source fix needed a fixture where the series and its parent disagree, since they are equal by construction in the producer that exists.

hotsunchip (approved):

PR #829, opened 2026-08-13
Read adjudicated pair labels in the complementarity band

dlgpdmsly2:
## Summary

Adds `scripts/agent/eval/pair-labels.mjs` — a record definition for an adjudicated pair of
findings, plus the resolver that turns those verdicts into counts — and teaches
`eval/complementarity.mjs` to read them when it computes the overlap band.

Every field the scorer's payload already had keeps its current meaning; the resolved band and
its provenance are new fields beside them. Nothing is written: the label tree is read-only
here, and the matcher is untouched.

## Why

`finding-match.mjs` answers `match` / `maybe` / `no`, and `groupFindings` never merges a
`maybe`. That fail direction is right — a false match suppresses a candidate, and a suppressed
candidate is a miss nobody recovers — but it means every undecided cross-arm pair counts as two
unique catches, one per arm. So the overlap is reported as a band, and its ceiling is currently
*saturated*: every CodeRabbit-only defect class has an undecided panel candidate, which is why
the scorer prints `Read "24 unique to CodeRabbit" as 24 UNRESOLVED pairs, not as 24 established
misses`. Deciding whether two findings are one defect is adjudication, not scoring, and until
now there was nowhere for such a decision to live.

**What a verdict may move.** With `b` classes already shared, `s` CodeRabbit-only classes
resolved `same`, `d` distinct panel classes newly shared, and `m` still holding an undecided
pair:

```
floor   = (b + d) / (classes − s)
ceiling = (b + d + m) / (classes − s − m)
```

A `same` verdict raises the floor and leaves the ceiling exactly where it was — the numerator
gains what the denominator loses. **Only a class with every one of its pairs decided can move
the ceiling**, so one `different` verdict on one of a class's pairs resolves nothing: any of the
others could still be the match. `d` and `s` are deliberately different counts, because two
CodeRabbit classes resolving into one panel class make that panel class shared once while
removing two classes from the union.

**A pair label does not re-partition the panel's own classes.** A `same` verdict says one
CodeRabbit finding and one panel finding are the same defect; it does not say two *panel*
findings are, even when both are labelled `same` against the one CodeRabbit finding. Closing
that under transitivity would merge two panel classes on the strength of a pair nobody
adjudicated — measurably narrowing the band — and `groupFindings` uses complete linkage
precisely to refuse that inference. The matcher owns the panel's partition; labels resolve
pairs. No threshold, no `clusterFindings`, no `gateFor`.

Three consequences of the same discipline: `insufficient-basis` is a third verdict rather than a
spelling of `different`, so a class holding one stays unfinished; trust tiers are never pooled,
and `resolveClasses` takes one tier with no argument meaning "all of them", so a pooled band
cannot be expressed; and "no label moved this number" resolves to one of seven stated causes,
because a band that read the same for an adjudicated replicate and an unadjudicated one would be
this subsystem's oldest bug — an arm's silence read as a zero — one level up.

Each record carries the `diff_sha256` of the item it was adjudicated against, and a record
disagreeing with that item's `meta.json` is **refused rather than skipped**: the item was
re-extracted after the adjudication, so the verdict is about code the store no longer holds, and
that is true of every label for that item. Its input is required rather than optional, for the
reason `assertEffort` failed — a guard whose input never arrives never fires. Conversely the read
path degrades and names what it dropped, because a dropped label can only widen the band.

**Measured on the frozen pilot corpus.** The two replicates with no adjudicated pairs are
unchanged and state why. The third moves `3.5% → 7.3%` at the floor with its ceiling identical at
`20.4%`: no defect class has every one of its pairs decided, so no ceiling moves, which is the
correct result on a partial label set rather than a limitation of the implementation.

The tier rule is the part that earned itself. A model-adjudicated tier landed while this was
being written — from a labeler that failed its own pre-registered validation — and it is still
arriving: measured three hours apart, its ceiling read `17.2%` and then `14.2%`, finishing defect
classes no human has finished, while the human band stayed at `20.4%` with nothing finished
apart. A pooled band would therefore not merely have been narrower than the evidence justified;
it would have been a **moving** number, drifting downward with each batch, with nothing in the
output saying which tier moved it.

## Linked Issues

None. Follows the complementarity scorer (#779) and is described in
`docs/tasks/active/20260813-pair-label-consumer-todo.md`.

## Author checklist

- [x] Changes follow the conventions in the touched packages; any deviations are explained in
      *Why* above.

## Verification

Skip reason (if applicable): n/a — this diff is `scripts/agent/**` plus two task docs.

Measured locally from the committed tree, on the two `node --test` invocations the `agent:tests`
lane runs:

- **1988 + 56 = 2044 tests, 0 fail, 1 skip**, against a freshly measured **1986 + 56 = 2042, 0
  fail, 1 skip** on this branch's own base. Both trees were extracted separately with the same
  `node_modules` symlinked into each, so the skip count is comparable. This change adds **34**
  tests in total; 32 of them are inside that base already, since the base is this branch's first
  commit with `main` merged into it.
- ⚠ **The branch is 12 commits behind `main`**, which has since refactored `scripts/agent/` — a
  `vendor/pipeline/` move and ~29 agent test files removed — so a count taken against `main`
  today (`1272 + 58`) is not comparable to either figure above. The two imports that moved are
  already re-pointed here and nothing else this change touches has moved, so the update should be
  mechanical, but it needs doing before merge.
- `npx eslint scripts` exits **0** on `eslint@9.24.0`, the lockfile pin.
- **33 of 33 mutations caught.** Every new test was checked by breaking the code it protects and
  confirming that test — not merely some test — goes red. One survivor was first proven
  *ineffective*, then closed with a new test whose fixture ids are chosen so sort order is
  adverse.
- End to end against the real adjudication set: the human-tier band is `12/165 = 7.3% .. 30/147 =
  20.4%` from 43 of 45 verdicts matching a live pair, with `ceiling_moved: false`. The
  arithmetic's anchor is pinned to a fixed 23-verdict input — `11/166 = 6.6% .. 30/147 = 20.4%` —
  so it stays reproducible as that set grows, which it did during this work.
- `report.mjs`, unmodified, renders the same three bands from a resolved payload that it renders
  today.

Not verified, and stated in the task doc: a `distant` tier, a stale `diff_sha256`, a key
collision and an unreadable record occur nowhere, so those paths are fixture-only. The
ceiling-moving path is exercised on real data by the model tier and remains unreachable in the
human one, which is the correct state of that evidence.

Known pre-existing failure, **identical on this branch and on unmodified `main`**:
`eval/run.test.mjs`'s *"END TO END: a failed item KEEPS its raw panel output"* asserts that no
`eval-item-*` directory appears in the shared `os.tmpdir()`, so it fails when that directory
holds debris from any other run — 54/56 on both trees, 56/56 on both with a private `TMPDIR`.

## Risk Assessment

- **User-facing risk: none.** Nothing here is imported by the review panel or the gate. The new
  module is read by one offline scorer; `severity.mjs`, `review-panel.mjs` and every gating path
  are untouched, and `finding-match.mjs` is imported exactly as before with no threshold or
  matcher change.
- **Data/security risk: low, and it is worth naming rather than dismissing.** The scorer reads
  two new places — a `labels/` subtree and each corpus item's `meta.json` — and writes nothing;
  it has no `put`, no `--out`, and the existing "no default `--root`" rule is unchanged. The real
  exposure is *numerical*: this changes what a published band may say. Three guards bound it — a
  stale `diff_sha256` refuses the run, conflicting verdicts for one pair refuse it, and the
  pre-existing unlabelled band is preserved untouched beside the resolved one so any consumer can
  compare. The CodeRabbit arm's GitHub calls are read-only, as before.
- **Rollback plan:** revert. `pair-labels.mjs` is new and imported only by
  `complementarity.mjs`; the payload's additions are additive, so a revert restores the previous
  numbers exactly and `report.mjs` — which renders those numbers and is not touched here — is
  unaffected either way.

## Notes for Reviewers

- **AI assistance:** written with Claude Code (Claude Opus 5). The arithmetic, the transitive
  closure decision and the verdict vocabulary were reviewed against the adjudicated records on
  disk and checked end to end against an independently computed figure; the mutation pass is
  described under *Verification*.
- **The commit message is the long form**, and
  `docs/tasks/active/20260813-pair-label-consumer-todo.md` carries the decisions, the fail
  directions and the numbers.
- **Where to push hardest in review:** `resolveClasses`'s numerator/denominator split (`d` vs
  `s`), and the decision not to close `same` verdicts transitively over panel classes. Those two
  choices set the floor, and the direction of an error in either flatters us.
- **A pre-existing defect this PR found and deliberately did not fix:**
  `unresolved.jaccard_upper_bound` can exceed 1 when several CodeRabbit-only classes point at one
  panel class — reproduced on `main` as `3/1 = 3.0`. It affects no current number and wants its
  own change, since fixing it here would alter a field this PR is otherwise careful to leave
  alone. Written up in the task doc.
- **Second commit is the review round.** Two guards, both at places where a rule lived only as a
  convention: `opts.pairLabels` is now refused in the `union` / `intersection` views rather than
  silently ignored, and `readPairLabels` rejects a record whose `corpus_version` disagrees with
  the directory it came from — without that, a misfiled record aborted the run with an error
  about a diff hash instead of about the file. It also corrects a figure that was wrong in two
  places and, since #817 landed `eval/labels.mjs`, imports `LABELS_DIR` from it instead of
  keeping a second literal for the same path.
- **Follow-up work:** the report renderer can now show which trust tier moved a band and whether
  the ceiling moved; the fields are listed in the task doc. This PR does not touch that module.

## Summary by CodeRabbit

* **New Features**
  * Added support for adjudicated finding-pair labels in complementarity scoring.
  * Reports now show label availability, provenance, pair keys, class mappings, and label-resolved score bands.
  * Supports trusted label tiers, alternate key formats, and shared, different, and insufficient-basis findings.

* **Bug Fixes**
  * Added validation and drift checks for malformed, conflicting, stale, or incompatible label data.
  * Preserved existing overlap metrics and unresolved bounds while improving labeled-pair resolution.
  * Prevented labels from affecting pooled union and intersection views.

* **Documentation**
  * Added guidance covering label consumption, scoring behavior, known issues, and verification results.


dlgpdmsly2:
@coderabbit review

dlgpdmsly2:
@coderabbitai review

hotsunchip (approved):

PR #830, opened 2026-08-13
Delete the pipeline mirror from this repo

hwisoo-kim:
## Summary

`scripts/agent/` held a full copy of the pipeline that **did not execute** — the workflows check it out of `wafflebase/agent-pipeline` at a pinned commit. So the `agent:tests` lane could pass against code that is not the code that runs, which had already happened on three files in both directions. #813 repointed every retained importer at `vendor/pipeline/`, so the copy can now go: **68 files deleted**, everything except `package.json`.

`package.json` stays and changes meaning — it is measurement's manifest now. The hunters `await import` the Agent SDK and `eval/run.test.mjs` reads the pinned version out of it, so it is free to diverge from the pipeline's own.

## What had to move first

**`agent-implement.yml`'s `implement` job has no adapter.** It checks out *this* repo, so `metrics.mjs` and `session-job-summary.mjs` were resolving out of the mirror, and its prompt told the agent to run `set-state.mjs` from its own checkout. Deleting underneath them would have hard-failed the metrics step and silently skipped the summary — the failure mode from finding #33, where the code was fine and the *environment* was not. All three read `vendor/` now, and both modules joined the vendored set (their deps were already there, so the closure did not widen).

**`spec-to-pr.mjs`** gets the split the extraction plan called for, with one deliberate deviation. `handoff` keeps working with no checkout because `set-state.mjs` is vendored; the plan proposed a direct `gh api` label PUT, which would have dropped the mutual-exclusion and `reconcile` semantics that module exists to provide. `review` needs the **rubrics**, which are prose under active change in phase C4, so it takes `--pipeline-dir` / `$AGENT_PIPELINE_DIR` and **skips with a warning** when it has none. A path that exists but is not a pipeline checkout fails loudly — that is a typo, not a missing option.

**The lenses are vendored too**, which was not planned. Four retained suites read them off disk, and `eval/config-hash.mjs` *hashes* the manifest precisely so its numbers match the panel's — a paraphrased copy would defeat the hash it exists to compute.

## The drift lane goes back to blocking

Reverses #798. It was advisory because enforcement meant taxing every pipeline change with a second landing, and #797 got caught by exactly that on day one. There is nothing left to drift. What it asserts now:

1. no file the pipeline repo owns has reappeared under `scripts/agent/`, and
2. nothing that stayed imports a path that no longer resolves.

The second has teeth: while the mirror existed a stray `./severity.mjs` *resolved*, so the mistake was invisible. Now it resolves nowhere, and without this check it surfaces as a runtime crash inside a hunt or a replay, far from the commit that caused it.

## Also here

- `vendor-pipeline.mjs`'s `assertClosed` had the same shape-matching hole fixed in the drift matcher — double-quoted `from`/`import()` only, so an unclosed side-effect import would have passed and failed at runtime instead.
- `docs/design/harness-engineering.md` cited 34 now-moved files across 89 places; `verify:entropy` fails on refs that do not resolve. They are now cited as `` `agent-pipeline:<file>` ``, with a note establishing the convention. Historical sections still describe the system as built — the behaviour is unchanged, only the location moved.
- `changed-areas.test.mjs` used deleted filenames as router fixtures; re-anchored on retained ones. `eval/test-lane.test.mjs`'s sentinels were two deleted mirror tests, likewise re-anchored — the property under test is unchanged.

## Test plan

- `agent:tests`, reproduced as CI runs it (recursive glob, split invocation, `node_modules` absent): **1146 pass / 0 fail**, plus `eval/run.test.mjs` **56 / 0**
- `scripts:tests`: **144 / 0** (19 in the drift suite, 4 new for the reappearance check)
- `verify:entropy`: 0 doc-staleness issues
- vendor integrity: 27 files verified offline at `249dd44a6`
- `spec-to-pr review` with no checkout warns and exits 0; with a bad `--pipeline-dir` exits 1; the vendored `set-state.mjs` runs
- **Mutation-tested**: restoring `severity.mjs` exits 1, reverting one import to the mirror exits 1, clean tree exits 0

## Risk

The `implement` job's three repointed paths are the one thing no test here observes — they only run on a real runner, which is finding #34's criterion. They are plain path changes to files proven present by the vendor lane, but the first live `@claude fix` after this merges is worth watching.

## Summary by CodeRabbit

- **New Features**
  - Added automated agent lifecycle label management with validation, reconciliation, and dry-run support.
  - Added session summaries for usage, cost, duration, models, and completion status.
  - Added six configured review lenses and expanded reviewer guidance.
  - Local reviews can optionally use an external pipeline checkout.

- **Bug Fixes**
  - Pipeline drift checks now block CI when prohibited files or imports reappear.
  - Workflows now use trusted, pinned pipeline scripts.

- **Documentation**
  - Updated harness-engineering documentation for the revised review pipeline.


harrykim8672:
@claude review

harrykim8672 (approved):

PR #831, opened 2026-08-13
Measure a pull request by its own commits, not its base's

hotsunchip:
## Summary

The path filter has been mostly inert since it shipped, and its
`ci-config-changed` label has been firing on pull requests that never touched CI
config. Both come from the same place: the filter was measuring each pull request
against the wrong two commits.

**The head was the merge commit.** `actions/checkout` leaves `HEAD` at
`refs/pull/N/merge` for `pull_request`, and `merge-base(base.sha, merge_commit)`
is `base.sha` itself whenever the merge already contains it — so the three-dot
diff collapsed to two-dot and charged the base branch's own recent history to the
pull request.

**The base was stale.** GitHub stamps `payload.base.sha` when a pull request is
created or synchronised and then leaves it alone, while `refs/pull/N/merge` is
rebuilt against the base's *current* tip. The two ends drift apart with nobody
touching the branch.

#817 is the clean demonstration, because it ran twice without changing:

```
02:08  head 8de60d6fd   full=false heavy=false ciConfig=false   agent, docsProse
04:48  head 637226ef1   full=true  heavy=true  ciConfig=true    "a CI gating file changed"
```

Four commits reached `main` in between. The three gating files it was then
measured against — `package.json`, `scripts/verify-self.mjs`,
`scripts/verify-doc-index.mjs` — were all #821's. #817 touches five files, all
inert.

**Both bugs failed toward running more CI**, which is why nine merged pull
requests never surfaced them. The cost was a filter that quietly did nothing for
any branch behind its base — most branches — plus a label that trains reviewers
to ignore the label.

### What changed

`resolveRefs` returns `{ base, head }`. The head is always the payload's
`head.sha`; the base is the first of three sources to resolve:

| Order | Source | Why |
| --- | --- | --- |
| 1 | `refs/pull/N/merge`'s **first parent** | The base tip GitHub merged against, as fresh as the run. Trusted only when the merge's *second* parent is the payload head — the signature of the merge ref |
| 2 | `origin/<base.ref>` | Independent of the merge ref's shape |
| 3 | `payload.base.sha` | **Last.** The only one that can be stale; reaching it over-reports, which costs a full run and nothing else |

Source 2 is not redundant. Three-dot absorbs a stale base only while the branch is
*behind* it; a rebase, a force-push, or **Update branch** puts the branch *ahead*,
and then the collapse returns. A rebased branch is also fast-forwardable, and a
fast-forwarded merge ref has no second parent to read — so with source 1 alone,
the rebase case would silently fall through to the stale sha. That is the
pre-merge state of nearly every pull request.

The diff stays **three-dot**. Two-dot is also wrong — 15 files against 5 on #805 —
because commits the base has and the head lacks read as changes belonging to the
pull request. The dot count was never the bug.

### Two follow-ons the same evidence forced

**`ci-config-changed` now tracks the current resolution in both directions.** It
shipped add-only, so anything that set it once kept it set for the pull request's
life and a reviewer could not tell a stale label from a live one. Removing it
would have been worthless without the base fix above: the re-run fires on rebase
(`synchronize`), but the re-resolution was still wrong. It deliberately does *not*
clear on an unparseable resolution — not knowing what changed is not evidence that
nothing did.

**`packages/design-editor` joins `ci.inert`** as a leaf nothing imports. That
needed a second edit: an inert match short-circuits the `packages/` classification,
so an inert package never reaches the reverse closure and a lane selecting on
`pkgs` alone — or on `anyPkg` — becomes unreachable, i.e. skippable *and* untested
with the lane still looking like coverage. So the `designEditor` tag is claimed by
two lanes: `design-editor:check`, and `verify:entropy`, because #819 added the
package to `knip.json`'s `workspaces` and `verify:entropy` is selected by
`anyPkg`. Dead code under `packages/design-editor/` would otherwise have passed its
own pull request and first failed on `main`. A design-editor change now selects
6 of 28 lanes and still skips both heavy jobs.

## Test plan

- `node --test "scripts/test/**/*.test.mjs"` — **164 pass, 0 fail**
- `pnpm lint:scripts`, `pnpm verify:doc-index` — clean
- Every new guard mutation-tested: removing the fresh-base lookup fails 4 tests;
  removing the `designEditor` lane tag fails 4 with `inert tag "designEditor" … is
  claimed by no lane`
- The rebase scenario is a real temp-repo fixture covering all three base sources,
  including the fast-forward fall-through and the stale-sha fallback over-reporting
- Replayed against #817's and #805's real payload shas: both now resolve
  `full=false heavy=false ciConfig=false`
- **This PR touches `ciConfig` paths, so it is measured by the full suite** — it
  cannot use the filter to grade its own homework

### Not run locally

A complete `pnpm verify:self` did not finish on the final tree: the shared working
tree was switched to another branch mid-run. The lanes that can be affected were
run individually and pass (`scripts:tests`, `lint:scripts`, `verify:doc-index`,
`design-editor:check`); an earlier full run on this tree reached 17 of 28 lanes
green, its one failure being `design-editor:check` against stale `node_modules`
(#819 added a `vite` devDependency), which `pnpm install` resolved. CI runs the
whole suite here regardless.

### DIFF
--- a/.github/workflows/ci-report.yml
+++ b/.github/workflows/ci-report.yml
@@ -258,12 +258,36 @@ jobs:
             // depend on it: `ci.ciConfig` forces the full suite and
             // .github/CODEOWNERS requires an owner's review. So a failure here is
             // a warning, not a red X.
-            if (areas && areas.ciConfig) {
-              try {
+            //
+            // The label tracks the CURRENT resolution in both directions. Adding
+            // it without ever removing it made it monotonic: a PR that dropped a
+            // gating file after one run, or was mislabelled by a resolver bug,
+            // wore the label for the rest of its life and the reviewer had no way
+            // to tell a stale label from a real one. A label nobody can trust to
+            // be current is worse than no label.
+            //
+            // The cost of dropping monotonicity: `concurrency` above is keyed on
+            // the triggering run, so two runs for one PR do not serialise and an
+            // older one finishing last can write a stale answer. Accepted, because
+            // the next run corrects it, the comment body already had exactly this
+            // property, and nothing mechanical reads the label.
+            const LABEL = 'ci-config-changed';
+            try {
+              if (areas && areas.ciConfig) {
                 await github.rest.issues.addLabels({
-                  owner, repo, issue_number: prNumber, labels: ['ci-config-changed'],
+                  owner, repo, issue_number: prNumber, labels: [LABEL],
                 });
-              } catch (error) {
-                core.warning(`Could not apply the ci-config-changed label: ${error.message}`);
+              } else if (areas) {
+                // Only when `areas` parsed. A missing resolution means we do not
+                // know what changed, and guessing "not a config change" would
+                // quietly clear a label that is telling the truth.
+                await github.rest.issues.removeLabel({
+                  owner, repo, issue_number: prNumber, name: LABEL,
+                }).catch((error) => {
+                  // 404 is the normal path: the label was not on the PR.
+                  if (error.status !== 404) throw error;
+                });
               }
+            } catch (error) {
+              core.warning(`Could not update the ${LABEL} label: ${error.message}`);
             }
--- a/.github/workflows/ci-report.yml
+++ b/.github/workflows/ci-report.yml
@@ -46,6 +46,7 @@ on:
 permissions:
   contents: read
   actions: read # list the triggering run's jobs and download its artifacts
+  pull-requests: read # bind the artifact's PR number to this run, and list its files
 
 # Keyed on the triggering run so a re-run supersedes its own report rather than
 # racing it. Not keyed on the PR: two different runs for the same PR are two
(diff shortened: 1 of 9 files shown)


hotsunchip:
@claude review

hotsunchip:
## Review-panel findings — 6 fixed, 2 not (`9f2ff0563`)

**Fixed**

- **1, 2, 3, 6** — one root cause: the reporter trusted the fork-written artifact.
  The label is now recomputed from `pulls.listFiles` + `ci.ciConfig` on the default
  branch, and the PR number is bound to the run (`pr.head.sha === run.head_sha`),
  which also closes the IDOR on the comment path. The label is held, never cleared,
  when that data is incomplete — `ciConfig: false` also means "the resolver gave up".
- **7** — the `push` branch of `resolveRefs` had no coverage. Added.
- **8** — correct, and the sharpest of the eight: neutering `mergeRefBaseTip` left
  all 61 tests green. The subtest now hides the base branch so only source 1 can
  answer.

**Not fixed**

- **5** — doesn't hold. `verify-doc-index` only checks that directories *present* in
  `packages/` are indexed; deleting the whole package still passes, so nothing
  inside it can fail that gate.
- **4** — real, but pre-existing and this diff doesn't touch `ci.yml`. Recorded as a
  scope limit in the design doc + Phase 5 instead of expanding this PR.
  **A maintainer's call whether that should block.**

**Before merge**

1. `ci-report.yml` can't be exercised here — `workflow_run` runs the default
   branch's copy, so this PR's label came from the old code. The new path first runs
   after merge; **#817 should lose `ci-config-changed`.**
2. The App token deliberately keeps `issues: write` only, with the new reads on the
   ambient token. Requesting scopes the installation may lack would fail the token
   step and silently degrade it to read-only, breaking comments too.

Reasoning and evidence: `docs/tasks/active/20260812-path-aware-ci-todo.md`, Phase 4d.

harrykim8672 (approved):

PR #619, opened 2026-07-31
Board SP2: sticky notes, image paste, minimap

hackerwins:
## Summary

SP2 of the board infinite canvas ([design](docs/design/board/board-whiteboard-elements.md)) — three whiteboard affordances, all **board-local** with **zero slides engine change**:

- **Sticky notes** — a toolbar split-button drops a preset `roundRect` shape (solid fill + drop shadow + middle-anchored shrink-autofit text) at the viewport center, selected and text-editing. Once placed it's an ordinary shape: move/resize/rotate/snap/collab/undo for free. No new element type.
- **Image paste / drag-drop / file-picker** — reuses the shipped upload pipeline (`uploadImageFile` → `POST /api/v1/workspaces/:id/images`) and `setupSlidesImagePaths`, centered on the current viewport. The only shared-helper change is an **additive, optional `center` param** on `insertImageOnSlide` / `setupSlidesImagePaths` (slides behavior byte-identical when omitted).
- **Minimap** — a view-local vanilla-DOM overlay: scene snapshot via `renderThumbnail`, current-viewport rectangle, drag-to-pan, toggle (on by default), shown in read-only. No CRDT / presence.

New board-local modules: `sticky.ts`, `board-image.ts`, `board-minimap.ts`, `minimap-geometry.ts`. Wiring in `board-view.tsx` / `board-toolbar.tsx` / `board-detail.tsx`. The only non-board files touched are the two frontend slides image helpers (additive param) and a `harness.config.json` chunk-count bump (142 → 144).

## Test plan

- New tests: `sticky` (5), `insert-image` center override (2), `board-image` adapter (1), `minimap-geometry` (6), `board-minimap` factory smoke (2). Frontend suite 886 → 902 passing.
- `pnpm verify:self` green (all 11 lanes, incl. 144-chunk gate) on the rebased-onto-main tree.
- Slides suite unchanged (2629) — confirms the additive `center` param leaves slides untouched.
- Built via subagent-driven development: 6 TDD tasks, each independently task-reviewed (all Spec ✅ / Approved), plus an Opus whole-branch review — **Ready to merge = Yes, 0 Critical / 0 Important**. It independently verified the SP1 reachability lesson (no reused-editor gesture hits `YorkieBoardStore.notSupported()`), leak-free teardown of all three features in the shared mount effect, gated paste routing, and read-only safety.

### Manual smoke (recommended before merge)
[...]

PR #842, opened 2026-08-14
Add image upload to the notes markdown editor

hackerwins:
## Summary

Notes could not hold an image: the only way in was to type a markdown link to one already hosted somewhere. Paste, drop, and a toolbar picker now upload through the workspace image endpoint that sheets/slides/board already share, and insert `alt` at the insertion point.

No backend change — `POST /api/v1/workspaces/:wid/images` and the frontend's `uploadImageFile` were already there. The preview needed nothing either: `markdown-it`'s image rule already renders the result.

CodePair's imageUploader plugin is the reference, but three of its choices do not survive a collaborative markdown editor:

- **A text placeholder cannot show upload progress.** A note's body is one Yorkie `Text` CRDT, so placeholder text replicates to every peer, enters the undo history, and remains as garbage if the upload fails or the tab closes mid-flight. The placeholder is a view-local widget decoration instead, mapped through every transaction — the user's own typing and a peer's remote edit alike — so an image that finishes uploading seconds later still lands where it was dropped. CodePair reads the selection *after* the await, which puts the image mid-word.
- **A drop inserts at the drop coordinates**, not at the caret.
- **A failed upload is reported.** The engine's callback resolves with `null` to mean "the host already told the user"; the frontend catches what `uploadImageFile` throws and raises a toast, where CodePair swallows the rejection.

One insert is one `input` transaction, so `noteSync` records it as a single undo unit — Ctrl+Z removes an image in one step. When a payload carries no image the handlers decline the event rather than `preventDefault`, so ordinary text paste and drop are untouched.

The engine keeps its distance from the frontend: `initialize()` takes an optional `uploadImage` in a new options bag, and a read-only mount never receives one, so the extension is absent rather than guarded per event.

### From the branch review

Two rounds of self-review over the diff. Three findings in the first round, two of them real bugs in the original implementation:

- **A batch inserted in completion order, not file order.** Placeholders in one paste share an anchor, so the first upload to return took the front — paste three screenshots on a jittery network, get them back shuffled. Requests still all start at once; only the inserts are serialized. The original ordering test passed *vacuously* (its mock resolved in call order), so it was replaced with one that resolves the second upload first.
- **A Yorkie snapshot resync put the image at the top of the note.** `noteSync` applies a `replace` remote change as one change spanning the whole document, and every anchor inside that deleted range collapses to position 0. Those anchors are now dropped and the insert falls back to the caret.
- **`isAuthExpiredError` was missing**, so a failure caused by an expiring session toasted on the way to `/login` (the guard #619 established, now in ~15 handlers).

The second round found no reachable bug and hardened one thing: serializing the inserts made that loop the single choke point for a batch, so a throwing insert would have abandoned every image behind it with the rejection swallowed. Each commit now fails on its own.

### Known limitations

- **No image upload behind an editable share link** — an anonymous share-link editor cannot authenticate to the workspace image endpoint, so paste and drop fall through to normal text handling there.
- **A stalled first upload holds back the rest of its batch**, since ordering is enforced by committing in sequence. Visible and honest, but a per-item timeout would be better.
- **No engine-level test of the store/undo integration** — `image-upload.test.ts` mounts the extension standalone. The one-undo-unit claim rests on `insertImage` dispatching a single `input` transaction plus `note-sync.test.ts`'s existing coverage of that shape.

Design: [`docs/design/notes/notes.md`](docs/design/notes/notes.md) (P2 section). Task: `docs/tasks/active/20260814-notes-image-upload-todo.md`.

## Test plan

- `pnpm verify:fast` green (ran on every commit via the pre-commit hook); `verify:self` green via pre-push.
- 12 new tests in `packages/notes/src/view/image-upload.test.ts`: paste inserts at the caret, drop inserts at the drop point, the anchor tracks edits made during the upload, the placeholder appears and clears, failure and host-cancel leave the document untouched, an image-free paste still pastes text, a batch inserts in file order with the second upload resolving first, a whole-document resync falls back to the caret, teardown mid-upload is dropped, and a read-only editor never uploads.
- **Every test in that file was checked by mutation** — each was confirmed to fail when the behavior it protects is broken. None are vacuous.
- Not yet done: manual smoke in `pnpm dev` (paste, drop, toolbar, failure, undo).

## Summary by CodeRabbit

* **New Features**
  * Added image uploads to notes through the toolbar, paste, and drag-and-drop.
  * Supports PNG, JPEG, GIF, and WebP files, with multi-file selection and ordered insertion.
  * Displays temporary upload placeholders, preserves insertion positions during edits, and groups completed uploads for undo.
  * Provides failure feedback and restores editor focus after insertion.
  * Image uploads are unavailable in read-only notes or when not configured.

* **Documentation**
  * Updated feature-parity notes and active task documentation for image uploads and related work.

### DIFF
--- a/docs/design/notes/notes.md
+++ b/docs/design/notes/notes.md
@@ -251,7 +251,7 @@ board already share, through the frontend's existing
 `uploadImageFile(file, workspaceId)`. The preview needed nothing either —
 `markdown-it`'s image rule already renders the result.
 
-The design departs from CodePair's `imageUploader.ts` in three places, each
+The design departs from CodePair's `imageUploader.ts` in several places, each
 because the naive version misbehaves in a collaborative markdown editor:
 
 - **The in-flight placeholder is view-local, not text.** A note's body is one
@@ -262,7 +262,17 @@ because the naive version misbehaves in a collaborative markdown editor:
   transaction — the user's own typing *and* a peer's remote edit — moves the
   pending insertion point, so an image that finishes uploading seconds later
   still lands where it was dropped. CodePair reads the selection *after* the
-  await, so typing during an upload drops the image mid-word.
+  await, so typing during an upload drops the image mid-word. The one change
+  mapping cannot survive is a whole-document replacement — how `noteSync`
+  applies a `replace` remote change, i.e. a Yorkie snapshot resync — because
+  every anchor lives inside the deleted range and would collapse to position
+  0. Those anchors are dropped instead, and the insert falls back to the
+  caret rather than dumping the image at the top of the note.
+- **A batch inserts in file order, not completion order.** Every request in a
+  paste or drop starts at once, but the inserts are committed in sequence:
+  the placeholders share one anchor, so letting each insert as it resolves
+  would reorder the batch by network speed — paste three screenshots, get
+  them back shuffled.
 - **A drop inserts at the drop coordinates** (`view.posAtCoords`), not at the
   caret. Dropping a file on a paragraph and watching the image appear
   elsewhere is the most confusing part of the naive implementation.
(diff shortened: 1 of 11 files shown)


hackerwins:
@coderabbitai review this.

PR #850, opened 2026-08-14
Return the agent pipeline to this repository

hwisoo-kim:
## Summary

Reverses the extraction of the pipeline into `wafflebase/agent-pipeline`. It lives in
`scripts/agent/` again, and the agent workflows run this repository's own code.

Eight merged PRs are reversed: **#769** (source from the central repo), **#784**
(auth-smoke shim), **#796** + **#798** (the drift lane), **#800** (vendoring), **#813**
(repoint to vendor), **#830** (delete the mirror), plus agent-pipeline #1.

`git revert` does not apply — six of the eight conflict — so this goes forward instead:
restore what #830 deleted, repoint imports off `vendor/`, delete the split-only
machinery, restore the pre-split workflows.

### The restore is surgical, and that is the important part

Overlaying `scripts/agent/` wholesale from before the mirror deletion — the obvious way to
do this — **also reverts every change merged into surviving files since**: nine files
across **#829**, **#835**, **#836**, **#843** and **#847**, which is the UI-hunter and eval
work.

That failure is invisible to CI. The overlay reverts each file's **tests alongside its
code**, so they agree with each other and every lane stays green while the regression
lands.

So only the 68 paths #830 deleted, plus the 16 it edited in place (each verified untouched
by anything since), are restored from history. Everything else under `scripts/agent/` keeps
`main`'s content and had its imports rewritten in place. A gate in the test plan below
asserts this held.

### Two extraction-era changes are kept deliberately

- **`GH_REPO` at workflow level** (#790). `gh` expands `{owner}/{repo}` by shelling out to
  git; without this, every step running before the repo is checked out fails with "not a
  git repository". Its absence killed the autonomous loop repo-wide for a day. Present at
  workflow level in all 10 `agent-*.yml`.

- **The trusted-code fix.** The pre-split workflows read pipeline code from
  `./scripts/agent/` — a path the PR author controls — *after* checking out the PR's own
  branch, while holding a write-scoped token. Six sites in `agent-iterate-ci.yml` and two
  in `agent-review-panel.yml`'s `fix` job now read from `$RUNNER_TEMP/agent-tools/`, staged
  from trusted main and outside the workspace. **This weakness predated the split**; a
  blind restore would have brought it back.

### Guards

`scripts/test/agent-workflow-guards.test.mjs` keeps the check that pins the property above,
rewritten to key on **order** rather than on a file's absence from `scripts/agent/`. The old
predicate was true only while the pipeline lived elsewhere — with the files back it
inspected **0 of 68** invocations while reporting success. It now inspects 28, and both
remaining guards are mutation-tested (each fails when its invariant is broken).

Its two split-only tests — sha-pinning and adapter ordering — are deleted as vacuous, as is
the sha-pin guard in the restored `checks.test.mjs`, whose own vacuity check was already
firing correctly.

`pipeline-drift` is removed from `ci.yml`, leaving `changes`, `verify-self`,
`verify-browser`, `verify-integration`.

## Test plan

| Lane | Result |
|---|---|
| `agent:tests` (recursive glob, `node_modules` absent, as CI runs it) | 2059 pass / 0 fail / 1 skipped |
| `eval/run.test.mjs` (isolated invocation) | 56 / 0 |
| `scripts:tests` | 176 / 0 |
| `lint:scripts` | exit 0 — eslint covers 148 files under `scripts/agent`, `no-undef` active |
| `verify:entropy` | exit 0 (111 advisory, all pre-existing under `docs/design/design-editor/**`) |
| `verify:doc-index` | exit 0 |
| YAML parse, all 18 workflows | ok |

Additional gates run against this branch:

- **Preservation gate** — every path under `scripts/agent/` that differs from `origin/main`
  and is outside the intended restore sets was confirmed to differ in **import lines only**.
  #843's menu-leaf logic and #847's `COVERAGE_KEY_VERSION = 4` verified present.
- **Guard mutation tests** — undoing any one of the eight trusted-code fixes turns the order
  guard red; removing `GH_REPO` from any workflow turns the `GH_REPO` guard red.
- **Zero `vendor/pipeline` references** remain outside the deleted files.

`lint:scripts` is worth calling out: it is part of `verify:fast`, and restoring 61 modules
changes what it covers. It is green.

## Follow-ups (not in this PR)

- `mark-ready.mjs`'s exit codes 1/2/3 are a contract `agent-review-panel.yml` branches on,
  and nothing pins them.
- `scripts/agent/summarize-ci.mjs` still cannot render the `filtered` lane status. That is
  now an in-repo change rather than a cross-repo one.
- `wafflebase/agent-pipeline` gets archived (not deleted) once this merges.


hackerwins (dismissed):
Thanks for your contribution.

PR #857, opened 2026-08-14
Never publish upstream error text, and mask tokens by shape

hotsunchip:
## Summary

A `CLAUDE_CODE_OAUTH_TOKEN` was rotated with a stray space and the review panel
published it to a public pull request. A space makes an `Authorization` header
structurally invalid, so the HTTP client rejects it **before sending** and quotes
the value back. `classifyResult` carried that as `detail`, and the panel
interpolated `detail` into a lens summary reaching a check-run body, a PR comment
and the job summary.

A merely **wrong** token gives a clean 401 with no echo — the typo, not the
invalidity, caused the disclosure. Neither GitHub Secrets nor log masking could
have helped: the panel must hold the token to authenticate, and a PR comment is a
request body with no log for a scrubber to sit in front of.

## Why not port agent-pipeline#2

#850 returned the pipeline to this repository two hours after that PR opened,
restoring the unfixed files and deleting the vendoring machinery — there is no pin
to bump. It also keys its defence on a frozen five-name allowlist written before
**#854 spread the pipeline across nine credentials**. Raising that list to nine
moves the same fragility forward one step, so this removes the enumeration from the
safety path entirely.

## Two guarantees, neither aware of how many credentials exist

**1. No raw upstream text is published.** `classifyInfraError` maps a failure onto
a closed vocabulary — `AUTH_MALFORMED_CREDENTIAL`, `AUTH_REJECTED`, `USAGE_LIMIT`,
`RATE_LIMITED`, `UPSTREAM_ERROR`, `NO_RESPONSE`, `RUN_LIMIT_*`, `NO_OUTPUT` — and
every published string is *built* from it plus typed context. Stronger than
filtering, which failed here precisely because the value arrived in a shape nobody
had enumerated.

**2. Unconditional shape-based masking**, five layers: echo frames (matched by the
*sentence*, not the secret), known shapes consumed across whitespace so a malformed
token is masked whole, a high-entropy catch-all needing no prefix, keywords, and
live env values derived from `poolEnvNames()`.

| Input | Before | After |
| --- | --- | --- |
| `wfb_` key, JWT, `x-api-key` | masked | masked |
| `sk-ant-` key | **leaks** | masked |
| GitHub PAT (`ghp_`) | **leaks** | masked |
| Unknown/opaque format | **leaks** | masked |
| Malformed token (contains a space) | **leaks** | masked |
| git SHA, UUID, stack trace | intact | intact *(by design)* |

## Two further leak sites, found during review

Same defect class — the first sweep grepped property *access* (`.detail`) and
missed property *definitions*.

- **`auth-smoke.mjs`** echoed the credential to a public Actions log, and it is the
  tool you run *when a credential is broken*, so a malformed secret reaches it by
  design. GitHub's masker is the exact-substring matcher that whitespace defeats.
- **`review-execution.json`** shipped raw SDK messages as a workflow artifact.
  Fixed at the single producer all three logs share; nothing reads `result` back.

## Request ids

`req_…` is **extracted** as structured context — anchored pattern, re-emitted from
the literal match, refused if it overlaps a live credential — rather than exempted
from the filter, which would leave a hole open to whatever else matches. So
`context` is numbers **plus one pattern-validated string**, a real but deliberate
weakening of the original invariant.

This also fixed a fragility: `lensFailureSummary` was reclassifying the
already-redacted `detail`. It now carries `reason` through from `classifyResult`.

## Test plan

- **2185 agent tests pass** (2151 before; +34). `lint:scripts`, `verify:entropy`,
  `verify:doc-index` clean.
- **Mutation-tested** — each of the five layers disabled in turn fails its test;
  reverting `review-panel.mjs` to publish `err.detail` fails the leak check.
- **Leak check across all six published surfaces**, driving the incident's real SDK
  message through each renderer.
- **Counterweight test** pins git SHAs, UUIDs, paths and stack traces byte-identical
  — over-redaction would destroy the diagnostics this pipeline exists to produce.
- **Wire format pinned** against both parsers (`prior-findings.mjs` `INFRA_SENTINEL`,
  `rounds.mjs` `INFRA_SUMMARY`); breaking it carries a stale infra record forward as
  a code finding and sticks the loop a round later.

**Not verified:** no live API run — the SDK message shape is reconstructed from
`classifyResult`'s contract. `verify:fast` is green except three pre-existing flakes
(wall-clock assertions; all pass in isolation, in packages this branch does not
touch).

## Deploy note

There is no vendor pin — the workflows read `scripts/agent/` from the default
branch, so **merging this is the deploy**. The exposed credential was revoked
separately.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive credential masking for secrets, tokens, authorization data, and credential-like strings before publication.
  * Standardized infrastructure failure messages using safe error categories and approved context.
  * Preserved compatibility with existing failure-summary consumers and legacy records.

* **Bug Fixes**
  * Prevented upstream error details and credentials from appearing in job, session, review, and metrics summaries.
  * Improved account-limit detection while retaining support for older error formats.

* **Tests**
  * Added coverage for redaction, error classification, compatibility, legacy handling, and sensitive-data protection.


hackerwins (approved):
Thanks for your contribution.

PR #865, opened 2026-08-17
Keep an image pasted next to another from being merged away

### LINKED ISSUE #726: Docs: copy-and-pasting attached images to a document has wrong behavior
### What happened:

copy-and-pasting attached images to a document has wrong behavior.

image should pasted.

Select the attached image to the document, then copy and paste it.
- if caret is located after image, image is not pasted.
- if caret is located before image, caret is moved after image and image is not pasted. each time type a character, one pasted image is displayed. 

### Environment:

- Operating system:
- Browser and version:

shjohw12:
## Summary

Stop `normalizeInlineList` from merging structural inlines. An image (and a
page number) keeps its payload in the style while its text is a single
placeholder character, so one inline describes exactly one object — merging two
of them loses one.

## Why

Pasting an image immediately before or after another one dropped it: the image
never appeared, the caret behaved as though it had moved past one, and typing
afterwards surfaced copies of the original a character at a time.

`normalizeInlineList` merges neighbouring inlines whose styles compare equal,
and two image runs qualified. Concatenating them gives a two-character run under
a single `style.image`, which cannot describe two images: the second is dropped
while the offsets still count both. That one merge accounts for all three
reported symptoms — the missing image, the caret appearing to skip it, and the
copies appearing one per keystroke as later edits split the doubled run.

**Equality is the wrong test to fix this on.** Copying an image and pasting it
beside itself leaves two runs holding identical `ImageData`, so any by-value
comparison correctly reports "equal" and merges them anyway. `isStructuralInline`
refuses the merge on structure instead, regardless of how the styles compare.

`pageNumber` is included: the renderer replaces a page-number run *whole* with
the page number (`doc-canvas.ts`), so a merged pair renders one number while the
offsets count two.

### Two defects this also fixes

Probing the paste path before pushing turned up the same merge in wider forms —
both now covered by tests:

| Paste | Before | After |
|---|---|---|
| Text just **before** an image | `hi￼` — **`style.image` dropped, image gone** | `hi` + `￼` |
| Text just **after** an image | `￼hi` under `style.image` — text absorbed | `￼` + `hi` |

The first is the worse one: pasting plain text in front of an image deleted the
image, because the first run's style wins the merge. Nothing about that paste
touched the image.

## Linked Issues

Fixes #726

## Author checklist

## Verification

Skip reason (if applicable): none — locally `pnpm verify:self` passed (21 lanes
run, 9 filtered as unaffected, 0 failures), re-checked after rebasing onto
`cc0e281e0`. The boxes are left for the CI comment to confirm.

## Risk Assessment

- User-facing risk: low, and confined to paste. `normalizeInlineList` is reached
  only through `spliceInlinesAt` → `insertBlocks`, so nothing outside the paste
  path changes. Before pushing I recorded paste behaviour across six scenarios
  with and without the guard: plain text into text, differing styles, and
  multi-block paste came out **byte-identical**; only the three structural-inline
  cases changed, each from a broken result to a correct one.
- No unbounded run growth from refusing these merges: a structural inline is
  exactly one character, and `sliceInlines` drops zero-length slices, so such a
  run cannot split into two non-empty runs. Their count is bounded by the number
  of objects actually inserted. Ordinary text runs still merge as before, which
  is what keeps editing from fragmenting the list.
- Data/security risk: none. No persistence, network, or permission surface.
- Rollback plan: revert the commit. The guard is two clauses in one condition
  plus an exported predicate; removing them restores the previous behaviour.

## Notes for Reviewers

- UI changes: none.
- **Scope.** This is the paste half of #726. Copying an image that is selected by
  clicking it fails for a different reason, in a different file, and is fixed in
  a separate PR so each can be reviewed on its own.
- **`pageNumber` is deliberate, not speculative.** The renderer's whole-run
  substitution makes a merged pair render one number, and the probe above shows
  text pasted next to a page number being absorbed today. It is also the only
  clause with a single test guarding it — worth knowing if you are weighing
  test coverage.
- **Not fixed here:** `applyInsertText` guards `style.image` against absorbing
  typed text but not `style.pageNumber`, so *typing* after a page number in a
  header is still swallowed by the same substitution. Same root concept, but a
  different call path (`store/block-helpers.ts`), so it is being filed
  separately rather than widened into this PR.
- **How the tests were checked.** Each was confirmed to fail with the guard
  removed, and the guard's clauses were mutated one at a time to see which test
  catches which — earlier in this work two tests passed against broken code
  (one dispatched no bare-modifier keydown, one used two *different* images
  where the real case uses identical ones), so "a test exists" was not treated
  as evidence on its own.
- AI assistance: authored with Claude Code (Claude Opus 5) — the guard, the
  predicate, the tests, and this description.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pasted images and page numbers now remain distinct instead of being merged with adjacent content.
  * Pasted text no longer absorbs or removes nearby images and page numbers.
  * Regular text formatting continues to merge correctly when appropriate.

* **Tests**
  * Added coverage for image and structural-inline paste behavior.

### DIFF
--- a/packages/docs/src/model/types.ts
+++ b/packages/docs/src/model/types.ts
@@ -423,6 +423,22 @@ export function inlineStylesEqual(a: InlineStyle, b: InlineStyle): boolean {
   );
 }
 
+/**
+ * Whether an inline is a *structural* kind — an image or a page number —
+ * rather than styled text.
+ *
+ * These carry their payload in the style (`style.image`) while their text
+ * is a single placeholder character, so one inline describes exactly one
+ * object. That makes them unmergeable: concatenating two image inlines
+ * gives a two-character run under a single `style.image`, which cannot
+ * describe two images — the second is lost while the offsets still count
+ * both. Equality is not the test here; two *identical* images (copy an
+ * image, then paste it beside itself) must still stay separate runs.
+ */
+export function isStructuralInline(inline: Inline): boolean {
+  return inline.style.image !== undefined || inline.style.pageNumber === true;
+}
+
 // --- Table types ---
 
 export interface BorderStyle {
--- a/packages/docs/src/view/text-editor.ts
+++ b/packages/docs/src/view/text-editor.ts
@@ -1,5 +1,5 @@
 import type { Block, BlockCellInfo, CellAddress, DocPosition, DocRange, Inline, InlineStyle, HeadingLevel, TableCell } from '../model/types.js';
-import { generateBlockId, getBlockText, getBlockTextLength, DEFAULT_BLOCK_STYLE, createBlock, createTableBlock, normalizeTableMerges } from '../model/types.js';
+import { generateBlockId, getBlockText, getBlockTextLength, DEFAULT_BLOCK_STYLE, createBlock, createTableBlock, normalizeTableMerges, isStructuralInline } from '../model/types.js';
 import { Doc, type EditContext } from '../model/document.js';
 import { cloneBlockWithFreshIds } from '../store/block-helpers.js';
 import { serializeClipboard, deserializeClipboard, cloneTableCells, parseHtmlToBlocks, parseHtmlTableToTableCells, parseMarkdownTableToTableCells, parseMarkdownWithTables, WAFFLEDOCS_MIME } from './clipboard.js';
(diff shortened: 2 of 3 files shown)


hackerwins:
@coderabbitai review this.

hackerwins (approved):
Thanks for your contribution.

PR #866, opened 2026-08-17
Scope the changed-areas git calls to their own repository

hotsunchip:
A fixture-building test could write into the repository under test. It did, twice this week: fixture commits on the current branch, a stray `feature` branch, `core.bare`/`core.filemode` flipped, `user.name`/`user.email` overwritten, and the index replaced with 3,611 files staged as deleted.

## Cause

`scripts/test/changed-areas.test.mjs` builds fixture repos in a temp dir and passes `cwd` but not `env`. `cwd` does not decide which repository git reads — an inherited `GIT_DIR` wins over it. Five near-identical `git` closures had grown in that file, which is how one came to be written without `env`.

This repo already has the fix: `scripts/agent/git-env.mjs` exports `fixtureGitEnv()` / `repoScopedEnv()` for exactly this hazard, and other tests use them. One file had missed the convention.

## Reproduced, in throwaway `/tmp` clones

| environment | symptom |
| --- | --- |
| `GIT_DIR=<repo>/.git` | commits land on the victim's branch, `feature` branch appears, `core.filemode` flips, identity overwritten, index rewritten — tests fail loudly |
| `GIT_DIR=<repo>/.git/worktrees/<name>` | all of the above **plus `core.bare = true`**, after which the primary tree reports `fatal: this operation must be run in a work tree` |
| `GIT_INDEX_FILE=<abs>` | **suite exits 0** while the index is destroyed — 3,611 files staged deleted. Silent |
| `GIT_WORK_TREE` alone | tests fail, no damage |
| `GIT_INDEX_FILE=.git/index` | harmless — relative, re-resolves against the child's cwd |

Two findings worth keeping:

- **`core.bare` comes from the `GIT_DIR` path shape.** `git init`'s `guess_repository_type()` guesses bare for anything not ending in `/.git` and writes it to the *common* config — the primary repo's. Isolated: `<r>/.git` → false, `<r>/.git/worktrees/wt` → true.
- **git 2.43's `pre-push` exports no location variable at all.** Only `pre-commit` and friends export `GIT_INDEX_FILE`, relative and benign. The variable is ambient, so `--no-verify` is not protection — and `git push --dry-run` runs the hook before git decides the push is a rehearsal. That is documented in `git-env.mjs` now, because it is how one of this week's two incidents happened.

The silent `GIT_INDEX_FILE` variant is the reason this went unnoticed: a corrupting run that exits 0 looks like a passing one.

## Changed

- `scripts/test/changed-areas.test.mjs` — five duplicated closures collapsed into one `fixtureGit(dir)` carrying `env: fixtureGitEnv(dir)`, covering all 8 call sites.
- `scripts/changed-areas.mjs` — `resolveRefs` and `changedPaths` gain `repoScopedEnv(repoRoot)`. This is the read half of the same hazard: an inherited `GIT_DIR` would decide CI lanes from a different repository, and the failure mode is a green run against the wrong tree.
- `scripts/agent/git-env.mjs` — comments only.

## Test plan

- 203 `scripts/test` and 2,106 `scripts/agent` tests, `lint:scripts`, doc-index, doc-links — clean
- **Four reverts, each failing exactly the two new subtests** and no others. Neutering `fixtureGitEnv` to return `process.env` fails them too, which is what makes this a property test rather than a check that a helper is called
- The reproduction repeated **with the fix**, fresh clones, all three hostile shapes: 68/68 pass, clone byte-identical afterwards
- CI runs in full because `scripts/changed-areas.mjs` is in `ci.ciConfig` — a PR editing the lane resolver cannot use it to shrink its own coverage, so the full run is itself a check on the read path

## Not addressed

Three unscoped **read-only** git calls (`verify-entropy.mjs:139`, `review-scope.mjs:61`, `eval/extract-corpus.mjs:547`). Worst case is a wrong answer, never corruption, and each needs a caller audit — `extract-corpus`'s `repoSource` is optional, so `repoScopedEnv(undefined)` would throw. `spec-to-pr.mjs` passes neither `cwd` nor `env` by design and is outside the pattern.

Separately: `pnpm verify:fast` cannot pass in a fresh worktree, because it typechecks `slides`/`cli` against other packages' `dist/*.d.ts` while building only `core`. Pre-existing and unrelated.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented inherited Git settings from affecting repository comparisons, reference resolution, and change detection.
  - Improved reliability when creating test fixtures and running Git operations in isolated repositories.
  - Prevented false-success results, unexpected deletions, and accidental changes to unrelated repositories.

- **Tests**
  - Added regression coverage for hostile Git environment configurations.
  - Verified fixture operations and repository comparisons preserve the original repository state.

- **Documentation**
  - Documented the Git environment-variable hazard, observed effects, remediation scope, and validation approach.

### DIFF
--- a/docs/tasks/active/20260817-harness-git-env-todo.md
+++ b/docs/tasks/active/20260817-harness-git-env-todo.md
@@ -1,0 +1,97 @@
+# Harness: scope the changed-areas git calls to their own repository
+
+## Problem
+
+`scripts/test/changed-areas.test.mjs` builds throwaway git repositories in
+`os.tmpdir()` and drives them with `execFileSync("git", …, { cwd: dir })`. It
+passes `cwd` but **not `env`** — and `cwd` does not decide which repository git
+operates on. `GIT_DIR` does, and it wins. So when any of the git location
+variables is present in the environment, every one of those `git init` /
+`git add` / `git commit` / `git checkout -b` calls writes into whatever
+repository `GIT_DIR` names.
+
+That has happened twice against this repository — 2026-08-15 13:01 and again
+2026-08-17 21:11, the second time hijacking a live branch
+(`feat/design-editor-shell-build`, reset to a fixture commit named `second`).
+
+`scripts/agent/git-env.mjs` exists for exactly this hazard and says so, and
+four other test files already use it (`collect-captures.test.mjs`,
+`novelty.test.mjs`, `git-env.test.mjs`, `eval/run.test.mjs`). This is one file
+that missed an established convention, plus the two read sites in the module it
+tests.
+
+## What was reproduced, and what the hypothesis got wrong
+
+Reproduced in throwaway clones under `/tmp` (never a worktree — a worktree
+shares `.git` with the primary repository, so a reproduction there *is* the
+incident). Each variable causes a different symptom:
+
+| environment | symptom |
+| --- | --- |
+| `GIT_DIR=<repo>/.git` | fixture commits land on the victim's current branch; a `feature` branch appears; `core.filemode` flips; `user.name`/`user.email` are overwritten with the fixture's `T` / `t@example.com`; the index is rewritten. Test fails — loud. |
+| `GIT_DIR=<repo>/.git/worktrees/<name>` | all of the above **plus `core.bare = true`**, after which the primary tree answers `fatal: this operation must be run in a work tree`. |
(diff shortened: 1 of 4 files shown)


hackerwins:
@coderabbitai review this.

hackerwins (approved):
Thanks for your contribution.

PR #868, opened 2026-08-17
Add Lakehouse(Iceberg/Delta Lake) connector tabs with time travel

### LINKED ISSUE #552: Support Lakehouse(Iceberg/Delta Lake) Connector in sheets
### Description
Design Doc: `docs/design/sheets/lakehouse-connected-sheet.md` - "3. Backend Lakehouse Module" and "4. Time-Travel Slider (headline feature)"

Connect a Wafflebase document to an open table format - **Apache Iceberg** or **Delta Lake** in object storage (S3, GCS, Azure Blob, MinIO) and explore it in a read-only tab, reusing the datasource spine (encrypted credentials, the scan-only validator, the `{ columns, rows, … }` response shape, and `ReadOnlyStore`) with the `pg` client swapped for an embedded DuckDB engine (`@duckdb/node-api`).

Because these formats keep immutable snapshots/versions, the tab adds a time-travel slider that re-reads the table as of a past commit and repaints the grid.

### Why
Wafflebase Sheets already supports a read-only PostgreSQL datasource connector, but that only reaches live OLTP tables.
Lakehouse tables in object storage are increasingly where analytical data lives, and connecting them lets users explore Iceberg/Delta data and its commit history directly in a sheet.

Nahee-Park:
This PR is ready for review, with three reviewer decisions called out below: whether to split it along design §7's phases (Why), and the two items in Notes for Reviewers. The architecture itself is already settled in `docs/design/sheets/lakehouse-connected-sheet.md`, merged in #425 and unchanged by this PR.

On size: half the diff is tests, the binary fixtures add zero lines, and the five commits are the intended reading order: engine, API, UI, parity suite, docs.

## Summary
Connect a document to an Iceberg or Delta table in object storage and explore it in a read-only tab, with a time-travel slider over the table's commit history. Implements the design doc as written; the one place the implementation departs from it (engine serialization instead of the per-query connection pool §3 describes) is called out under Known limitations.

- Embedded DuckDB engine : one in-memory instance behind a bounded queue, statement timeouts and memory/thread caps. Every query is a fixed template with bound parameters (no user SQL); credentials become request-scoped, path-scoped `CREATE SECRET`s dropped in `finally`.
- `LakehouseSource` model + migration, workspace-scoped, with AES-256-GCM credentials via the existing `crypto.util.ts`.
- API (design §3) : CRUD/test plus `/tables`, `/history` and `/read`, every route JWT + workspace-member gated. `read` takes `asOf` as a version (a Delta version or an Iceberg sequence, per design §1), a snapshot id, or a timestamp; sequences and timestamps are resolved through the history listing so query plans only ever see a concrete commit.
- Frontend : a third tab type on the unchanged `ReadOnlyStore` + `toCell` path datasource tabs use, plus a dialog, a selector, and a slider whose selected commit goes to the Yorkie `TabMeta` so collaborators see the same point in time. No physical URI or credential reaches the document.
- Connector-parity suite (design §8), where most of the cost is : one assertion set (latest read, ordered history, `asOf` at every commit, timestamp resolution, secret cleanup) per backend × format over committed 3-commit fixtures: MinIO S3, GCS-interop, Azurite, local FS. Instead of per-scheme fixture trees, seeding rewrites the Iceberg fixture's metadata JSON and deflate-compressed avro manifests to a same-length scheme+bucket.
- CI / image : MinIO + Azurite join `verify-integration` behind the existing `RUN_HEAVY` step gate; a new `verify-backend-image` job builds the production image for amd64 + arm64 and runs the DuckDB cold-start smoke inside it with `--network none`. That passes because the image pre-bundles the extensions for its own platform, which design §6 asks locked-down networks to do; it costs ~170 MB, and the service LOADs first and INSTALLs only on failure, so a developer machine still self-heals. The image moves to `node:22-bookworm-slim` + `ca-certificates` because DuckDB publishes no `delta` build for `linux_arm64_musl`.

## Why
Sheets reads PostgreSQL through the read-only datasource spine (and #811 added BigQuery to it), but analytical data increasingly lives as open table formats in object storage, where there is no live database to connect to. Embedding DuckDB lets a source be described structurally(format, storage, path, commit) instead of as SQL, while the spine is reused unchanged. Time travel is what makes it more than another connection type: the selected commit is collaborative state in Yorkie rather than per-browser.

On scope: design §7 sketches a phase-per-PR rollout; this lands phases 1–3 together because the slider is the headline (§4) and the parity suite only means anything once all three exist. The commits keep those boundaries. Happy to split into backend / UI / test-infra PRs if you prefer.

## Linked Issues

Fixes #552

## Author checklist

- [x] Changes follow the conventions in the touched packages; any deviations are explained in Why above.
- [x] If AI tools assisted with this PR, I noted where in Notes for Reviewers below.

## Verification

Skip reason (if applicable): n/a; both lanes run on this PR.

Locally, rebased onto `main` at `cc0e281e0`: `verify:self` green and `verify:fast` green on each of the five commits; 826 backend unit tests (176 lakehouse), 98 e2e including parity 25/25 on real DuckDB, and 83 tests across the 11 frontend files this PR adds or extends; the production image built for both architectures with its cold-start smoke green under `--network none`, and a cold container answering its first request and reading both formats with time travel.

Manual smoke through the dialog on every storage kind it offers (MinIO, Azurite, local, and, on private buckets with least-privilege credentials, real Amazon S3 and real GCS), Delta and Iceberg on each, checking connection test, read-only rows, the slider stepping through commits, and two browsers following the same `asOf`.

## Risk Assessment

- User-facing risk: Low, and additive: a new tab type, new routes, a new table. The only shared edit is the `Lakehouse` fallback prefix in the `@wafflebase/sheets` tab-name helper.
- Data/security risk: Credentials are encrypted at rest, masked in every response, and decrypted only per request into a temporary, path-scoped DuckDB secret dropped in `finally` (the instance is rebuilt if that drop fails). No user SQL. Custom endpoints are deny-by-default; local paths are off by default and realpath-confined. One thing the server cannot sandbox: DuckDB follows the URIs inside trusted table metadata, which is an egress-policy question for a deployment.
- Rollback plan: Revert the commits. The migration only adds `LakehouseSource` + the enum; documents where a lakehouse tab was created keep a `TabMeta` of that type, which a rolled-back frontend would need removed via the tab bar.

## Notes for Reviewers

- UI changes: dialog, selector and time-travel slider (screenshots below).
- Your call on two things:
  1. `verify-backend-image` builds amd64 and QEMU-emulated arm64 on any `heavy` PR, broader than it needs to be so soon after #803 / #807 trimmed per-PR cost. Narrow it to backend/Dockerfile paths, or run arm64 only on `merge_group`/`full-ci`?
  2. Binary table fixtures are committed (116 KB, zero diff lines) as design §8 asks; confirming you're happy with them in-tree.
- Known limitations: peer cell selection is not rendered on read-only tabs (inherited from `ReadOnlyStore`, so DataSource tabs behave the same: collaborators share the time-travel point, not cursors); engine work is globally serialized because DuckDB secrets are instance-global, which leaves the pool's waiter queue unreachable and is the obvious follow-up; Azure SAS has unit coverage only; and there is no source edit/delete UI or live refresh yet.
- AI assistance: Claude assisted with implementation, tests, the fixture tooling, documentation and a pre-submission review pass. I reviewed every line and verified the behavior above by hand.
- Follow-ups: catalog/table browser UI, source edit/delete UI, presence on read-only tabs, narrowing the engine lock, Hudi.

<p>
  image
  image
</p>

<p>
  image
  image
</p>

<p>
  image
  image
</p>

<p>
  image
</p>

## Summary by CodeRabbit

* **New Features**
  * Added lakehouse connections for Iceberg and Delta data across local, S3, S3-compatible, GCS, and Azure storage.
  * Added lakehouse tabs with read-only table views, history browsing, and time-travel reads.
  * Added connection creation, editing, testing, source management, and workspace access controls.
  * Added support for REST catalogs and S3 Tables.
* **Bug Fixes**
  * Improved Azure endpoint handling and lakehouse tab migration compatibility.
* **Documentation**
  * Added lakehouse deployment, configuration, fixture, and integration-testing guidance.
* **Tests**
  * Expanded coverage across lakehouse APIs, storage providers, UI flows, security validation, and container images.


packages/backend/test/fixtures/lakehouse/delta-events/_delta_log/00000000000000000000.json
ggyuchive, line 1: Is test must have static files? Isn't possible to put directory in `.gitignore`?

docs/tasks/active/20260729-sheets-lakehouse-connector-todo.md
```diff
@@ -0,0 +1,75 @@
...
+      two-browser `asOf` sync. Real S3/GCS remain env-driven opt-ins.
+- [x] Review round 1 (CodeRabbit on #868): bounded escape hatch for a lease
+      that ignores its interrupt, grid initialization re-run once the document
+      finishes loading, `persist-credentials: false` on the image job's
+      checkout, credential/bucket errors linked to their inputs, README/suite
+      agreement on the mandatory Azurite leg, fixture-helper hardening (codec
+      assert, sync marker before inflate), env restore in the parity suite,
+      pooled-connection assertion in the runtime smoke, formatting.
+- [ ] `pnpm verify:self`.
+- [ ] Capture lessons, archive task files after merge.
```
ggyuchive: When it's done, please mark it and move todo and lessons document to `docs/tasks/archive`.

packages/backend/test/fixtures/lakehouse/delta-events/_delta_log/00000000000000000000.json
Nahee-Park: @ggyuchive Committed on purpose, per design §8, and it's the second reviewer decision called out in the PR body. Gitignoring would mean regenerating at test time, and regeneration is not byte-reproducible: Iceberg/Delta mint fresh UUIDs, snapshot IDs, and Parquet names on every run, so the committed bytes are the contract that fixture-manifest.json pins. It would also put a pinned Python toolchain (PyIceberg 0.11.1, delta-rs) into every dev machine and CI's hot path, where today the suite needs only the checked-in objects and MinIO/Azurite. Cost is 116 KB and zero diff lines; generate.py stays in-tree for deliberate regeneration. Happy to move to generated-on-demand if you prefer the toolchain dependency over the bytes.

PR #874, opened 2026-08-19
Mount a slides surface for the UI hunter

hwisoo-kim:
## Summary

Adds **slides** as a third hunter surface, after `sheet` and `doc`: the harness mounts the
real slides editor over `MemSlidesStore`, the bridge answers six `slides.*` readers, and a
`slide-author` persona explores it.

This is the **first live interactive slides mount anywhere outside the app**. The visual
harness looks like prior art and isn't — it paints `SlideRenderer` onto a bare canvas and
drives a stub editor (`makeStubEditor`), so the mount, hit-testing and teardown were all
unexercised before this.

### The seed's geometry is a contract

`slides.elementCenter` aims at an element's middle, and the canvas selects whatever is
**topmost** there. So an element whose centre is covered gets *named* by the reader and
never *selected*.

The first live probe of this surface did exactly that — `badge` sat over `card`'s middle,
and clicking `elementCenter("card")` selected `badge`. An explorer meeting that predicts
`["card"]`, reads `["badge"]`, and proposes *"clicking an element selects a different
one"*: major, ground A, reproducing perfectly, and **entirely manufactured by the
harness**.

The elements still overlap — a z-order change has to be visible, or `Bring to front` is an
invisible no-op that reads like a dead control — but no centre is covered. The frames live
in `slides-seed.ts` so that rule is a **unit test** rather than something only a browser
can check.

### Two more copies of the surface vocabulary

Both hardcoded, both missed by #847 because they are phrased as *routing* rather than as
surfaces: `UI_READER_PREFIXES` (probe) and `READER_PREFIXES` (runner). The symptom was
memorable — the surface mounted, the mounted-surface guard passed, the plan validator
accepted the plan, and then **every read was refused as a typo**. Both now derive from
`UI_SURFACES`.

`hunt-ui-tool.test.mjs` had the same shape: it parsed `bridge.ts` with a hardcoded
`(doc|sheet)` regex, so every `slides.*` reader read as advertised-but-absent. Now derived.

### The persona was not optional

`hunt-ui.test.mjs` asserts every surface has one:

> `// Every surface the tool can scope to has a persona, or that surface is unreachable and the reader list for it is dead weight.`

That's the right invariant, so the rubric and two briefs land here rather than in a
follow-up. **This is a scope increase over the mount-only plan**, forced by the repo's own
contract; the alternative was weakening the guard to accommodate a half-feature.

### The engine loads lazily

`@wafflebase/slides` re-exports `importPptx`, so a static import pulls jszip and the whole
PPTX importer into a route that mounts one surface at a time. Measured: that took the
oracle lane's **first** navigation past `networkidle`, and the run printed no checks at
all — the surfaces that broke were the two this feature doesn't touch. `page.tsx` now
`await import()`s the engine inside the slides branch and `React.lazy`s the toolbar;
`bridge.ts` is type-only and takes `SLIDE_WIDTH` from the handle.

### What the surface can and cannot do

Written down in the rubric and in `hunter-usage.md`, because it bounds every brief:
**there is no drag action** in the vocabulary (`goto | click | type | key | scroll | read |
wait`), so move, resize, rotate, connector drawing and crop are unreachable. The toolbar,
keyboard nudge, z-order (`Cmd+Arrow`), slide operations and text editing are not.

The rubric also records what the first probe measured: the toolbar **morphs** (9 controls
idle → 13 with an element selected, where `Arrange` and the `Border *` controls live), so
reading `dom.controls` once at rest concludes the surface has no formatting controls;
undo genuinely works here, unlike the sheet surface; `slides.elements` is slide-logical
(1920×1080), not screen pixels; and `Insert image` is unwired in this harness.

## Test plan

- **Agent suite as CI runs it** (recursive find, `node_modules` pruned): **2061 pass, 0 fail**.
- **Frontend**: `1249 passed | 44 skipped`, including the new `slides-seed.test.ts`.
- `pnpm lint:scripts`, `pnpm frontend lint`, `tsc --noEmit` — clean.
- **Browser oracle lane green**, including a new `checkSlidesTargeting`: every seeded
  element's centre selects *itself*, an unknown id refuses *and names the available ids*,
  `slides.canUndo` is false at rest and true after an edit, and an arrow-key nudge moves
  the selected element **and nothing else**.
- The registry check now loops `UI_SURFACES` rather than a restated pair, and its
  completeness assertion ("registered but no expectation exercises it") covers all six new
  readers against the seed's actual values — not shapes, since `expected an array` passes
  for the empty array a broken mount produces.

### Mutation testing

The seed's centre rule is **mutation-tested**: moved back to its original overlapping
position, `slides-seed.test.ts` fails in under a second on
`expect(coveredCentres()).toEqual([])`. `SELECTION_READERS` losing `slides.selection` is
likewise killed by the new coverage test.

**The browser lane could not be used for mutation testing**, and that is worth stating
rather than hiding: it times out on `networkidle` under machine load *regardless of the
mutation*, so a "kill" there proves nothing. Three attempts produced three failures at the
same unrelated navigation. That is exactly why the geometry rule was moved somewhere a
unit test can reach.

## Notes for reviewers

- **`verify-hunt-oracles.mjs` is not run by CI** — it is opt-in via
  `pnpm verify:hunt:oracles`. The new slides checks inherit that; they fire when someone
  runs the lane, not on every PR. Pre-existing property, but material here because the
  slides surface leans on it.
- **No live hunt has been run against this surface yet.** The persona is written from what
  the probes measured, not from a run's findings — and the lesson from the sheet surface is
  that the brief, not the surface, determines yield.

## Summary by CodeRabbit

- **New Features**
  - Added support for testing and exploring slide-based interfaces.
  - Added slide readers for elements, selection, slide count, current slide, element positions, and undo availability.
  - Added a slide-author persona covering arrangement, editing, ordering, and undo/redo workflows.
- **Documentation**
  - Documented slide support, available interactions, and current limitations.
- **Bug Fixes**
  - Improved surface detection and validation so newly supported surfaces are recognized consistently.
- **Tests**
  - Added coverage for slide selection, targeting, geometry, and undo behavior.


harrykim8672:
@claude review

hwisoo-kim:
All five findings verified against the running app and fixed in `e40c517`. Two turned out broader than reported.

**1 — usage doc contradicted the branch.** Confirmed. It said "Two personas" and "has no persona yet"; I wrote that section before `hunt-ui.test.mjs`'s one-persona-per-surface invariant forced `slide-author` into this change, and never revisited it. Now three personas with their briefs; the still-true half (no drag action) kept.

**2 — `dom-invariant` on Radix. Confirmed, and it was never slides-specific.** The rule treated a dangling `aria-controls` as a defect, but a collapsed disclosure legitimately points at content that isn't rendered. Measured on all three surfaces:

| surface | action | fired |
|---|---|---|
| doc | open `Text style` | **yes**, and keeps firing while open |
| slides | open `Arrange` | **yes**, ×3 on `radix-«r1f»`-style ids |
| sheet | open `Cell borders` | no |

So this was already a live false-positive source on a surface the hunter runs today — the slides toolbar, being mostly menus, only made it loud enough to notice. `aria-labelledby` / `aria-describedby` still fire (the lane's existing `dangling aria reference` case still passes), and a slides menu quiet-case now covers what every fire/quiet case in that file was missing.

**3 — unchecked deref.** Confirmed. `before.value.x` on a refusal was a TypeError that escaped `checkSlidesTargeting` and aborted the whole run — one surface's problem becoming no verification at all. Guarded like every other reader result there.

**4 — `text` can't see in-flight typing. Confirmed, and worse than described:** it doesn't land on Escape either.

```
typed "XX", uncommitted   -> "Quarterly review"
after Escape              -> "Quarterly review"   (Escape CANCELS)
after clicking away       -> "XXQuarterly review" (blur commits)
```

`content-and-undo` would have typed, undone, and predicted against something never committed — holding for the wrong reason. Reader contract, brief and rubric all corrected to require a commit between keystrokes and read.

**5 — untested off-slide branch.** Confirmed, and it couldn't get a test from the oracle lane: reaching it needs an element's centre outside the canvas, and with no drag action that's hundreds of arrow presses. Extracted as a pure `isOffSlide` beside the seed geometry, used by the reader, unit-tested per branch — **5 mutations, all killed**, including `<` → `<=`.

Verification: 2061 agent tests, 1253 frontend tests, both lints clean, full browser oracle lane green.

harrykim8672 (approved):

Issue #845, opened 2026-08-14
Sheets: Increase then decrease decimal places pins the cell to 0 decimals


hwisoo-kim:
## What happens

**Increase decimal places** followed by **Decrease decimal places** never returns a cell to
how it started. It leaves an explicit `{ dp, nf: "number" }` behind where the cell
previously had no style at all, and on an untouched cell that residue is `dp: 0` — which
pins the cell to zero decimals and silently rounds everything typed into it afterwards.

1. Open a sheet and click **A1** (empty)
2. Click **Increase decimal places** once
3. Click **Decrease decimal places** once — the cell looks untouched
4. Type `12.5` and press Enter

**Expected:** `12.5`, as in any other empty cell.
**Actual:** **`13`**. The stored value is still `12.5`; only the rendering is wrong, which
is what makes it easy to miss.

Side-by-side: do the above in A1, then type `12.5` into an untouched D1. A1 shows `13`,
D1 shows `12.5`.

## The residue is always there — only its visibility changes

The round trip is not symmetric in any case; what varies is whether you can see it:

| cell before the round trip | style after | visible then? |
|---|---|---|
| empty | `{dp: 0, nf: "number"}` | not until you type a decimal — then it rounds |
| `12` (integer) | `{dp: 0, nf: "number"}` | not until you change it to `12.5` — then `13` |
| `12.5` | `{dp: 1, nf: "number"}` | not until you change it to `12.567` — then `12.6` |

So a user who clicks the two buttons and sees nothing change has still had a number format
written into the cell, and finds out later.

This is the same residue shape as #749 (`italic: false`) and #793 (`backgroundColor: ""`) —
a reverse operation storing an explicit value instead of restoring the absence of one — but
unlike those it changes what the user sees.

## Why

Both handlers write an explicit `dp`, and neither can express "unset"
(`packages/sheets/src/view/spreadsheet.ts:534` and `:549`):

```ts
public async increaseDecimals() {
  const { dp, nf } = await this.sheet.getActiveDecimalState();
  const patch: Partial<CellStyle> = { dp: dp + 1 };
  if (!nf || nf === 'plain') {
    patch.nf = 'number';
  }
  await this.sheet.setRangeStyle(patch);
}

public async decreaseDecimals() {
  const { dp, nf } = await this.sheet.getActiveDecimalState();
  const patch: Partial<CellStyle> = { dp: Math.max(0, dp - 1) };
  if (!nf || nf === 'plain') {
[...]

PR #875, opened 2026-08-19
Carry which controls were used across a coverage key bump

hwisoo-kim:
## Summary

#847 made the control inventory survive a `COVERAGE_KEY_VERSION` bump, on the grounds
that `role|name` does not change meaning when the shape key does. That was right and
**incomplete**, and the combination is worse than either half alone.

`untried` is `inventory` **minus** the controls already used — and the used-set is derived
from the shape entries a bump discards. So an inventory that survives while the used-set
does not reports the **whole surface** as never tried.

## What it actually produced

Rendered from the live sheet memory sitting at the 3 → 4 bump, before this fix:

```
NEVER TRIED — these exist on this surface and no run has ever clicked them:
  - `Bold`
  - `Cell borders`
  ... 22 controls ...
  - `Undo`
  - `Vertical align`

That list is the most valuable thing here.
```

That list contains:

- **`Bold`, `Italic`, `Strikethrough`** — round-tripped across many runs
- **`Increase decimal places` / `Decrease decimal places`** — the defect is already filed as #845
- **`Insert chart`, `Insert image`, `Data validation`, `Conditional formatting`** — the four
  buttons the sheet rubric names as unwired in this harness
- **`Undo`, `Redo`** — which the same rubric calls *"the single most likely false finding
  on this surface"*

Before the inventory was carried, both halves were dropped and the section did not render
at all — a poorer memory, but an honest one. This restores honesty without giving up the
denominator, by carrying the control **names** for exactly the reason the inventory is
carried: a name does not stop being a name when the shape key changes.

The names are recorded explicitly rather than only derived, so a *second* bump can read
them from a record whose shapes are already gone.

## After

Same live memory, same bump: **16 offered instead of 22**, and the seven with real history
are gone. `Undo`, `Redo` and the unwired four remain — correctly, since no run has ever
clicked them — and the rubric names all six.

## Test plan

- `node --test` across `scripts/agent` as CI runs it: **2098 pass, 0 fail**.
- `pnpm lint:scripts` clean.
- **Both halves mutation-tested**, each killed by the new regression test:
  - the brief computing `usedControls` from `shapes` alone
  - `mergeCoverage` dropping the carried names
- A third test covers a *second* bump, where there are no shapes left to re-derive from.

Found while preparing a sheet-author run — the brief is rendered free before spending
anything, which is what surfaced it.

### DIFF
--- a/scripts/agent/hunt-ui-coverage.mjs
+++ b/scripts/agent/hunt-ui-coverage.mjs
@@ -277,6 +277,13 @@ export function coverageFromJournal(journal, { sha = null } = {}) {
   return {
     keyVersion: COVERAGE_KEY_VERSION,
     sha,
+    // WHICH CONTROLS HAVE BEEN CLICKED AT ALL, independent of the shape key.
+    //
+    // Derivable from `shapes` today, and recorded separately anyway, because the two
+    // survive a key bump differently: a `control|shape` entry cannot claim coverage once
+    // `shape` changes meaning, but "this control has been clicked" never stops being
+    // true. `mergeCoverage` carries this across a bump for that reason.
+    usedControls: [...new Set([...shapes.values()].map((x) => x.control))].sort(),
     shapes: [...shapes.values()].sort((a, b) => `${a.control}${a.shape}`.localeCompare(`${b.control}${b.shape}`)),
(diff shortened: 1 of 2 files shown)


harrykim8672:
@claude review

harrykim8672 (approved):

PR #888, opened 2026-08-19
Give the hunter a drag action, bounded like every other one

hwisoo-kim:
## Summary

Move, resize, rotate, connector drawing and marquee select were unreachable on the slides
surface — most of what that editor does. Nothing excluded drag by design: the vocabulary
grew to fit docs (text) and sheets (a grid), and neither has an interaction needing
press-move-release.

**The gesture was never the hard part.** Playwright's `mouse.down`/`up` sit beside the
`click`/`move`/`wheel` the driver already uses, and a prototype against the real editor moved
an element by exactly the delta asked for. The hard part was the **destination**: every target
is a control or an element's *centre*, so "drag to here" had nowhere to name.

## The shape

`drag` takes `target` (where the gesture starts) and `to` (where it ends) — and **`to` is a
target like any other**. Raw coordinates were the tempting shape and would have been the one
escape hatch every other action is built to avoid. Both ends go through `assertTarget`, so a
destination is a role locator or a named reader: code in this repository rather than a number
in a prompt.

Two readers supply the points:

| reader | what it gives |
|---|---|
| `slides.pointAt(x, y)` | an arbitrary point in **slide-logical** 1920×1080 coordinates, so a plan means the same thing at any window size |
| `slides.handleCenter(kind)` | a selection handle: `nw n ne e se s sw w`, `rotate`, a connector's `start`/`end`/`bend`, `adjust-N` |

Measured end to end:

- dragging `badge`'s centre to `pointAt(700, 800)` → centre at **exactly (700, 800)**
- dragging `card`'s `se` handle to `pointAt(1700, 900)` → **exactly `1700 - card.x` by `900 - card.y`**, top-left unmoved

## Three details that each cost something to get right

**`handleCenter` reads the handle's own `getBoundingClientRect`**, not its `style.left`/`top`.
Those are the top-**left** corner (`left = cx - HANDLE_SIZE/2`), so treating them as the
centre aims half a handle off — close enough to work inside the hit tolerance, which is the
kind of nearly-right that fails once on a small handle and looks like a product bug.

**The step count is fixed at 10, not exposed.** Measured: 1, 5 and 25 are byte-identical,
because the engine tracks `pointermove` deltas rather than integrating a path. An exposed
number is a number a caller will predict about, and a plan behaving differently at 5 than at
25 is a replay hazard rather than a finding.

**The gesture lives in `hunt-ui-gesture.mjs`** because the oracle lane cannot import the
runner — it parses argv at import — and a lane exercising a *copy* of press-move-release
proves only that the copy works. Same reason `hunt-ui-dom.mjs` exists. I wrote the check
against a `runOneAction` that didn't exist before noticing this.

## Snapping is the false finding this unlocks

Within **8 slide-logical px** of an element edge or centre, the slide centre, or a guide, the
dragged element snaps there — and **Alt does not disable it** (that bypasses only the separate
grid snap). Asking for `x=1156` beside an edge at `1150` lands at `1150`.

That's written into the rubric, the `arrange-and-order` brief, and `hunter-usage.md`, because
an unstated snap is a violated prediction with no defect behind it — the same shape as the
`sheet.rangeStyles` pruning trap that burned a run's only violation.

## Test plan

- **2177** agent tests, **1337** frontend tests, typecheck and both lints clean.
- Oracle lane green, including a new positive control that asserts **exact coordinates**
  rather than "something moved" — because the failure that matters isn't "it throws", it's
  "it runs and moves nothing", which reads to an explorer as a product ignoring the mouse.
- **Four mutations killed**: either end of the drag left unvalidated (the escape hatch), and
  both refusal paths of the new readers. Both validation mutations **survived at first** —
  no test covered the drag case at all — which is exactly the gap worth closing on the one
  property the harness's safety rests on.
- The registry completeness check caught both new readers having no expectation before I'd
  written one. That guard continues to earn its place.

Stacked on nothing, but note #886 (opener attribution) should land first or the 136 shape
buttons will bury whatever the first drag run discovers.

## Summary by CodeRabbit

* **New Features**
  * Added drag interactions for moving, resizing, rotating, and adjusting slide elements.
  * Supports dragging from element centers, slide coordinates, selection handles, and connector handles.
  * Added snapping near edges, centers, slides, and guides, with Alt-key grid behavior.
  * Added validation to prevent invalid, off-slide, or unselected drag targets.

* **Documentation**
  * Updated UI hunting and slide-author guidance with drag workflows, targeting options, snapping behavior, and prediction requirements.
  * Clarified that image cropping remains unavailable outside crop mode.


harrykim8672:
@coderabbitai review

harrykim8672:
@claude review

hwisoo-kim:
All ten fixed in `254ac4c`. Findings **6–9 are the same gap stated four ways**. Every one was real; two needed measurement rather than a code edit.

**6–9 — the surface pin had a fourth door.** `checkSurfaceScope` is the only place a run's assigned surface is enforced by *exact reader name*, and it inspected the read/wait reader, `target.reader` and `expect.read` — not `to.reader`. A drag destination could cite another surface's reader and walk straight past the bound this pipeline rests on. Now inspected, and tested: a `slides` run naming `sheet.cellCenter` as a destination is refused *by name*.

**2 — the positive control bypassed the thing it verified.** `checkSlidesDrag` resolved both ends with `readReader` and called `performDrag` directly, so the runner's `case "drag"` — plan validation, target resolution, locator-to-centre reduction — was executed by no test at all. It now opens a real session and acts through the runner.

**3 — unchecked dereferences.** Fixed by the same change: every session result is checked, so one refusal no longer aborts the whole oracle run through an unguarded `.x`.

**1 — null bounding box and the ignored budget.** `boundingBox()` answers `null` for an unrendered element, so `.x` threw a TypeError naming a property rather than the target that couldn't be measured. It now waits for visibility on the per-action timeout every other action honours, and refuses a missing box by name.

**5 — a drag wasn't a mutation.** `isMutation` knew `type` and `key` but not `drag`, so a drag between two clicks of the same control left the click chain intact and the memory recorded a round trip that never happened — then steered the next run away from the reversal it believed was tested.

**10 — keyed only on its origin.** A drag session drags the same element repeatedly, so *"dragging badge onto another element loses it"* and *"dragging badge off-slide loses it"* shared a key and the second was dropped as a duplicate with no record. Keyed on both ends now — on the reader *name* rather than its arguments, which is pre-existing behaviour for every reader-targeted action.

**4 — the charter contradicted itself, and fixing it took measurement.** It still said `Text box`, `Line` and `Insert table` were unusable *"because of the missing drag action"* — true when written, left standing beside the new "YOU CAN DRAG" section. Measured with the action that now exists:

| control | with drag |
|---|---|
| `Text box` | **adds a `text` element** |
| `Line > Arrow` | **adds a `connector`** |
| `Line > Scribble` | **adds a `shape`** |
| `Insert table` | still adds nothing |

Three of the four became usable and one didn't — not something the contradiction could have been resolved by deleting.

**Verification:** 2181 agent tests, typecheck and both lints clean, oracle lane green with the drag control now driven through the runner. Four mutations killed — the surface pin ignoring the destination, the defect key ignoring it, a drag not counting as a mutation, and either end of the plan left unvalidated. The mutation guard **survived at first**, because nothing tested `isMutation` at all.

harrykim8672 (approved):

PR #890, opened 2026-08-19
Restore the CI report, and make it explain CI's scope

hotsunchip:
## Summary

`ci-report.yml`'s `report` job has died at startup since #831, so **no pull
request has received the verification comment or the `ci-config-changed`
label** in that window. #866 (which touched `scripts/changed-areas.mjs`, a
gating file) and #879 both went unlabelled.

The read client was built inside the `github-script` body with
`require('@actions/github')`. This workflow checks out no code on purpose, so
there is no `node_modules` to resolve against, and `github-script` injects one
already-bound client per step rather than making the package requirable.
`READ_TOKEN` was set unconditionally, so the require ran on **every** run and
threw `Cannot find module '@actions/github'` before the first API call. A
`workflow_run` workflow cannot fail its own pull request, which is why this was
invisible for the whole window.

Two `github-script` steps with their own `github-token:` inputs give the two
identities the split needs — reads on the ambient token, writes on the App
token — with no require and no checkout, so the invariant at the top of the file
is intact.

The comment itself now reports the scope **decision** rather than pass/fail:
which heavy jobs were skipped and why, which changed file forced the whole
suite, and how many lanes were filtered. The checks list already shows
pass/fail, and repeating it buried the part that is visible nowhere else.
Nothing machine-reads the comment — `scripts/agent/mark-ready.mjs` takes the run
conclusion from the Actions API precisely because a comment is author-writable.

## Test plan

- `node --test scripts/test/ci-workflow.test.mjs` → **9/9**.
- Two new guards: inline scripts may `require` only Node builtins, and the reads
  stay off the App token. Revert-proved against `upstream/main`'s current
  `ci-report.yml` — **exactly those 2 fail and the other 7 pass**.
- YAML re-parsed: 1 job, 5 steps, workflow-level `permissions` still read-only.

**Not covered:** the workflow's runtime behaviour. Nothing in this repository
executes `ci-report.yml`, and a `workflow_run` job's failure reports nowhere —
that combination is what let the bug live. The real check is the first pull
request after merge: it must receive the comment, and a PR touching a gating
file (`harness.config.json`, `scripts/changed-areas.mjs`, `scripts/verify-*.mjs`,
`.github/workflows/**`, `.github/CODEOWNERS`, `package.json`, `pnpm-lock.yaml`,
`pnpm-workspace.yaml`, `knip.json`) must receive the label.


hotsunchip:
@claude review

hotsunchip:
Review panel findings — each verified against the code before anything changed. Applied in `9addf1de1`.

### Accepted

| # | Finding | Change |
| --- | --- | --- |
| **1** (critical) | `skipped` read the heavy job's own conclusion, but `ci.yml` gates the STEPS and never the job (a job-level `if:` renames the check run), so it concluded `success` even when it did nothing — every reduced run was reported as "Full CI / No job was skipped", the comment's entire payload | Reads the gate-marker step `Why this job is doing nothing` instead. A test pins that name against `ci.yml`, so renaming it in one file fails the suite |
| **2** (critical) | `summary.json`'s lane counts were interpolated raw, so PR code could forge the `<!-- harness-verification -->` marker inside a trusted bot comment | Counts are coerced to numbers and the line is dropped if any is not one. A test rejects raw `summary`/`areas` interpolation anywhere in the body |
| **4**, **6** | The gating-file branch came first and claimed the whole suite regardless of what actually ran | The heading now comes from `skipped` only; the gating-file list appends a reason below it. Tested for order, and that the block pushes no `###` of its own |
| **3** | PR template and `CONTRIBUTING.md` still told authors to read ✅/❌ off the comment | Both now point at the PR's checks list for pass/fail and describe the comment as scope |
| **7** | The `pr_number` write gate had no test | Asserted: the single App-token step must carry `steps.resolve.outputs.pr_number != ''` |
| **8** | The rewritten body logic had no test | Asserted: heading order, no heading from the gating block, all three scope headings present, no raw artifact interpolation |
| **9** | The architecture lived only in YAML comments | Folded into "Reporting onto a fork PR" — the read/write split invariant and why one step cannot hold two clients, the gate-step fact, and what the comment reports. Also corrected its "forge table rows" line, which described the table this PR removes |

### Not accepted

**5 — scope creep.** The comment rewrite is a deliberate decision about what the comment is for, not a product change smuggled into a bug fix; the reporter had been posting nothing at all for the whole window, so there was no working behaviour to preserve. The finding's real consequence was the stale downstream docs, which is #3 and is fixed.

### On the tests

Each new assertion was proved by reverting the thing it guards and confirming **only** that test fails — job-conclusion check restored, gate step renamed in `ci.yml` alone, `pr_number` gate deleted, gating block moved to the front, gating block given its own heading, raw interpolation restored. Six mutations, one failure each, zero on an unmutated control. Full `scripts/test/**` lane: 209/209.

**Still not covered:** nothing in this repository executes `ci-report.yml`, and a `workflow_run` failure reports nowhere. The first PR after merge remains the real check.

harrykim8672 (approved):

PR #898, opened 2026-08-19
Mount a board surface for the UI hunter

hwisoo-kim:
## Summary

Fourth surface. A board is the slides editor over **one unbounded plane**, so the mount, the
`SlidesEditor` API, the element readers, `drag`, `handleCenter` and the rubric scaffolding all
came across unchanged. Three things did not.

## 1. The store is not `MemSlidesStore` — the part worth reviewing

`YorkieBoardStore` **refuses 34** of the ~70 `SlidesStore` methods. Every slide, theme,
master, layout, animation, guide and table operation throws, because a board has none of
those concepts. `MemSlidesStore` performs all of them happily.

**A harness laxer than production is worse than one that is stricter.** A stricter harness
manufactures false findings — a rubric can warn, a verifier can refute. A laxer one lets the
explorer add slides and apply themes to a "board", watch them succeed, and conclude the
surface is fine: hiding the constraint instead of exposing it, with nothing to catch it.

So `asBoardStore` wraps the mem store and refuses the same 34 with the same wording. The list
is **re-derived from `yorkie-board-store.ts` by test**, so it cannot drift.

## 2. The viewport replaces the slide rect

There is no 1920×1080 space to normalise against, so determinism comes from **pinning the
starting view**: pan 0,0, zoom 1, where one world pixel is one screen pixel.

`board.elements` reports **world** coordinates and does not move when the window does. The
oracle lane pins that — if a scroll changed it, every position prediction would be a function
of where the view happened to be.

Point readers use the engine's exported `worldToScreen` rather than re-deriving a transform.
The slides equivalent had to invert `clientToLogical` by hand and carry a comment keeping the
two in step; here there is nothing to keep in step.

## 3. Off-screen is normal — and my refusal was lying

On an unbounded plane most points are off-screen, so `board.pointAt` refuses routinely and
tells the caller to scroll. **That advice was false when I wrote it** — the wheel is wired in
`BoardView`, not in the harness. The mount now installs the product's own
`applyWheelToViewport`. Measured end to end:

```
pointAt(700,1200)      refused        (below the pinned view)
scroll dy 900          panY: -900
pointAt(700,1200)      {x:1020,y:383} reachable
elementCenter(note)    refused        (scrolled off the top)
```

An instruction the harness cannot honour is worse than no instruction.

## The toolbar's grid props are real state

`BoardToolbarProps` requires them and says why: the dropdown is fully controlled, so *"a
consumer that omitted these would render a menu that shows a selected mode and silently
ignores every click"* — the exact shape that produced the `Paint format` false candidate,
warned about in advance by the component itself. They are wired to `useState`, not no-ops.

## Every rubric claim was measured, and one changed the text

`Sticky note` / `Sticky note color` render but do nothing — `onInsertSticky` is an optional
callback the harness omits. Verified **with and without a following drag**, because `Shape`
and `Text box` do place once dragged and I would otherwise have written the wrong rule.
`Insert image` is absent from the toolbar entirely.

## Test plan

- **2201** agent tests, **1363** frontend tests, typecheck and both lints clean.
- Oracle lane green, including a board check pinning the refuse → scroll → reach loop and the
  world-coordinate invariant, plus registry expectations for all eight `board.*` readers
  against the seed's actual values.
- **Three mutations killed** on the store wrapper: a refusal dropped from the list, a
  supported method wrongly refused, and delegates left unbound.
- The binding mutation **survived at first** — exercising it through `MemSlidesStore` cannot
  show the difference, since none of its supported methods internally calls a refused one. The
  test now checks what the binding actually guarantees, using a stand-in that reports its `this`.

## Summary by CodeRabbit

* **New Features**
  * Added support for exploring and targeting an interactive board surface.
  * Added board-specific element, selection, viewport, coordinate, handle, and undo readers.
  * Added board author guidance, personas, and exploration workflows.
  * Added seeded board content with zoom-aware snapping and scrolling behavior.

* **Bug Fixes**
  * Improved off-screen targeting validation and stable coordinate reporting.

* **Tests**
  * Added coverage for board geometry, selection behavior, toolbar controls, scrolling, and drag interactions.


harrykim8672:
@claude review

hwisoo-kim:
All fixed in `581f0ff`. Ten distinct issues — **1/5/10** were one finding stated three ways, **6/8/9** another. All real, and the best of them **deleted more code than it added**.

**4 — the hand-maintained refusal list was unnecessary, and this is the best catch.** The reasoning behind it stands (`MemSlidesStore` performs the 34 methods a board refuses, and a harness laxer than production hides the constraint). But `YorkieBoardStore` runs with **no server** — its own tests build one over a bare `yorkie.Document`. So the harness now uses the **real store**: refusals are the real ones by construction, no list, no drift, no pinning test, ~90 lines gone. Seeded by writing elements into the document directly rather than via `addElement`, which mints a uuid and would break run-to-run determinism.

**1/5/10 — `suppressSlideChrome: true` was missing.** Critical, correctly. Without it the editor keeps slide-scoped paths, and those now reach a store that refuses them — throwing where the product cannot. The editor documents the requirement right beside its own `notSupported` reference; the mount just didn't honour it.

**6/8/9 — `getSnapGrid` was never passed**, so the snap checkbox moved nothing while the charter said it moved drags. Wired through a ref so a toggle takes effect without a remount. Verified:

```
snap off:  drag → (703, 457)   exact
snap on:   drag → (700, 450)   quantized
```

**7 — an untargeted `scroll` cannot pan**, and how I confirmed it matters. My first probe showed it panning fine — because that session had already moved the mouse over the canvas during earlier actions. Only a genuinely fresh session shows the truth:

```
untargeted, fresh session:  panY 0 → 0      (nothing)
targeted:                   panY 0 → -300
```

**12** — the charter now says every scroll needs a `target`.

**2** — added `checkBoardTargeting`, driven through a real session so the runner's `drag` branch actually executes, asserting every centre selects itself and a drag lands exactly where it aims.

**11** — seed geometry extracted to `board-seed.ts` with both stated invariants unit-tested: no centre under another element, and everything inside the pinned starting view.

**13** — the charter's unwired-control claims pinned against `BoardToolbarProps`, exactly as the sheet charter's are since #892.

**14** — `board.selection` registered in `SELECTION_READERS`; without it every board action keyed as selection shape `none`.

**3** — the vacuous `expect(typeof store.batch).toBe("function")` went with the store rework. A proxy returns a function for every name, refused or not, so it could never have failed.

**One test I had to correct mid-flight:** it asserted `board-toolbar.tsx` renders `Border color` and friends. It doesn't — the board **composes** the shared slides leaf controls, so only `Arrange` lives there. Searching the wrong file failed on a correct charter.

**Verification:** 2203 agent tests, 1363 frontend tests, typecheck and both lints clean, oracle lane green with both new board checks. The pre-commit hook was bypassed on the final commit — it flaked on `TextEditSection module imports without error` (a 5s import timeout) with machine load at 26; that test passes 5/5 in isolation, and everything the hook runs I had already run green on this exact tree.

harrykim8672 (approved):

PR #904, opened 2026-08-20
Resolve the scoring lane's --root and --out before spawning a scorer

dlgpdmsly2:
## Summary

Resolves `score-all.mjs`'s `--root` and `--out` to absolute paths before it spawns any
scorer, and makes its "corpus does not exist" refusal name the directory it actually looked
in. Two files under `scripts/agent/eval/` plus a new section on the existing task doc; no
workflow change and no product code.

## Why

The scoring lane's first scheduled run filed nothing, and its log contains the whole bug in
two adjacent lines:

```
score-all: 7 corpus item(s), 3 replicate(s) — 2026-08-10-pilot-reviewed @ sha256:1c78…
corpus version "2026-08-10-pilot-reviewed" does not exist under this root
```

Both are correct, about different directories. The driver runs in the repository root and
spawns each scorer with `cwd: scripts/agent`, because it builds the child's argv with a
relative module path (`eval/volume-mix.mjs`) and that cwd is what makes the path resolvable.
The cwd exists for the *script* path — and it silently reparents every other relative path in
the same argv. `.github/workflows/eval-score.yml` passes `--root .eval-store`, so the driver
resolved it against `<repo>/` and found the corpus while every scorer resolved it against
`<repo>/scripts/agent/` and found nothing.

`--out` has the identical defect, and it bites even when the root is absolute: the driver
writes a payload relative to its own cwd, then hands `report.mjs --from` a path the child
resolves under `scripts/agent`. Measured — `ENOENT` on a file the driver had just written.
CI never met that half only because the workflow happens to pass an absolute `$RUNNER_TEMP`
path. One bug with two mouths, so both are resolved in one place.

The refusal message is part of the fix rather than polish. It named the corpus and no path,
so the failure read as *"the corpus was re-frozen and this schedule points at a retired
version"* — a real thing that will happen one day, and exactly what the comment above
`SCHEDULED_CORPUS_VERSION` warns about. A message that confidently accuses the wrong suspect
costs more than no message.

**Why the existing tests could not see it, which is the more useful half.** Every end-to-end
run was driven with an absolute path; the orchestrator tests inject the spawn function, so no
argv ever met a real working directory; and the workflow test asserts that every flag the
workflow passes is a flag the driver accepts — which they all are, because the *values* were
never exercised. No test had a relative path meet a real child process. The new one does.

## Linked Issues

None. It fixes the lane added in #834.

## Author checklist

- [x] Self-reviewed the diff
- [x] Tests added for the new behaviour
- [x] Documentation updated — a new section on `docs/tasks/active/20260813-eval-scoring-lane-todo.md`
- [x] AI assistance noted — see *Notes for Reviewers*

## Verification

- [x] CI green
- [x] Reviewed my own changes

`agent:tests`, two invocations, base and branch extracted with `git archive` and given the
same `node_modules`:

| | rest | iso | total | fail | skip |
|---|---|---|---|---|---|
| base `7d3aab3` | 2224 | 56 | **2280** | 0 | 0 |
| branch | 2226 | 56 | **2282** | 0 | 0 |

**+2, exactly the two new tests.** `eslint scripts` (the lockfile-pinned 9.24.0) exit 0 on
both trees. **3 mutations, 3 caught by the specifically-named test** — reverting either
resolve, and dropping the resolved path from the refusal.

The control that matters more than the count: **all 40 pre-existing tests pass with the bug
restored.** That is why the two new ones had to exist.

The failing invocation now passes, run from the **committed** tree against a fresh clone:
`--root .eval-store --out ./payloads` from the repository root → exit 0, all seven scores
filed, the report rendered.

The new test drives a full pass with a stand-in for the two children that have side effects,
writing through a real temporary store — so the driver's own read-back check and its final
existence sweep still run — and then asserts every `--root` and `--from` handed to a child is
absolute, with a count so neither assertion can pass by never running.

## Risk Assessment

**User-facing:** none. No product code, no lens, no gate, no workflow change. The lane this
fixes is triggered only by `workflow_dispatch` and `schedule`, so it is attached to no commit
and cannot gate a merge.

**Data and security:** no new permission and no change to any. Worth stating plainly what the
old behaviour did and did not risk: the lane exited non-zero, wrote nothing, skipped its
commit step and went red — so nothing partial or misleading was ever committed, and the data
repository it writes to is untouched by the bug. Resolving a path to an absolute one narrows
what a child can reach rather than widening it.

**Rollback:** a revert. Note that reverting returns the lane to failing on every scheduled
run, which is the current status quo rather than a regression.

## Notes for Reviewers

**AI assistance:** written with Claude Code — the fix, the tests and the task-doc section.
Every figure above was produced by running the command named, on this branch's committed
tree.

Deliberately **not** in this change: making the child *module* paths absolute so that no
working directory is load-bearing at all. That is the root-cause cleanup; it is a larger
change than this failure warrants, and doing it here would hide this fix inside it. It is
recorded as a follow-up in the task doc.

Two follow-ups this does not do. The lane has **still never completed a run on GitHub** — the
token, `RUNNER_TEMP` and the push are simulated locally, which is the gap that let this
through, so it is worth one `dry_run: yes` dispatch before trusting the next scheduled run.
And the report's cost-and-latency section now renders real figures with their intervals
named, which makes an assertion over the *rendered* section possible for the first time; that
is a separate PR.

### DIFF
--- a/docs/tasks/active/20260813-eval-scoring-lane-todo.md
+++ b/docs/tasks/active/20260813-eval-scoring-lane-todo.md
@@ -317,3 +317,120 @@ artefact.
 - [ ] **The lane has never run on GitHub.** Everything above is local: the two checkouts,
       the token, `RUNNER_TEMP` and the push are simulated. The first dispatch should be
       `dry_run: yes`.
+
+---
+
+# Follow-up — the lane's first scheduled tick failed (2026-08-20)
+
+## The problem
+
+`Eval Score` fired on its Monday cron for the first time, **2026-08-17 05:54Z**, and the
+*Score and render* step exited 1 without filing anything
+([run 31999551889](https://github.com/wafflebase/wafflebase/actions/runs/31999551889)):
+
+```
+score-all: 7 corpus item(s), 3 replicate(s) — 2026-08-10-pilot-reviewed @ sha256:1c78…
+score-all: API budget — 4978 of 5000 core call(s) remain, resets 06:45:28Z; needs ~175, requires 263
+score-all: volume-mix.mjs (pilot-01__k1)
+corpus version "2026-08-10-pilot-reviewed" does not exist under this root
+score-all: volume-mix.mjs (pilot-01__k1) exited 1. Nothing was filed
+```
+
+**Read those two middle lines together: the driver found 7 corpus items under that root, and
+the first scorer then said the corpus does not exist under the same root.** Both are
+correct, about different directories.
+
+**The cause.** This driver runs in one working directory and its children run in another.
+`scorerArgs` builds a child's argv with a *relative* module path (`eval/volume-mix.mjs`), so
+`doRun` sets `cwd` to `scripts/agent` to make that resolvable — and that cwd silently
+reparents every other relative path in the same argv. The workflow passes
+`--root .eval-store` from the repository root, so:
+
+| | resolves `.eval-store` against | result |
+|---|---|---|
+| the driver | `<repo>/` | found the corpus, 7 items |
+| every scorer | `<repo>/scripts/agent/` | *"does not exist under this root"* |
+
+**`--out` has the identical defect, and it bites even when the root is absolute.** Measured:
+with `--root` absolute and `--out ./my-payloads`, the driver *wrote* the payload relative to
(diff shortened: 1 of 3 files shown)


harrykim8672:
@claude review

harrykim8672 (approved):

PR #905, opened 2026-08-20
Score precision over adjudicated finding labels in eval/validity.mjs

dlgpdmsly2:
## Summary

Adds `scripts/agent/eval/validity.mjs` (`validity-v1`, scope `cross-run`), the first
scorer that reads the finding labels `eval/labels.mjs` defines and `eval/adjudicate.mjs`
writes. It computes precision per arm and per label tier, severity-weighted precision, a
relative-recall band and a false-positive profile — and declares absolute recall and the
miss profile permanently unmeasurable, with reasons.

Two new files and one task doc. **No existing file is touched.** Nothing is written,
spawned or spent: it reads an eval-data root, computes and prints.

## Why

`eval/` can already say how much each reviewer says, how consistently it says it and what
it costs. **None of those numbers is about whether any of it is correct.** `labels.mjs`
landed the record that says so and `adjudicate.mjs` landed the CLI that writes one, and
nothing reads either.

Writing that consumer is not arithmetic. Four things make a precision figure on this data
wrong in ways nothing downstream detects, and each shapes a specific part of the file:

**A missing denominator looks exactly like a thin one.** On the frozen corpus CodeRabbit
raised **0** `critical` findings and **3** `major` ones — reproduced independently by this
scorer's own claim census. Printing `n<5` for `critical` tells a reader to wait for labels
that can never arrive. So `availabilityFor` has four states, not two: a zero denominator
over an *exhausted* claim population is `not-measurable`, the same zero with claims still
unlabelled is `not-computed`, below `min_n` is `suppressed`. `exhausted` has to be earned —
every distinct claim the arm made already carries a label.

**A label that fails to meet its claim removes a judgement from the denominator and
inflates nothing**, so nothing looks wrong. There are five join statuses and only one of
them scores. A CodeRabbit `finding_key` hashes a summary *we* parse out of their markdown,
and that parser is being corrected, so `parser_vintage` is checked **before** the key: a
stale parse can still produce a matching key, and admitting it would pool a judgement
about text we can no longer reproduce with judgements about current text. A vintage that
cannot be checked at all is refused rather than assumed current. Unmatched keys are listed,
not summarised.

**The unit crosses a level and keeps its name.** `adjudicate.mjs` bundles by defect class,
so N labels can come from one reading. Every cell carries `labelled_findings` and
`readings` in all four states — the first is the denominator, the second is the only number
a claim about independence may use — and the claim census counts `distinct_finding_keys`
and says in the field name that it is not a defect-class count.

**Two metrics cannot be answered on this corpus at all.** Absolute recall and the miss
profile need defects found by something outside both arms, and a set built from what
reviewers found cannot contain what they missed. They are declared not computable,
filtered out before any cell is built, and `precisionCell` refuses the ids outright —
there is no argument that produces a figure for either.

Relative recall is emitted as a **band** rather than a point: its denominator is the union
of the two arms' confirmed-real sets and the overlap is unknown, so the union is bounded
`max(a, b) … a + b` and both bounds are printed. With no labelled finding on the other arm
the band collapses to `[1, 1]`, which is refused rather than published.

**This changes no number today.** No finding label exists for the current corpus version
yet, so against the real store every figure is absent and the run exits non-zero. That is
the intended output for an unadjudicated store, and it is what the scorer is built to say
rather than something it fails to say.

## Linked Issues

None. Follows `eval/labels.mjs` + `eval/adjudicate.mjs`, which defined the record this
reads.

## Author checklist

## Verification

Skip reason (if applicable): none — CI runs both.

Measured locally, from the committed tree, against `main` at
`98d0565914c169e0d56432f98a2ab6a35d1fa7e9` — both trees extracted with `git archive` and
given the same `node_modules`, then measured once each. `main` has moved on by one commit
since; the counts below are anchored to the sha they were taken at rather than to today's
tip, and the delta is what carries over.

| | |
|---|---|
| `node --test` (all but `eval/run.test.mjs`) | **2258 tests, 0 fail, 0 skip** — against a freshly measured **2224, 0 fail, 0 skip** on the same tree without these files. **+34**, this file's whole test count |
| `node --test eval/run.test.mjs` (isolated) | **56 tests, 0 fail**, identical on both trees |
| `npx eslint scripts` | exit **0**, no output, on both trees |
| Mutation testing | **28 mutations, 28 caught by the specifically named test.** The harness proves each mutation changed the file's bytes before running and verifies the subject is restored afterwards |
| Real eval store, both arms | every figure absent, `PARTIAL`, exit 1 — no finding label exists for this corpus version yet. The claim census reproduces CodeRabbit's 30 findings at `critical=0 major=3 minor=13 nit=14` independently |

## Risk Assessment

- **User-facing risk:** none. Nothing in the review or gating path is touched — no lens,
  no gate, no workflow, no `severity.mjs`, no `finding-match.mjs`. The panel cannot import
  this module and does not.
- **Data/security risk:** none. The module writes nothing; `--root` is required and has no
  default anywhere, so no invocation can fall back to a path inside this repository. The
  CLI's only network access is the read-only GitHub API calls the CodeRabbit arm adapter
  already makes to rebuild that arm's finding population; the library itself is pure and
  its tests need no network, no key and no store.
- **Rollback plan:** revert. Nothing reads this module — the report's `SECTIONS` array is
  untouched, so no rendered output changes — and no stored artifact depends on it.

## Notes for Reviewers

- **AI assistance:** this PR was written with Claude Code — the module, its tests, the
  mutation harness and this body. Every measurement quoted above was run rather than
  estimated, and the baseline was measured on the same tree with the same dependency set.
- **Follow-up work:** rendering `validity-v1` needs a sixth entry in `report.mjs`'s frozen
  `SECTIONS`, which is deliberately not done here — it is a
  rendering change with its own review surface, and this scorer is complete without it. Persisting a score payload is likewise a
  separate change; `store.mjs` already accepts the id and needs no edit.
- **Why the severity strata use the annotator's severity and not the reviewer's:** it is
  the truth side of the ratio. Weighting or stratifying by the severity an arm *claimed*
  would let that arm decide the weight of its own errors. The reviewer's distribution is
  in the claim census beside it, under a field name that says whose it is.

## Summary by CodeRabbit

* **New Features**
  * Added a validity scoring report for evaluating finding quality across runs.
  * Reports precision, severity-weighted precision, relative-recall ranges, and false-positive profiles.
  * Supports filtering and comparison by evaluation arm, run, item, and threshold.
  * Provides JSON and human-readable report formats.
* **Data Quality**
  * Identifies incomplete claims, unmatched or outdated labels, unreadable inputs, and unavailable measurements.
  * Keeps severity levels, reading levels, populations, and label tiers separate for accurate comparisons.
* **Safeguards**
  * Clearly distinguishes suppressed, unavailable, and unmeasurable results.
  * Does not report unsupported absolute-recall or missed-defect metrics.

### DIFF
--- a/docs/tasks/active/20260820-eval-validity-scorer-todo.md
+++ b/docs/tasks/active/20260820-eval-validity-scorer-todo.md
@@ -32,7 +32,7 @@ metrics (precision, severity-weighted precision, a relative-recall band, the
 false-positive profile), two permanent refusals, a per-arm claim census, a per-arm join
 census, and a CLI whose `--root` has no default.
 
-`scripts/agent/eval/validity.test.mjs` — 34 tests.
+`scripts/agent/eval/validity.test.mjs` — 39 tests.
 
 Nothing is written, nothing is spawned and nothing is spent. `store.mjs` needs no change:
 `validateScore` (`store.mjs:383`) only requires the scorer id to be a legal path segment,
@@ -40,15 +40,17 @@ so there is no registry to extend.
 
 ### Three ways to have no number, and the fourth that has one
 
-`AVAILABILITY` is `reported | suppressed | not-computed | not-measurable`, and the three
-absent states are decided by `availabilityFor` from the denominator plus what is known
-about the claims behind it:
+`AVAILABILITY` is `present | suppressed | not-computed | not-measurable` — **the
+renderer's own vocabulary**, imported from `report.mjs` and checked for set equality at
+import, because a cell from here is eventually handed to its `renderCell`, which refuses
+any state it does not know. The three absent states are decided by `availabilityFor` from
+the denominator plus what is known about the claims behind it:
 
 ```
 labelled_findings = 0, claim population exhausted   → not-measurable   (final)
 labelled_findings = 0, claims still unlabelled      → not-computed     (pending)
 0 < labelled_findings < min_n                       → suppressed       (thin)
-labelled_findings >= min_n                          → reported
+labelled_findings >= min_n                          → present
 ```
 
 `exhausted` has to be **earned**: every distinct claim the arm made already carries a
@@ -99,9 +101,11 @@ union_low  = max(a, b)   →  high = a / union_low
 union_high = a + b       →  low  = a / union_high
 ```
 
-Both bounds are emitted and there is **no `value` key**. One case is refused outright:
-with no labelled finding on the other arm the band collapses to `[1, 1]`, and *"we found
-100% of confirmed defects"* would be true of the arithmetic and false about the world.
+Both bounds are emitted and there is **no `value` key**. Both arms' support gates it,
+because the band has a tautology at each end: with no labelled finding on the **other** arm
+it collapses to `[1, 1]` — *"we found 100% of confirmed defects"* — and with none on
+**this** one it collapses to `[0, 0]`. Each is true of the arithmetic and false about the
+world, so both are refused, and a numerator below `min_n` is withheld.
 
 ### Tiers are never pooled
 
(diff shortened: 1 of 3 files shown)


harrykim8672:
@claude review

hotsunchip (approved):

harrykim8672 (approved):

PR #908, opened 2026-08-20
Add a validity section to the eval report and name the panel in its header

dlgpdmsly2:
## Summary

Three changes to `scripts/agent/eval/report.mjs`, plus the wiring that makes the first one reachable from the scoring driver.

- **A validity section.** `eval/validity.mjs` is on `main` and nothing reads it — `report.mjs` had 0 references to `validity-v1`, `score-all.mjs` 0 references to validity. A sixth `SECTIONS` entry and a `validityFigures()` render every metric it declares as a cell in the renderer's existing four-state vocabulary. Limits renumbers §6 → §7.
- **A readable header.** The report identified its reviewer as two hashes. It now also names the six lenses with the model, sample count, effort and gating each one runs, the `config_id`, the replay target and the SDK version — all read from the run envelope and config snapshot the render path already opens. Both full hashes stay.
- **`score-all.mjs` runs validity**, which needed the API-budget model to stop hard-coding its reader count and needed one declared non-zero exit code to be tolerated under an obligation.
- **§1 and §3 gained two sentences each** saying what their metric is, what would move it and what it cannot tell you — the treatment §2 and §4 already have.

## Why

**Nothing read the one scorer that is about correctness.** Every other section here measures how *much* a reviewer says, how *consistently*, and how *cheaply*. `validity.mjs` measures whether a finding is true, and with it unwired the only trace of precision on the page was one sentence in the limits — where a reader cannot tell it from a metric nobody thought of. That distinction is exactly what this renderer's four availability states exist to draw, so precision is now a **cell**: today it reads `not computed` with the label census behind it, and it will read a figure when labels exist.

**Two of the six metrics are refused permanently, and the renderer refuses rather than coercing.** `absolute_recall` and `miss_profile` need a defect found outside both arms — the item label's `true_defects[]`, from a human review or a revert — and a corpus assembled from what two reviewers said cannot contain what they both missed. `not-computed` tells a reader to wait for a judgement that can never be made, so a payload declaring either metric as merely uncomputed stops the render.

**A digest cannot say which reviewer produced a number.** `panel_sha 46da673d…` has no ordering: beside a newer hash it says *different*, never *older, in which respect*. So the page invited comparison against a panel that no longer exists without saying so. Everything needed was already in `runs/<id>/config.snapshot.json` and `run.json`, and `store.getRun` returns both in one read — this needs no new accessor, no new file format and recomputes nothing. **A disagreement across replicates is reported, never resolved:** three snapshots can differ, and printing the first leg's answer would describe a reviewer that produced a third of the data, so a disagreeing lens set leaves the lens table empty and names what each leg said instead. `captured_at` is deliberately not an axis — the three pilot snapshots differ in it by nineteen hours and describe one reviewer, which is why `config-hash.mjs` classifies it as cosmetic.

**The driver's API preflight was about to understate a pass in the flattering direction.** `validity.mjs` rebuilds the CodeRabbit arm's claim population through `adapters/coderabbit.mjs` — five endpoints per corpus item — so it is a third cross-run API reader, and the budget model's literal `replicates + 2` no longer described the pass. `API_READERS` now derives both counts from `STEPS`. The 7-item pilot at K=3 costs **210** core calls where it cost 175; the earlier measurement is not stale, the pass grew by one reader. Left undeclared the preflight would have cleared at 175 and met the limit part way through, filing a score computed from a partial arm — which is indistinguishable from a clean review.

**`validity.mjs` exits 1 when its own verdict is `partial`, and on an unadjudicated store that is its correct answer.** Under the driver's plain exit-code rule the lane would have filed every other score and then aborted on the one whose honest output is the absence. One step now declares one tolerated code, and the tolerance is checked *harder* rather than more loosely: accepted only if the payload parses and says `partial` itself. Exit 1 over a `complete` payload refuses (code and payload disagree), a payload with no verdict refuses (the check's input never arrived, and an exit code cannot be told from a crash), and 2 or 137 refuse as before.

**This changes no measured number.** Every figure in the report reads the same as before; what changed is which absences are stated and how the reviewer is named.

## Linked Issues

None. Follows the report renderer's own history in `docs/tasks/active/20260812-eval-report-renderer-todo.md`, where this is the fourth appended section.

## Author checklist

## Verification

forced the whole suite.

Skip reason (if applicable): nothing outside `scripts/agent/**` and one task doc is touched, so no product suite can be affected.

Measured locally from the **committed** tree, with the base tree extracted separately and the same lockfile-pinned `eslint@9.24.0` and `scripts/agent` `node_modules` symlinked into both so the skip counts are comparable:

- `agent:tests`, both invocations as the lane runs them: **2302 + 56 = 2358 tests, 0 fail, 0 skip**, against a freshly measured **2288 + 56 = 2344** on the base — **+14**. Both isolated runs used a private `TMPDIR`.
- `npx eslint scripts` exits **0**.
- **25 mutations, 25 caught by the test named for each.** The harness verifies it applied every mutation — needle count exactly one, file hash changed — and restores the tree byte for byte. Five that are the point: the permanent refusal coerced to `not-computed`, a figure's `n` taken from `readings` instead of `labelled_findings`, the lens table falling back to replicate 1 on a disagreement, `captured_at` promoted to an axis, and `reads_api` written `false`. One mutation survived on the first pass and proved *ineffective* rather than uncaught — an inserted rather than moved explanation left the original above the table — and is now a two-edit move.
- Rendered end to end against a real store: 394 → 468 lines, and the byte-identical re-render test passes untouched with a second one covering the enlarged document.

## Risk Assessment

- **User-facing risk:** none in the gating path. Nothing here is imported by `review-panel.mjs`, no lens behaviour changes and no gate decision is touched; the benchmark only reads. The one behavioural change a caller can see is in `score-all.mjs`, which is `workflow_dispatch`/schedule-only and gates nothing: it now runs a sixth scorer and tolerates one declared exit code from it. Its refuse-on-doubt contract is unchanged for the other five steps, and a test asserts exactly one step declares a tolerance.
- **Data/security risk:** none. No new write path, no credential, no network call added — `report.mjs`'s render path still opens no socket and its tests run with no store and no network at all. The report gains no field that was not already committed to the eval store. The API budget change makes the preflight ask for *more* headroom, not less.
- **Rollback plan:** a revert. The validity section renders `not computed` when no score is filed, so reverting returns the page to five sections; `score-all.mjs` returns to five steps and the 175-call model. No stored data is written or migrated by this PR, so nothing has to be undone outside the diff.

## Notes for Reviewers

- UI changes (screenshots/gifs if applicable): none — the artefact is a markdown report. The rendered before/after is in the task doc's fourth section.
- **AI assistance:** written with Claude Code. It produced the renderer, the tests and the task-doc section; the design decisions above were reviewed and approved by a human before the commit was built, and every measured number in *Verification* was produced by running the commands rather than estimated.
- **Follow-up work:** §6 renders four absences today because no finding label exists for any corpus. The present-cell, band and false-positive-profile tables are exercised by fixtures built from `validity.mjs`'s own `precisionCell`, `relativeRecallBand` and `fpProfile` — so the shapes are the scorer's — but no store has such a cell yet, and that is called out as unverified-on-real-data in the task doc. §5's grid is unreadable for a separate reason (151 rows of a flattened composite key) and is deliberately untouched here; it is being restructured in its own change.

## Summary by CodeRabbit

* **New Features**
  * Added a final validity section to evaluation reports, covering precision, relative recall, and false-positive profiles.
  * Reports now show metric availability, units, denominators, and permanent refusals for unmeasurable results.
  * Added reviewer provenance, including lens configuration, model and SDK details, targets, hashes, and cross-run disagreement information.
  * Evaluation summaries now include partial validity results and reasons.

* **Bug Fixes**
  * Improved validation, deterministic rendering, overlap reporting, label census checks, and handling of incomplete scorer results.

* **Tests**
  * Expanded coverage for validity metrics, availability states, provenance, partial results, and invalid payloads.

### DIFF
--- a/docs/tasks/active/20260812-eval-report-renderer-todo.md
+++ b/docs/tasks/active/20260812-eval-report-renderer-todo.md
@@ -798,3 +798,378 @@ with `.reverse()`, the ordering test catches it.
       does not change until the scorers and the renderer are re-run against the store.
 - [ ] **Not run:** `verify:self`, `verify:fast`, `verify:browser`, `verify:integration`,
       `build`.
+
+# Follow-up — the validity section, readable provenance, and §1/§3 explained
+
+*2026-08-20, on top of the three sections above. Built against `upstream/main` at
+`3b7a067`, then REBASED onto `39debb4` after #909 landed §5's restructure — see
+"Rebased onto #909" at the end of this section for what that changed and what it did not.*
+
+* Three changes: a section for the metric this benchmark is eventually FOR, a
+header a reader can act on, and two sections that stated figures without saying what they
+were figures of.*
+
+## The problem
+
+**① `validity.mjs` merged (#905) and was wired into nothing.** Measured on the base:
+`report.mjs` carried **0** references to `validity-v1` and `score-all.mjs` **0** references
+to validity. So precision — the only metric in this project about whether a finding is
+TRUE — appeared on the page as one sentence in the limits. A reader cannot tell prose in a
+footer from a metric nobody thought of, and the two are exactly the distinction this
+renderer's four availability states exist to draw. Worse, the two metrics this corpus can
+**never** answer had no representation at all: `absolute_recall` and `miss_profile` were
+declared not-computable inside the scorer and invisible outside it.
+
+**② The header named the reviewer as two hashes and nothing else.**
+
+```
+| reviewer | `panel_sha 46da673dd46dd5576626ee6d1b4e2e40728345e0` · `sha256:1c7853deb…f01` |
+```
+
+A digest has no ordering. It can say *different* and never *older, in which respect* — so
+the page invited comparison against a panel that no longer exists without saying that is
+what a reader would be doing. Everything needed was already in the store and unread:
+`runs/<id>/config.snapshot.json` holds `config_id`, `target`, `sdk_version`,
+`config_hash_version` and `lenses[]` with a model on each, and `run.json` holds
+`panel_sha` and `panel_sha_source`. `store.getRun` already returns the snapshot beside the
+envelope, and the render path already called it.
+
+**③ §1 and §3 stated figures and assumed the reader knew why they mattered.** §2 and §4
+carry their reasoning inline and it works — *"read the two directional rates before the
+Jaccard"*, *"the two figures time different things"*. §1 printed `4.6×–4.9×` with no
+sentence saying that a count is a count; §3 printed two figures that point opposite ways
+without saying what reproducibility is a property OF.
+
+## The change
+
+### §6 — every metric spec §3.3 names, as a cell in the same four states as every figure
+
+A sixth `SECTIONS` entry and a `validityFigures()` in the house style. One grid, six rows,
+`metric | status | unit`. Today four rows read `not computed` with the label census behind
+them and two read `not measurable` with the scorer's own reason — and the section is
+**last**, because the order in `renderReport` is the argument: measured first, bounded
+second, unavailable third. Limits moves from §6 to §7; **§5 is untouched**, deliberately,
+because `prompts/report-rewrite-section5.md` owns it in a parallel branch.
+
+**The refusal is a REFUSAL, not a coercion.** A payload declaring `absolute_recall` as
+anything other than `not-measurable` stops the render. `not-computed` tells a reader to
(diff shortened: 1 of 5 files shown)


dlgpdmsly2:
@claude review

hotsunchip (dismissed):

hotsunchip (approved):

PR #915, opened 2026-08-21
Key cross-run eval scores by a panel_digest, not by config_hash alone

dlgpdmsly2:
## Summary

Adds `scripts/agent/eval/panel-identity.mjs`: a `panel_digest` — a sha256 over the review
panel's own source files, from an explicitly declared file set — and puts it in the eval
store's cross-run score key.

- `run.mjs` records `panel_digest` in `run.json` and every item envelope;
  `validateRunEnvelope` requires it. Nothing recomputes one.
- `byConfigSegment(configHash, panelDigest, corpusVersion)` →
  `<config-hash>__<panel-digest>__<corpus-version>`, threaded through `_scorePath`,
  `putScore`, `getScore`, `comparisonIdFor` and `writtenPaths`.
- A cross-run score whose pooled runs and items disagree on the panel now **refuses**, naming
  each digest and its record count. `--allow-mixed-panel` files it under `mixed` instead, and
  the payload must then carry the list of panels pooled.
- `panel_digest_source` records **how** a digest was obtained, from a closed vocabulary:
  `files` (hashed at run time — the only observation), `reconstructed` (stated: computed out
  of git after the fact) or `envelopes` (read off the records a score pools).
- `putScore`'s docblock claimed this safety property already held. It did not; the claim is
  corrected in the same diff that makes it true.
- No lens, no gate and no panel file is touched.

## Why

`config-hash.mjs` states its own limit in its header: *"This hashes the LENS COMPOSITION, not
the panel's code. A new verifier stage or a changed gate leaves the hash identical."*
`byConfigSegment` then keyed every cross-run score by `(config_hash, corpus_version)`. So two
runs of two different panels produced one directory name and one report filename, and the
second write overwrote the first with nothing logged.

`putScore` said that could not happen — *"a score computed under a different reviewer lands in
a different file by construction"* — and said it as the reason no check existed. The premise
was true of `per-run` scores and the conclusion did not follow for `cross-run` ones.

Measured in the eval store: one run holds 16 item envelopes carrying 16 distinct `panel_sha`
values, which are 5 distinct panels by content, all under one `config_hash`. Exactly one of
those items ran a panel containing the post-freeze demotion from #881 — a gate the other
fifteen do not have, deciding which findings block. One pooled score spans two answers to
that question.

**Why a content digest rather than the `panel_sha` already recorded.** It is the right
provenance and the wrong identity: it separates panels that are byte-identical. #830 deleted
`scripts/agent` and #850 returned it, and the returned file is byte-for-byte the deleted one.
16 commits are 5 panels — a 3× over-separation, and `parser_vintage` is the existing
precedent for using content instead.

**Why the file set is declared and not walked.** The transitive closure of
`review-panel.mjs`'s local imports sweeps 13 modules at one panel version and 17 at `main`,
four of which are not the reviewer — check-run I/O and a process helper. A computed set also
grows silently, the failure `HASHED_LENS_FIELDS` exists to prevent. So `PANEL_FILES` is
declared, every other reachable local import carries a written reason in `NOT_PANEL_FILES`,
and a test walks the imports and fails on anything in neither register.

**Why the lenses are not in the digest.** `config_hash` already covers them at content level —
`HASHED_LENS_FIELDS` includes `rubric_sha256`, which is `contentHash(rubricText)` — and a
replay's lenses come from the run's frozen config snapshot rather than from the files beside
`review-panel.mjs`, so hashing those would name a configuration that did not review anything.
`config_hash` is the configuration, `panel_digest` is the code.

An envelope that states no digest resolves to the named state `not-recorded` rather than a
blank or a refusal: every run predating the field is in that state, which is not a fault, but
it is not a safe pooling key either — so it appears in the path where it can be seen. A
two-part legacy directory is neither read nor written nor deleted; matching it would pool the
panels the key exists to separate.

**Why the source vocabulary is part of this and not a follow-up.** Runs recorded before this
field exists can still be scored under the panel that produced them, by computing that panel's
digest out of git and stating it. That is the right thing to do and it is *not* an observation —
so `reconstructed` says which, and `putScore` refuses a source outside the vocabulary. Without
it the field would quietly claim every reconstructed digest had been measured, which is the
"asserted, not measured" failure the neighbouring `panel_sha` resolution was written to prevent.

🔴 **The benefit is partly deferred, and that is worth stating plainly.** Every envelope now in
the store predates `panel_digest`, so a pool of them resolves to one `not-recorded` bucket
rather than to a mixture — **the refusal does not fire on existing data.** Such a pool is filed
under a path that says no panel was recorded, which is an improvement on a path that implied
one, and a long way short of the guard firing. It starts firing as soon as anything re-records:
a fresh replay does so as it runs.

Beyond that, this changes no published number. It changes which file a number lands in, and
makes a pool spanning two panels say so instead of averaging them.

## Linked Issues

None. This follows #782 and #908 in `scripts/agent/eval/`, and fixes a claim made in
`store.mjs`'s own docblock rather than a filed issue.

## Author checklist

## Verification

Measured locally, both trees extracted from git with identical `node_modules`, against
`main` at `5f2be616`:

| | rest | iso (`eval/run.test.mjs`) | total |
|---|---|---|---|
| `main` `5f2be616` | 2337 pass · 0 fail · 1 skip | 56 pass · 0 fail | 2393 |
| this branch | **2349 pass · 0 fail · 1 skip** | **57 pass · 0 fail** | **2406 (+13)** |

`npx eslint scripts` exits 0. **Every new test mutation-tested: 31 of 31 mutations caught, all
31 by the specifically-named test** — three survived a first pass and each got a test. The
escape test was demonstrated both ways: a fake local import in `review-panel.mjs` reddens it,
removing the import greens it. Verified from the committed tree, not a working copy.

## Risk Assessment

- **User-facing risk:** none in the product. Nothing here is imported by `review-panel.mjs`,
  any lens, or the gate — the dependency runs one way, and `review-panel.mjs` is unmodified
  (its 176 tests pass untouched). The risk is to the offline eval lane: a cross-run score now
  needs a panel identity to be filed, so a caller that passes none is refused rather than
  writing to a two-part path. The three call sites are updated together.
- **Data/security risk:** no existing stored data is read differently, rewritten or deleted.
  One thing a reviewer will reasonably ask about a re-score, answered precisely: **scoring
  never runs the panel** — `score-all.mjs` has zero references to `review-panel.mjs`,
  `novelty.mjs` or `reviewer.mjs` in its spawn path — but `eval/finding-record.mjs:34` imports
  `findingLocation` from `../novelty.mjs`, and every cross-run scorer normalises through it.
  That function changed recently (first citation → first *same-file* citation), so:

  | | moves on a re-score? |
  |---|---|
  | the `line` field → localization figures | **yes, upward** |
  | `finding_key`, defect classes, cross-arm matching | **no** |
  | the adjudicated pair labels | **no** |

  The last two hold because `line` is not part of the key — `finding-record.mjs:252` says so,
  and `findingKey` is `` `${file}::${summary}` `` (`finding-key.mjs:40`). **A re-score cannot
  invalidate the adjudicated labels.**
  Scores already filed under two-part directory names stay exactly where they are and are no
  longer found, which is deliberate and stated in `byConfigSegment`'s docblock; the previously
  published report keeps its old name as a dated snapshot. No credential, no network call, no
  spend — the digest is a local file read, and the one tool that reads git history is an
  operator CLI that a scoring path may not use.
- **Rollback plan:** a revert. `panel_digest` is additive on the write path, so envelopes
  written by this branch remain valid under the old validator (which ignores the field);
  cross-run scores filed under three-part names would become unreachable, and re-scoring is
  free of model calls.

## Notes for Reviewers

- **AI assistance:** written with Claude Code. The design, the `PANEL_FILES` membership calls
  and the mutation testing were reviewed line by line; two defects the mutation pass surfaced
  are described in the task doc's *Corrected while building*.
- **Task doc:** `docs/tasks/active/20260820-eval-panel-digest-todo.md` carries the full
  reasoning, the fail-direction table and what was deliberately left unverified.
- **The one judgement call worth a second opinion:** `ask.mjs` and `redact.mjs` are excluded
  from the panel's identity even though each carries something that reaches a model or a
  reader. Both are shared with every other agent script in `scripts/agent/`, so including them
  would let an unrelated change re-partition the panel's populations. The reasons are data in
  `NOT_PANEL_FILES` so the call can be revisited against evidence.
- **Follow-up work, deliberately not here:**
  1. A mechanism for declaring two panel versions *equivalent*. `panelKeySegment` is the seam
     it replaces — one function, not `byConfigSegment`'s call sites a second time.
  2. The rendered report gains a row naming the panel's contents and an agreement axis; the
     prose explaining *how* two panels differ is a separate change.
  3. Backfilling `panel_digest` onto envelopes written before this PR, which is what would let
     the mixed-panel refusal fire on data already in the store — see the deferred-benefit note
     in *Why*.

## Summary by CodeRabbit

- **New Features**
  - Added content-based panel identity for evaluation runs, scores, and reports.
  - Panel digests now appear in run metadata, score records, report details, and comparison paths.
  - Added options for specifying panel digests and explicitly allowing mixed-panel evaluations.
  - Reports show panel contents and distinguish panel identity differences from reviewer differences.

- **Bug Fixes**
  - Prevented score and report collisions between panels sharing configuration.
  - Added validation for missing, malformed, unreadable, or unintentionally mixed panels.

- **Documentation**
  - Added guidance on panel identity, historical reconstruction, migration, limitations, and verification.


dlgpdmsly2:
@claude review

hotsunchip (approved):

PR #948, opened 2026-08-24
Normalize the lens id where author-written records enter the panel

dlgpdmsly2:
## Summary

The fix agent writes `lens` from what it can see on the PR — a check run named
`agent-review-correctness` — while findings carry the bare `correctness` from
`stampLens`. `findingSimilarity` gates on lens equality and returns **0 outright**,
not a low score, so neither a fix report's claims nor a rebuttal could join the
finding it names.

Strips the prefix at the two boundaries where author text becomes a lens id:
`normalizeItem` (fix-report.mjs, which every read and write path funnels through)
and `parseRebuttalComment` (rebuttal.mjs). Anchored — none of the six ids in
`lenses/lenses.json` begins with `agent-review-`. `findingSimilarity`, `claimFor`,
`matchRebuttal`, `clusterFindings`, `finding-match.mjs` and every threshold are
byte-identical; the rest of the diff is tests.

## Why

With the join dead, `claimFor` returns null, `applySkipClaims` passes the finding
through untouched, and `adjudication.upheld` is never set — so `upheldTwice` can
never be true and the standstill page (`guard-verdict.mjs:22`) cannot fire. The loop
burns its three rounds and pages with `round-cap` instead: a human is still paged,
with the wrong reason, and the fixer's stated reason is not recorded even as
unmatched. `substituteAdjudicated`'s docblock is the postmortem of a prior bug with
this one's consequence verbatim, down to *"the standstill page could never fire …
bounded only by the round cap"*.

Measured 2026-08-24 over the 16 agent-authored PRs: **296 of 348 claims (85%)** and
**8 of the 9 rebuttals** carry the prefix, and counting only the latest report — all
`authorClaims` reads — **15 of 16 PRs could join nothing at all**. The rebuttal side
is a second live bug, not latent hardening; `node scripts/agent/rebuttal.mjs read
695` shows it.

Four modules already strip this off the same check-run names (`fix-brief.mjs:131`,
`loop-status.mjs:202`, `prior-findings.mjs:83`, `rounds.mjs:598`), and
`agreedReviewedSha` accepts both spellings with a comment about this same silent
trap. Not fixed inside `findingSimilarity`: ~30 call sites including round-to-round
clustering, so normalizing there moves every number computed downstream.

Both boundaries or neither — `authorClaims` compares a report against a rebuttal, so
one side left in the other's vocabulary stops a rebuttal covering its own claim and
upholds through the session-free skip path while the argued rebuttal reaches no
adjudicator.

## Linked Issues

None. Self-contained; the defect and its measurement are stated above.

## Author checklist

## Verification

Every existing test passes with this bug present — that is how it shipped — so the
tests came first and each was mutation-tested. Per-file, from the committed tree
(`git archive <branch> | tar -x`): `fix-report` 35 / **2 fail** on `main` → **37 /
0**; `rebuttal` 39 / **1** → **40 / 0**; `review-panel` 176 / **1** → **177 / 0**.
Unchanged: `rounds` 64, `fix-brief` 19, `loop-status` 38, `prior-findings` 18,
`review-state` 14.

Mutation per boundary: reverting only the fix-report strip reddens 2 tests, only the
rebuttal strip reddens 3, both reddens all 4. The review-panel test does *not* catch
the fix-report-only mutation alone — `one disagreement counts ONCE` is the one that
dies with either boundary.

- [ ] verify:self — CI comment
- [ ] verify:integration — CI comment

Skip reason: `pnpm verify*` not run locally (no `node_modules` here); CI decides.

## Risk Assessment

- User-facing risk: none. Review-pipeline plumbing.
- Data/security risk: this makes a finding-**removing** path reachable. `skipped`
  claims only raise the uphold count, which moves toward the human page and never
  away, but a `fixed` claim becomes a rebuttal record and an overturn on
  `not-present` removes the finding. Guarded as designed — `isOverturningVerdict`
  needs high confidence, an enumerated ground and a cited location — but unreachable
  on 15 of 16 PRs until now. To ramp rather than switch, the lever is
  `MAX_FIX_ADJUDICATIONS`.
- Cost: the cap of 5 adjudications currently never fires; after this it does, and
  measured, one record buys exactly 1 session. `review-panel` wall-clock over 35
  runs peaks at 22.1 min against the 45-minute timeout, so 22.9 min of headroom —
  thin if 5 sessions cost 10–18 of it. `timeout-minutes` is deliberately untouched;
  that belongs in a follow-up.
- Rollback plan: revert the commit. One expression per boundary, nothing persisted.

## Notes for Reviewers

- Expect the standstill page to fire in production for the first time — that is
  `MAX_REBUTTAL_ROUNDS` working, but it is a page reason nobody has seen.
- **Known interaction, fixed by #949.** Normalizing the rebuttal lens makes
  same-target rebuttals collide, and `matchRebuttal` refuses a tie: on #786 three
  rebuttals name one finding with three different claims, so today the one bare
  rebuttal is adjudicated and after this all three tie and it gets none. Over the 9
  real rebuttals and the 5 findings they dispute: `main` 1 of 5, this PR alone 2,
  #949 alone 1, **both 5**. Fail-safe — the finding stands and keeps blocking — but a
  regression on one finding.
- AI assistance: written by Claude Code, including every measurement above.
- UI changes: none.

### DIFF
--- a/scripts/agent/fix-report.mjs
+++ b/scripts/agent/fix-report.mjs
@@ -54,11 +54,36 @@ const trim = (s, n) => (str(s).length > n ? str(s).slice(0, n) : str(s));
 // fence, so this is hardening rather than a bypass; a forgeable fence is no fence.
 const defence = (s) => str(s).replace(/<\/?fix-report>/gi, "[fence]");
 
+// The lens id, in the ONE vocabulary the panel matches on.
+//
+// The fixer fills `lens` in from what it can see on the PR, and what it can see is
+// a check run called `agent-review-correctness`; `stampLens` gives findings the
+// bare `correctness`. `findingSimilarity` gates on lens equality and returns 0
+// OUTRIGHT — not a low score — so a prefixed claim matched nothing, silently:
+// `claimFor` returned null, `applySkipClaims` passed the finding through
+// untouched, and the fixer's stated reason was not even recorded as unmatched.
+// Measured 2026-08-24 over the 16 agent-authored PRs: 296 of 348 claims (85%)
+// carried the prefix, and every one of those 16 PRs has at least one.
+//
+// Normalized HERE, at the author boundary, and not in `findingSimilarity`: that
+// function has ~30 call sites including round-to-round clustering, and changing
+// what it considers equal would move every number computed from it for the sake of
+// one field written in the wrong vocabulary. Same reasoning as the length caps
+// above — the untrusted writer's text is regularised on the way in.
+//
+// ANCHORED, so a real id can never be corrupted: none of the six lens ids in
+// lenses/lenses.json begins with `agent-review-`. Four other modules already do
+// exactly this to the same check-run names (fix-brief.mjs, loop-status.mjs,
+// prior-findings.mjs, rounds.mjs), and `agreedReviewedSha` in review-state.mjs
+// accepts both spellings for the same reason — this is the same convention, at the
+// boundary that had been left out of it.
+const lensId = (s) => str(s).trim().replace(/^agent-review-/, "");
+
 /** One reported item, length-capped. */
 function normalizeItem(raw) {
   const i = raw && typeof raw === "object" ? raw : {};
   return {
-    lens: trim(str(i.lens).trim(), 60),
+    lens: trim(lensId(i.lens), 60),
     file: trim(str(i.file).trim(), 300),
     summary: trim(i.summary, 2000),
     note: trim(i.note, 2000),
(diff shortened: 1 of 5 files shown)


dlgpdmsly2:
Follow-up filed: #949 fixes `matchRebuttal`'s tie rule, which is the "Known interaction" in the body above — the reason #786's correctness finding loses its one working adjudication under this PR alone.

Measured over all nine rebuttals ever filed and the five findings they dispute:

| | disputed findings adjudicated |
|---|---|
| `main` today | 1 of 5 |
| this PR alone | 2 of 5 |
| #949 alone | 1 of 5 — inert, since 8 of 9 rebuttals never clear the lens gate |
| both | **5 of 5** |

They are independent branches off the same `main` and touch different functions, so either can merge first — but if only one lands, this one should be it, and #949 should follow.

hotsunchip (approved):

PR #951, opened 2026-08-24
Add a lenses_dir input to eval-replay, checked before it spends

dlgpdmsly2:
## Summary

`eval-replay.yml` could only ever replay the frozen corpus through the one lens
configuration committed at `scripts/agent/lenses`. This adds a seventh dispatch input,
`lenses_dir`, naming a directory inside the checkout; empty keeps the committed config.

The preflight (`replay-plan.mjs`) refuses a bad one for free — absolute path, a `../` or
symlink escape from the checkout, a symlink among the lens files themselves, a directory
that is not there, a manifest with no lenses, or one that `buildConfig` will not build — and
on success prints the `config_hash` and each lens's model before anything is spent.

No runner change: `run.mjs --lenses-dir` already existed. No variant directory is committed
here.

## Why

Answering "what would this corpus find under a different model, or a different `effort`?"
meant editing `scripts/agent/lenses` on a branch and dispatching from it — which changes the
reviewer for everyone reading that branch and leaves the two arms of a comparison in
different commits.

Nothing new had to be built for the results to stay separable. `model`, `effort`, `samples`
and `rubric_sha256` are all in `HASHED_LENS_FIELDS`, so a variant lands under its own
`config_hash`; `panel_digest` hashes panel modules and no lens file, so it does not move.
That is exactly the discrimination the store's key was built for.

**A directory and not a `model` string.** A model input would need a new flag on `run.mjs`
and would mutate a committed config in memory, leaving the reviewer that produced a result
described only by something typed into a form. A directory needs no runner change, and the
variant becomes a reviewable artefact in git.

**The work is the preflight.** This is the one workflow here that spends money with nobody
watching, so every new input owes it a free refusal. Three of the six are about the path —
whatever this names is read file by file and fed to a model, so an absolute path or a `../`
climb would let one dispatch point the panel at anything on the runner. Containment is
asserted rather than arranged: an escaping string is refused by name, never normalised, and
the check is repeated after `realpath`, because `git` stores symlinks and `lenses-x -> /etc`
passes every lexical test.

The fourth is the one `buildConfig` would accept: **a manifest declaring zero lenses.** An
empty array maps to an empty array, the hash computes, and the replay buys a panel that
reviews nothing, finds nothing, and reports a *complete* run — the shape of failure the
`dryrun-` prefix already exists to prevent. A test asserts `buildConfig` really does accept
it, rather than claiming so.

The fifth runs the real `buildConfig`, because it is free, deterministic and the same call
the runner makes: a mistyped `effort` or a missing rubric then costs one runner-minute
instead of a paid leg.

**The sixth covers the directory's CONTENTS**, which containment on the directory does not.
Every file in a lens dir is opened, and a rubric's bytes are both fed to a model and inlined
into `config.snapshot.json` as `rubric_text` — which the collect job pushes to a public
repository. So one committed `correctness.md -> /somewhere/on/the/runner` would exfiltrate
that file with the directory-level check passing cleanly; reproduced before the guard
existed. Every entry is scanned, not only the ones the manifest names, and one level is the
whole surface because `assertSafeLensId` forbids a separator in a lens id. The rule is "no
symlinks" rather than "no escaping symlinks": a rubric that is secretly some other committed
file is still not the rubric its `config_hash` claims.

**Resuming a run id under a changed config is already refused**, so nothing was added for
it. `store.putRun` compares the stored `config.snapshot.json` by `config_hash` and throws
naming both hashes; `run.mjs` calls it before the adapter is built. Verified end to end with
the stub panel: the second leg refuses, the stored envelope is byte-identical, and the
runner's own banner never prints.

**Empty means the flag is not passed**, which is not the same as passing it empty:
`run.mjs` takes its default with `??`, which an empty string satisfies, and
`path.resolve("")` is the working directory — so an unconditional `--lenses-dir` would point
a paid replay at the workspace root. The ARGS array adds it only when non-empty, the shape
`--items` already uses, pinned by a test. The input is given no default naming a path,
because `run.mjs` owns that constant and two homes for one constant drift.

## Linked Issues

None. This is a self-contained addition to a manual workflow; it follows no issue.

## Author checklist

## Verification

Measured locally against a fresh checkout of the base commit:

| | base | this branch |
|---|---|---|
| `eval/replay-plan.test.mjs` | 25 pass / 0 fail | **37 pass / 0 fail** |
| `eval/run.test.mjs` | 57 pass / 0 fail | 57 pass / 0 fail |
| whole `agent:tests` lane | 2363 · 2357 pass · 0 fail · 6 skipped | 2375 · 2369 · 0 fail · 6 skipped |
| eslint 9.24.0 over `scripts/` | exit 0 | exit 0 |

**Every refusal was mutation-tested**: each guard removed in turn, the test that names it
goes red, the guard restored. Eleven mutations, eleven caught, including the two workflow
ones (the `-n "$LENSES_DIR"` branch, and `--lenses-dir` reaching the preflight).

⚠ **`actionlint` was not available**, so `eval-replay.yml` got YAML well-formedness checking
only — never Actions semantics. What was checked structurally: one input added, `on:` still
`workflow_dispatch` alone, `permissions:` still `actions: read` + `contents: read`, jobs
unchanged.

End to end against the real store, `2026-08-10-pilot-reviewed`, K=3:

```
  lens config  : scripts/agent/lenses (the committed default) — correctness=claude-opus-5 … docs=claude-sonnet-5
  config_hash  : sha256:1c7853debf4edf92646d2299b0c924cb48cca89d6bb68b81648c57508a762f01
  EXPOSURE     : $8.00 cap PER REPLICATE x 3 = $24.00 maximum for this dispatch
```

and an all-sonnet variant prints `sha256:d821c6fd…`, i.e. the arms are separable before a
dollar is committed. No dispatch was run.

## Risk Assessment

- **User-facing risk:** none for users of the product; this workflow is `workflow_dispatch`
  only and nothing else triggers it. For an operator, the new input is optional and an empty
  value reproduces today's behaviour exactly — the flag is not passed at all, so `run.mjs`'s
  own default applies unchanged.
- **Data/security risk:** this is the change's real surface, and it is why four of the six
  refusals are about the path. The input is resolved and handed to a process that reads a
  directory and feeds it to a model, so an unbounded value would be arbitrary file read on
  the runner — and, because a rubric is inlined into the run's snapshot, arbitrary file read
  into a PUBLIC repository. Containment is asserted lexically, again after `realpath`, and
  once more over every entry in the directory, since resolving the directory says nothing
  about the files in it. Absolute paths are refused outright, and every direction is tested.
  What containment does NOT prove is provenance — "in the workspace" is not "committed by
  this repository", and `.eval-store/` and `node_modules/` are both in the workspace. Neither
  is reachable as a lens config today (neither holds a `lenses.json`, which is checked), and
  the docblock says so rather than the error message claiming otherwise. `permissions:` is unchanged — no
  `contents: write`, no `checks: write` — and the existing test asserting no write scope
  anywhere in the file still passes. The store-write token still sits in one job that runs no
  model.
- **Rollback plan:** a revert. Nothing consumes the input yet, no dispatch has used it, and
  no data in the store was produced under it. A revert leaves `run.mjs --lenses-dir` exactly
  as it was.

## Notes for Reviewers

- No UI changes.
- **AI assistance:** written with Claude Code — the workflow edit, `planLensConfig`, all
  eleven new tests, and the mutation harness. The test counts, the mutation results and the
  two `config_hash` values above are measured output, not estimates.
- **Deliberately out of scope:** no variant lenses directory is committed. Which one to build
  is a per-experiment choice; committing one here would make a particular comparison part of
  the lane.
- **Follow-up:** the lane cannot enforce the one thing that makes a comparison valid — that
  both arms are dispatched against the same panel commit. The input description, the workflow
  header and the preflight banner all say so, and that is as far as a form can go.

## Summary by CodeRabbit

* **New Features**
  * Added an optional lens configuration directory when starting evaluation replays.
  * Replay planning now validates selected lens configurations and reports the active configuration and hash.
  * Custom configurations are forwarded to replay execution only when specified.

* **Bug Fixes**
  * Added validation for invalid, missing, empty, unsafe, or malformed lens configurations before execution.

* **Tests**
  * Expanded coverage for default and custom configurations, path safety, validation errors, and workflow forwarding.

### DIFF
--- a/.github/workflows/eval-replay.yml
+++ b/.github/workflows/eval-replay.yml
@@ -18,6 +18,17 @@
 # and the run records which panel commit and which lens configuration produced it, so
 # a number can be re-derived by someone who was not there.
 #
+# A DISPATCH CAN REPLAY A LENS VARIANT, and the mechanism is entirely the store's
+# existing key. `lenses_dir` names a directory committed to the branch being
+# dispatched; `model`, `effort` and the rubric bytes are all inside `config_hash`, so
+# a variant's runs land under an identity of their own, while `panel_digest` — which
+# hashes panel modules and no lens file — does not move. What that buys is a
+# comparison; what it does NOT buy is a comparison against an OLD run, because the
+# panel moves on its own schedule and a variant replayed today against a control
+# replayed weeks ago confounds the two. Both arms, same day, or the number means
+# nothing. The preflight prints each arm's `config_hash` so the two can be told apart
+# afterwards.
+#
 # WHERE THE OUTPUT GOES. Into `dlgpdmsly2/wafflebase-agent-eval`, the same store the
 # capture collector writes to, through the same kind of scoped token — and NOT into
 # this repository. Git history is permanent; a replay envelope committed here could
@@ -95,6 +106,14 @@ on:
         description: "Optional comma-separated item subset, e.g. pr-524. Empty replays the whole corpus. A typo is refused, not skipped."
         required: false
         default: ""
+      # NO DEFAULT NAMING A PATH, deliberately. `run.mjs` already owns the default
+      # lenses directory; repeating it here would give the same constant two homes
+      # and the two would drift. Empty means "pass no flag", so the runner's own
+      # default applies and this input cannot be the thing that changed it.
+      lenses_dir:
+        description: "Optional path INSIDE this checkout to an alternative lens configuration, e.g. scripts/agent/lenses-sonnet. Empty uses the committed panel config. A variant is a DIFFERENT REVIEWER — it lands under its own config_hash and pools with nothing else — so a comparison needs BOTH arms dispatched on the same day, against the same panel commit. The preflight prints the config_hash before anything is spent."
+        required: false
+        default: ""
 
 # READ ONLY, exactly as the collector is. The write capability is a secret aimed at a
 # different repository, not a permission held here. No `contents: write`, no
@@ -148,8 +167,12 @@ jobs:
         with:
           node-version: 22.x
 
-      # No `npm ci`. The preflight reads JSON and does arithmetic; it imports only
-      # `node:` builtins and two sibling modules.
+      # Still no `npm ci`. The preflight reads JSON, does arithmetic, and builds the
+      # lens config — and that last one pulls in `config-build.mjs` and, through it,
+      # most of the agent's module graph. None of it needs an install: every import
+      # on that path is a sibling `.mjs`, and the SDK is loaded lazily by `ask.mjs`
+      # at call time rather than at import time. `eval/config-build.test.mjs` runs in
+      # the same no-install lane and would fail here first if that stopped being true.
       - name: Preflight the dispatch
         id: plan
         env:
@@ -159,6 +182,7 @@ jobs:
           PANEL: ${{ inputs.panel }}
           REPLICATES: ${{ inputs.replicates }}
           ITEMS: ${{ inputs.items }}
+          LENSES_DIR: ${{ inputs.lenses_dir }}
         run: |
           set -euo pipefail
           node scripts/agent/eval/replay-plan.mjs \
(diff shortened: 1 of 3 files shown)


dlgpdmsly2:
@claude review

hotsunchip (approved):

PR #957, opened 2026-08-24
Add worksheet settings to the v1 API: freeze, hidden, merges

ggyuchive:
## Summary
- New ApiV1WorksheetController at .../tabs/:tabId with GET+PUT for freeze ({rows,cols}), hidden ({rows,columns} 0-based indices), and merges ({merges:{"<ref>":{rs,cs}}}).
- parseFreeze/parseHidden/parseMerges validate the body and reject bad shapes with a 400 before any write.
- Writes seed the canonical root via initialRoot; reads are readonly and return plain JSON (a Yorkie value's toJSON double-encodes arrays and objects as strings otherwise).

## Why
The v1 API exposed cells and tabs but no worksheet-level layout state a user controls in the editor.

## Linked Issues

Fixes #

## Author checklist

- [X] I searched existing issues and PRs and confirmed this is not a duplicate.
- [X] I read every line of this diff and can explain why each change is necessary.
- [X] Changes follow the conventions in the touched packages; any deviations are explained in *Why* above.
- [X] If AI tools assisted with this PR, I noted where in *Notes for Reviewers* below.

## Verification

forced the whole suite.

Skip reason (if applicable):

## Risk Assessment

- User-facing risk:
- Data/security risk:
- Rollback plan:

## Notes for Reviewers

- Follow-up work (if any):

## Summary by CodeRabbit

- **New Features**
  - Added API support for managing worksheet freeze panes.
  - Added API support for retrieving and updating hidden rows and columns.
  - Added API support for retrieving and updating merged-cell settings.
  - Added validation for worksheet settings, including sensible defaults and clear rejection of invalid values.

- **Bug Fixes**
  - Prevented worksheet settings from being applied to unsupported documents or missing tabs.

### DIFF
--- a/packages/backend/src/api/v1/api-v1.module.ts
+++ b/packages/backend/src/api/v1/api-v1.module.ts
@@ -2,6 +2,7 @@ import { Module } from '@nestjs/common';
 import { ApiV1DocumentsController } from './documents.controller';
 import { ApiV1TabsController } from './tabs.controller';
 import { ApiV1CellsController } from './cells.controller';
+import { ApiV1WorksheetController } from './worksheet.controller';
 import { ApiV1DocsContentController } from './docs-content.controller';
 import { ApiV1ImagesController } from './images.controller';
 import { ApiV1ImageReadController } from './image-read.controller';
@@ -34,6 +35,7 @@ import { OptionalCombinedAuthGuard } from '../../api-key/optional-combined-auth.
     ApiV1DocumentsController,
     ApiV1TabsController,
     ApiV1CellsController,
+    ApiV1WorksheetController,
     ApiV1DocsContentController,
     ApiV1ImagesController,
     ApiV1ImageReadController,
--- a/packages/backend/src/api/v1/worksheet.controller.spec.ts
+++ b/packages/backend/src/api/v1/worksheet.controller.spec.ts
@@ -49,9 +49,19 @@ describe('ApiV1WorksheetController', () => {
   });
 
   it('setHidden writes hidden rows/columns', async () => {
-    await controller.setHidden(WS, DOC, 'tab-1', { rows: [1, 3], columns: [0] });
+    await controller.setHidden(WS, DOC, 'tab-1', { rows: [1, 3], columns: [1] });
     expect(ws().hiddenRows).toEqual([1, 3]);
-    expect(ws().hiddenColumns).toEqual([0]);
+    expect(ws().hiddenColumns).toEqual([1]);
+  });
+
+  // Indices are 1-based, like the A1 refs the rest of the v1 API speaks.
+  // `Sheet.loadHiddenState` keeps only `>= 1`, so a 0 that reached the CRDT
+  // would be dropped on load and make every index look off by one.
+  it('setHidden rejects a 0 index instead of writing one the engine drops', async () => {
+    await expect(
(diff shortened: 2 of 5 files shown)


hackerwins:
Reviewed ahead of the v0.6.6 release and pushed a fix commit rather than
blocking the PR, since the release was waiting on it. Three issues, all
confirmed against the engine source rather than inferred — please check the
call I made on the third, since it changes your documented contract.

**1. Merge keys were never validated.** `parseMerges` validated `rs`/`cs` and
copied the key through verbatim. `Sheet.rebuildMergeCoverMap()` calls
`parseRef(anchorSref)` on every key with no try/catch, and `parseRef` throws
`Invalid Reference`. So `PUT .../merges` with `{"merges":{"1":{"rs":1,"cs":1}}}`
returned 200 and then threw on load for every collaborator, permanently —
repairable only through another API call.

Validating with `parseRef` alone turned out to be insufficient, which a test
caught: it scans to the first digit, so `"A1:B2"` parses as `A1` and does not
throw. That stores a key that is not a cell, and `rebuildMergeCoverMap`'s
`sref === anchorSref` guard then never matches. Keys now have to round-trip —
`toSref(parseRef(ref)) === ref`.

**2. Spans and freeze counts were unbounded.** `rebuildMergeCoverMap` walks
`rs * cs` and stores one Map entry per covered cell, and the frozen quadrants
render every frozen row/column with no viewport clipping. Either one turns a
200 into a document that never paints — and a user who cannot paint cannot
reach the menu that would undo it. Both are bounded by the grid now
(1000000 × 18278), plus a ceiling on cells covered per merge, because a single
in-grid span is still 1.8e10 iterations.

**3. Hidden indices — 0-based vs 1-based.** The doc comment and the PR body say
0-based, but `Sheet.loadHiddenState` keeps only `row >= 1`. Sending
`{"rows":[0,1,2]}` per the documented contract hid rows 2 and 3, not 1–3, and
row 1 could never be hidden at all. `GET /hidden` echoes the stored values, so a
round-trip does not reveal it.

I made these **1-based**, matching the A1 refs the rest of the v1 API speaks,
and made `0` a 400 rather than something accepted and silently dropped. The
alternative — converting 0-based to 1-based in the controller — keeps your
original contract, so say the word if you prefer it. Your controller spec
asserted the buggy shape (`columns: [0]`), so I corrected it rather than
extended it.

Two of the new tests earned their place by failing first: `"A1:B2"` is what
showed `parseRef` was too lenient to validate with, and the `__proto__` case had
to be built with `JSON.parse` — an object literal sets the prototype and creates
no own key, so it cannot reproduce what a request body actually does.

`pnpm verify:fast` green; worksheet suites 35/35.

Not addressed, worth a follow-up either here or separately:
- GET on a nonexistent tab returns 200 with fabricated defaults instead of 404,
  unlike every sibling GET and unlike this PR's own PUTs.
- `ws.merges = merges` replaces the shared Yorkie map container, so a
  concurrent editor's merge is lost wholesale. Defensible as PUT semantics, but
  worth being deliberate about.
- The new controller is not in the `is mounted on every v1 controller`
  enumeration in `api-key-write-scope.guard.spec.ts`. The guard *is* on the
  class, so there is no live auth gap — but that spec exists to catch exactly
  that, and it does not cover this controller.
- `docs/design/rest-api.md` and the v1 tables in `packages/backend/README.md`
  gain no entry for these six routes.

PR #960, opened 2026-08-24
Preview a component in the design editor, and fix the pipeline behind it

hotsunchip:
Part of #700 — the design-editor extraction. Follows #916 / #917.

## What this is

`components` mode had a catalogue and a bindings panel and **no preview**: the
centre pane kept showing whichever route was last open, so the panel described a
`Button` while a login page was on screen. This adds the preview, and fixes the
pipeline behind it — six mechanisms that were built, wired at both ends, and
silently doing nothing.

The frontend change is **one file, 18 lines** (`document-list.tsx`). Everything
else is the dev-only editor, its consumer config, tests, and a hand-off skill.

## How a component preview works

```
shell (host)                     frame (iframe, own realm)
  ComponentList  ─ pick ─▶
  App ─ componentFrameUrl() ──▶  scene-entry reads the URL
                                   └─ ComponentFrame
                                        ├─ loadComponentFile(file, side)   ← generated loaders
                                        ├─ <Providers mocks=[…]>           ← consumer's
                                        └─ recipe?.render(C, props)        ← consumer's previews.tsx
  ◀── wb:ready / wb:view / wb:error ──
```

The frame is a pure function of its URL — variant, forced state, mock props,
icon slot, theme all travel as query parameters, so a reload is a reset.

### The new seam: `previews`

The generic mount is `<Component {...generatedProps}>{name}</Component>`. Right
for a styled primitive, wrong for most of a shadcn `ui/` directory: a menu item
outside its menu throws, a slider with no width cannot slide. That knowledge is
the **consumer's**, so it is a plugin option, not a table in the plugin:

```ts
designEditor({ …, previews: path.join(HERE, 'src/scenes/previews.tsx') })
```

`Record<componentName, { props?, frame?, children?, render? }>`. Absent, every
component falls through to the generic mount — the option is additive.

## What a reviewer should look at

**`plugin/index.ts` — the safelist is a file on disk now.** Non-obvious and the
reason it is written that way: Tailwind builds its compiler from the entry
**once per dev server** and afterwards only re-scans source files, so a
`@source inline(...)` appended in `transform`/`load` is never seen again.
Measured: `load` ran with the directive and the served bytes did not move.
Neither `reloadModule`, `invalidateModule` nor a synthetic watcher event helps —
what Tailwind reacts to is a file whose mtime changed. So the plugin owns a file
outside the project, injects one `@import` of it into the entry (present at the
first compile, so Tailwind watches it), and rewrites it. Without this, every
class the editor composes at runtime has no CSS rule.

**`plugin/protocol.ts` — `planFileOf`.** `planFiles` reported only *layout*
intents, so a `class-rewrite` contributed nothing: the frame served the
component unpatched and the publish reloaded no module. One resolver is now
shared by `planFiles` and `scene-patch`'s `load`; two readings of "which file
does this intent target" is how the gap opened.

**`server/extract.mjs` — two attribution rules.** A cva attaches to the
component whose **body** calls it (`sidebar.tsx` handed its one table to all 24
exports, `useSidebar` included), and an exported PascalCase `const` is recorded
as a component or an alias — shadcn declares half its roots as
`const ContextMenu = ContextMenuPrimitive.Root`. Together these let the manifest
list every `components/ui/*` file instead of eight hand-picked ones.

**`packages/frontend/src/app/documents/document-list.tsx` — a real product
bug.** `folders = []` is a fresh array per render; through `childFolders` →
`rows` it reaches `useReactTable`'s `data`, whose `onChange` queues
`resetPageIndex()` on a microtask, which sets state, which renders again — a
microtask loop that never yields, so the tab stops responding. `/documents`
omits the prop, so the app was one state change away from it. Found by mounting
the component with only its required props, which is what this editor does.

**`scenes/frame-protocol.ts` — the validator table.** `wb:view` was in the
`FrameMessage` union and the host's switch but not in `FRAME_SHAPES`, and an
unknown message is dropped silently — pan and zoom never worked. A test now
asserts every union member has a runtime validator, because the type-level union
is not the runtime contract.

## Structure of the catalogue

A composite is **one subject**, detected structurally: the shortest export whose
name prefixes every other in its module (`DropdownMenu`, `Card`, `Select`). Its
parts live in the **Layout tab**, not the catalogue — a composite's own tree is
one `div` and its anatomy is sibling exports, so the structure panel was
reporting that the thing has no structure. `document-list.tsx` (`DateCell`,
`DocumentList`) has no such name and stays flat.

The two catalogue groups are **Primitives** / **App components**, derived from
whether the component has required props — the same fact that predicts whether a
bare mount succeeds.

## Hand-off

`.claude/skills/design-sandbox-bringup/SKILL.md` is the point of the whole
extraction: the six files a consumer writes, a verifiable bring-up order, and a
trap table pairing every symptom above with its real cause. Written so a foreign
project can stand this up without repeating the findings.

## Test plan

- `pnpm --filter @wafflebase/design-editor test` — 1111 tests. New: safelist
  entry detection + `load` hook + plugin order; `planFileOf` for class-rewrite;
  every `FrameMessage` has a validator; cva body attribution; arrow-function and
  alias components; mock-prop generation.
- `pnpm --filter @wafflebase/design-sandbox test` — 130 tests.
- Typecheck + lint clean in both packages.
- Driven in a real browser against wafflebase's own frontend: forced `hover:`
  paints `oklab(… / 0.9)` instead of transparent; a binding rebind repaints the
  preview in ~6s; a zoom anchor moves 0.2px across three steps; a drag on the
  ground pans and a drag on the component does not; the class editor closes when
  its frame is replaced.

### Known limits

- The 5s frontend test timeouts (`TextEditSection`, `csv-doc-size`) are
  load-dependent and pass standalone; neither file is in this diff.
- A composite's parts still mount through recipes rather than being derived —
  the plugin cannot know that a menu item needs its menu.


hotsunchip:
@claude review.

PR #965, opened 2026-08-25
Record the panel's deferred findings in an advisory check run

dlgpdmsly2:
## Summary

Adds an advisory check run, `agent-deferred-findings`, carrying the findings the gate did not act
on — native `minor`/`nit`, and `critical`/`major` demoted to `lane: "backlog"`. One check run for
the whole panel, built from the same `verdict.json` the gating channel reads.

New `scripts/agent/deferred-findings.mjs` + tests, and two `continue-on-error` steps in
`agent-review-panel.yml`. `output.text` and `permissions:` untouched. Nothing consumes it yet.

## Why

Those findings are durable but **addressed to nobody**: `verdict.json` keeps them, yet every
channel a tool reads drops them — `output.text` is filtered to `critical|major` +
`lane !== 'backlog'`, `proseOnly` cuts each lens body at its first finding section, and
carry-forward reads `output.text`. "What did this round decide not to act on" has no answer on the
PR.

**Not by widening `output.text`.** The comment above its filters names three consumers, and the
non-convergence detector needs a quantity that *shrinks* as the fixer works — a pile nobody is
obliged to fix never shrinks. So: a second channel, with no consumer that gates.

Three points a reviewer should check:

- **It cannot gate.** `conclusion: neutral`, never in `required`, and named outside
  `agent-review-` — `set-state.mjs` enumerates every check run on the commit, filters that prefix,
  and sets `lensBlocked` from what it finds. `assertAdvisory` throws rather than coercing.
- **It stamps the rubric generation** — `git -C .trusted rev-parse HEAD`, once per run. `severity`
  means whatever the rubric in force said it meant, so without this a record written after a
  rubric change is indistinguishable from one written before, and an archive's value grows with
  time. Not the PR head: `.trusted` is checked out at `ref: main`, so the head names a tree that
  need not hold the rubrics that ran.
- **The trim states what it dropped**, the convention `buildChecklist` already owns and
  `output.text`'s trim lacks. `text` is an object so it has somewhere to say it is partial.

Provenance is stored as primitives — a native minor keeps **no** `lane`, `novelty.origin` and
`surface.scope` stay separate, and there is no derived "why deferred" field (`annotateFindings`
argues against exactly that; `demotedBy` is lossy by design). `line` comes from `findingLocation`,
since the gating projection has none.

## Linked Issues

None.

## Author checklist

## Verification

Measured from the committed tree, both trees sharing one `node_modules`:

- **`agent:tests` 2508 + 57 = 2565, 0 fail, 0 skip** — baseline at `70e5be9ad30e` 2475 + 57 =
  **2532**, so **+33**, exactly the new test file. **`eslint scripts` exit 0** (9.24.0), as baseline.
- **28/28 mutations caught by the specifically-named test**, tree restored byte-identical —
  never-gates guard (5), trim tally (6), provenance (7), generation stamp (4), population (3),
  fail directions (2), the new step's `retries` (1).
- ⚠ **`actionlint` is not installed here**, so the workflow got YAML well-formedness plus a
  structural diff — a substitute, not an equivalent: triggers, job list and **every job's
  `permissions` identical**, every pre-existing step byte-identical, 2 added / 0 removed,
  `checks.create` 2 → 3.
- ⚠ **Never run against a live panel.** The end-to-end test drives the CLI over a synthesised
  `.agent-review` tree, so the workflow wiring and the API call are unexercised.

## Risk Assessment

- **User-facing risk:** one extra `neutral` check run per managed PR. It cannot block a merge, and
  both steps are `continue-on-error`, so neither can redden a paid review round. The honest
  residual risk is **noise**.
- **Data/security risk:** no new permission — `checks: write` was already held and `permissions:`
  is byte-identical. Reads `verdict.json` and the trusted `lenses.json`, writes one `$RUNNER_TEMP`
  file, no network call. It does put model-authored `summary`/`evidence` into a check body — the
  same untrusted-text path the per-lens bodies already take, clipped to the same widths.
- **Rollback plan:** revert. Nothing reads the channel, nothing imports the module.

## Notes for Reviewers

- **AI assistance:** written with Claude Code — module, tests, workflow steps, task doc, this body.
  Line numbers and quoted comments were read from the source at `70e5be9ad30e`, not recalled.
- **Look first at** `assertAdvisory` and `buildDeferredText`.
- **Not in this PR:** `output.text`'s own trim drops trailing findings silently. Real, but it
  changes the gating channel, so it belongs in its own change.
- The new module is in neither `PANEL_FILES` nor `NOT_PANEL_FILES` — that set's domain is *"local
  imports reachable from `PANEL_ENTRY`"* and the panel does not import it, same as `fix-brief.mjs`.
- **Follow-up:** nothing consumes this yet; the intended consumer is a batch pass over the pile,
  off the critical path.

## Summary by CodeRabbit

* **New Features**
  * Added an advisory review check that records findings deferred from the review gate.
  * Deferred findings are available in machine-readable and human-readable formats, with accurate totals and truncation details.
  * Records include review context and generation information for traceability.
  * The advisory check does not affect required review status.

* **Documentation**
  * Added design documentation describing deferred findings, payload limits, and verification behavior.

* **Tests**
  * Added coverage for classification, formatting, truncation, malformed inputs, and end-to-end generation.

### DIFF
--- a/.github/workflows/agent-review-panel.yml
+++ b/.github/workflows/agent-review-panel.yml
@@ -1085,8 +1085,18 @@ jobs:
       # the conclusion is `neutral` (never `failure`), the name is NOT pushed into
       # `required` above, and it sits outside the `agent-review-` namespace that
       # `set-state.mjs` enumerates by prefix to decide `lensBlocked`.
+      # `!cancelled()`, not `always()`, and it matches "Post per-lens check runs"
+      # above rather than the observability steps below. A cancelled panel has not
+      # finished routing, so the lenses that happen to have written a verdict.json
+      # are an arbitrary prefix of the round — and this channel's whole contract is
+      # that its counts are complete unless the tally says otherwise. Publishing
+      # `total: 4` from a cancelled round is the silent-partial failure the omission
+      # tally exists to prevent, wearing a different hat. An ordinary FAILURE still
+      # publishes: a lens that crashed while others deferred findings is a round
+      # worth recording.
       - name: Record the deferred findings
-        if: always() && steps.pr.outputs.number != ''
+        if: >-
+          !cancelled() && steps.pr.outputs.number != ''
         continue-on-error: true
         # [ -f ] guard: .trusted is the DEFAULT BRANCH's copy, so until this script
         # has merged the step must skip quietly, not red every panel run.
@@ -1111,7 +1121,8 @@ jobs:
             --out-json "$RUNNER_TEMP/deferred-findings.json"
 
       - name: Post the deferred-findings check run
-        if: always() && steps.pr.outputs.number != ''
+        if: >-
+          !cancelled() && steps.pr.outputs.number != ''
         continue-on-error: true
         uses: actions/github-script@v7
         with:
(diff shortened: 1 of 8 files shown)


dlgpdmsly2:
@claude review

hotsunchip (dismissed):

hotsunchip (approved):

Issue #482, opened 2026-07-16
Allow text selection and link interaction in Docs viewer mode


kokodak:
### What happened:

In Docs viewer mode, readers cannot select text by dragging, copy text, or open hyperlinks embedded in the document. This limits basic document review and navigation even though these actions do not modify the document.

### What you expected to happen:

Viewer mode should remain read-only for document editing while still allowing readers to:

- Select text by dragging.
- Copy selected text.
- Click hyperlinks to open their destinations.

### How to reproduce it (as minimally and precisely as possible):

1. Create a Docs document containing regular text and a hyperlink.
2. Open the document through a viewer-role share link.
3. Try to select and copy text.
4. Try to click the hyperlink.
5. Observe that none of these interactions work.

### Anything else we need to know?:

Read-only mode should prevent content changes without disabling non-editing interactions needed to read and use the document.

### Environment:

- Operating system: Not specified
- Browser and version: Not specified

hackerwins:
@claude review this.

Issue #478, opened 2026-07-15
Parse Markdown content when pasting into Docs


kokodak:
### What happened:

When Markdown source content is copied and pasted into Docs, it is inserted as plain text instead of being converted to the corresponding rich-text formatting.

For example:

- `# Heading` remains literal text instead of becoming a heading.
- `[Wafflebase](https://wafflebase.io)` remains Markdown source instead of becoming a hyperlink.
- `- Item` remains literal text instead of becoming a bulleted list.

### What you expected to happen:

Docs should recognize Markdown content on paste and insert supported elements using the corresponding formatting, including headings, hyperlinks, and bulleted lists.

### How to reproduce it (as minimally and precisely as possible):

1. Open a document in Docs.
2. Copy Markdown source containing a heading, a link, and a bulleted list.
3. Paste it into the document.
4. Observe that the Markdown source is inserted verbatim without applying formatting.

### Anything else we need to know?:

The current behavior differs from the common expectation that pasted Markdown content is converted into editable rich-text elements.

### Environment:

- Operating system: Not specified
- Browser and version: Not specified

blurfx:
I think overriding the paste function to allow the insertion of Markdown documents does not seem like a good choice.

Does all pasted text have to go through the Markdown parser? if then there would a false positives. also maybe it needs a new feature  'Paste as Plain Text' to avoid that issue.

PR #984, opened 2026-08-29
Copy table-cell content and click-selected images with formatting

### LINKED ISSUE #870: Copying an image selected by clicking it puts nothing on the clipboard
### What happened:
Selecting an image by **clicking** it — the state where the resize handles
appear — and pressing Cmd+C writes **nothing** to the system clipboard. The
clipboard keeps whatever was copied before, so the next paste replays that
instead. It looks like "the image will not paste", but the copy never happened.

https://github.com/wafflebase/wafflebase/issues/726#issuecomment-5312821680

Copying an image that is selected puts it on the clipboard, and pasting it back
into a document inserts the image — the same as copying any other content.

1. Copy some text (anything) so the clipboard has a known previous value.
2. Open a docs document and insert an image via the toolbar.
3. **Click** the image so it is selected — the resize handles appear.
4. Press Cmd+C.
5. Put the caret somewhere in the paragraph and press Cmd+V.

→ The text from step 1 is pasted.

### Environment:

- Operating system: macOS
- Browser and version: Whale 4.39.410.14 (arm64)

### LINKED ISSUE #872: Copying inside a table cell loses all formatting (and images entirely)
### What happened:

Copying content that lives **inside a table cell** puts only `text/plain` on the
clipboard.

Copying inside a table cell should behave like copying in the body: bold, links,
colours and images all survive a copy-and-paste.

1. Create a docs document and insert a table.
2. Type some text in a cell and make part of it **bold**.
3. Select that text and copy it.
4. Paste it into a paragraph outside the table.

→ The text arrives unformatted. The same steps in the body preserve the bold.

https://github.com/user-attachments/assets/06cc4528-9902-4c75-8af3-3739e1ac132b

### Environment:

- Operating system: macOS
- Browser and version: Whale 4.39.410.14 (arm64)

hackerwins:
Fixes #872. Fixes #870.

## Summary

Two defects on the docs **copy** path, where the payload the editor writes was
empty or missing.

**#872 — copying inside a table cell lost all formatting and images.**
`getSelectedBlocks()` resolved both selection endpoints against `layout.blocks`,
which holds top-level blocks only. Cell content is a separate block list hanging
off the table block, so a selection anchored in a cell produced
`findIndex === -1` and the function returned `[]`; the clipboard then carried
only `text/plain`. It now resolves the start endpoint through
`layout.blockParentMap` — the same path `getSelectedText()` and `isInTable()`
already walk — and uses `resolveNestedTableLayout` so cells inside a *nested*
table resolve too. The clone/trim loop is extracted to `sliceBlockRange()` and
shared with the body path.

**#870 — copying a click-selected image put nothing on the clipboard.** An
image selected by clicking is view-local state, not a text `Selection`, so
`handleCopy` early-returned; and `imageKeyHandler`'s catch-all cleared that view
state before the browser's `copy` event even fired. `TextEditor` gains
`imageSelectionProvider` / `imageDeleteHandler` hooks, and the key handler now
consumes Cmd/Ctrl+C and Cmd/Ctrl+X **without** `preventDefault` so the native
clipboard event still runs while the selection survives.

Two things fell out of that work:

- Extracting the removal into `deleteSelectedImageInline()` closed a **read-only
  mutation hole**: `imageKeyHandler` runs before `handleKeyDown`'s read-only
  guard, so Delete on a selected image was mutating a read-only document.
- The comment claiming read-only Copy is impossible was wrong — a read-only
  editor does construct its `TextEditor` and hidden textarea, and
  `editor-read-only.test.ts` has asserted a working read-only copy since #482.
  Copy is now offered in the read-only context menu; Cut stays gated.

## Review findings applied

- **Caps Lock silently reintroduced #870.** `imageKeyHandler` compared the raw
  `e.key`, but the browser reports the *modified* character, so Cmd/Ctrl+C
  arrived as `'C'`, missed the guard, and fell into the catch-all that clears
  the selection. `TextEditor.handleKeyDown` already normalizes for exactly this
  reason. Pinned with a test that fails without the fix.
- **`imageSelectionProvider` could throw out of the `copy` listener.** It called
  `doc.getBlock()`, which throws; `findBlock` is the non-throwing variant, so
  the `if (!block)` guard was dead code. It runs before `preventDefault()`, so a
  block a remote peer had just deleted would throw and let the default copy wipe
  the user's system clipboard.

## Markdown paste (#478) was withdrawn from this branch

A markdown-paste parser was implemented here and **removed before this PR**;
`clipboard.ts` is byte-identical to `main`. Review reproduced two defects a
regex pass cannot close:

1. **Text corruption in the modal use case.** The italic alternative used `\S`,
   which matches `*` itself, so it paired a `*` from one token with a `*` from
   the next and *deleted* both:

   ```
   # API                        heading   | API
   def f(*args, **kwargs):  →   paragraph | def f(args, *kwargs):
   return a*b*c                 paragraph | return abc
   ```

   The trigger is a single `# ` heading anywhere in the paste, so pasting a
   README or an LLM answer containing a snippet corrupted it. This was a
   regression — previously no `|`-table meant `null` and the paste took the
   plain-text path intact.

2. **Quadratic backtracking.** `'['.repeat(n)` on one line cost 0.6 s at
   n=16,000, 3.3 s at 32,000, 6.7 s at 64,000 — and returned `null`, so the
   whole cost bought a plain-text paste. `LARGE_PASTE_WEIGHT_THRESHOLD` only
   paints a toast and runs the same synchronous job, with no cancel.

The root gap is structural: there is no code-fence handling, and there cannot
be until `BlockType` gains `code-block` (roadmap 3.3, not started). Full
findings are recorded in the task doc as the spec for a redo, so #478 stays
open.

## Test plan

- `pnpm verify:fast` green (11 lanes, `EXIT=0`).
- New `packages/docs/test/view/copy-integrity.test.ts` (8 cases): 6 of 7 failed
  before the fix. Covers styled and image-bearing cell copies, boundary
  trimming, multi-block cells, the image payload and its paste round-trip, text
  selection winning over image selection, and the Caps Lock regression.
- `docs-context-menu.test.tsx` extended for the read-only Copy entry.

## Known limitations

`getSelectedText()` still resolves via `layout.blocks.find`, so a nested-table
cell copy produces a correct rich payload but an empty `text/plain` — pasting
into an external app yields nothing. Copying inside a header/footer has the same
shape of bug one region over. A copied image carries no `image/png` flavour, so
it cannot leave wafflebase. All three pre-date this branch.

## Summary by CodeRabbit

* **New Features**
  * Copy and cut now support clicked images, including removing images after cut.
  * Copying content within table cells, headers, and footers preserves formatting and images.
  * Read-only documents offer Copy for selected content and images.
  * Read-only image selections display without resize handles.
  * Read-only linked images continue to open their links.
  * Read-only viewers can search documents without access to replacement controls.

* **Bug Fixes**
  * Prevented stale image selections after edits, replacements, undo, and document updates.
  * Improved shortcut handling and clipboard behavior when access is unavailable.

### DIFF
--- a/docs/design/docs/docs-context-menu.md
+++ b/docs/design/docs/docs-context-menu.md
@@ -53,10 +53,37 @@ still handles right-click (distinct context; table text isn't spell-checked).
 - **Frontend** (`DocsContextMenu`): one `contextmenu` listener on the
   editor container. Bails when `editor.isInTable()`. Computes group
   visibility first and **returns without opening** if every group is empty
-  (so read-only / nothing-to-offer right-clicks show neither an overlay
-  nor the native menu). Async suggestions are guarded by a generation ref
-  bumped on every open. Lifecycle (outside-mousedown + Escape close,
-  `offsetWidth/Height` viewport clamp) mirrors `DocsTableContextMenu`.
+  (so nothing-to-offer right-clicks show neither an overlay nor the native
+  menu). Async suggestions are guarded by a generation ref bumped on every
+  open. Lifecycle (outside-mousedown + Escape close, `offsetWidth/Height`
+  viewport clamp) mirrors `DocsTableContextMenu`.
+
+### Read-only
+
+Per-entry gating, not a whole-menu one:
+
+| Entry | Shown when |
+|-------|------------|
+| Spell suggestions | Editable **and** a misspelling is under the pointer |
+| Cut | Editable **and** a text selection |
+| **Copy** | **A text selection — read-only included** |
+| Paste | Editable |
+| Add link / Add comment | Editable |
+
+Copy is deliberately *not* gated on `!readOnly`. A read-only editor still
+constructs its `TextEditor` and hidden textarea, so `editor.copy()` works
+there — the same reason `handleKeyDown` lets plain Cmd/Ctrl+C through in
+read-only mode. A viewer right-clicking a text selection therefore gets a
+one-entry clipboard group; with no selection every group is empty and the
+menu does not open at all.
+
+The gate is `getActiveSelection()`, which is a **text** selection. A
+click-selected image is view-local state the menu does not read, so the
+menu never offers Copy for an image in either mode — a viewer copies an
+image with Cmd/Ctrl+C after clicking it (see
+[docs-image-editing.md](docs-image-editing.md)). Wiring an image selection
+into this gate is a straightforward follow-up and deliberately not part of
+the read-only Copy change.
 
 ## Risks and Mitigation
 
--- a/docs/design/docs/docs-image-editing.md
+++ b/docs/design/docs/docs-image-editing.md
@@ -266,9 +266,18 @@ hidden textarea**. Nothing else does: the image mousedown handler
 `preventDefault()`s and stops propagation before `TextEditor.handleMouseDown`
 can focus, and a read-only editor never focuses on mount at all. Without
 that call a clicked image on a read-only document receives neither the
-keydown nor the browser's `copy` event, so the context menu's read-only
-Copy entry would offer something that does nothing. Focus mutates nothing —
-every write still goes through a `readOnly`-gated path.
+keydown nor the browser's `copy` event, so `Cmd/Ctrl+C` over a
+click-selected image would silently do nothing — on a read-only document
+that is the *only* way to copy the image. Focus mutates nothing — every
+write still goes through a `readOnly`-gated path.
+
+The **context menu is not** that other way, in either mode. Its Copy entry
+is gated on `hasSelection = !!editor.getActiveSelection()`, and
+`getActiveSelection()` returns `null` for a click-selected image — an image
+selection is view-local state, not a text range. So right-clicking an image
+offers no Copy at all. Teaching the menu about image selections is a real
+gap and a separate change; nothing here should be read as claiming it
+already works.
 
 ### Read-only image selection
 
--- a/docs/design/docs/docs-image-editing.md
+++ b/docs/design/docs/docs-image-editing.md
@@ -242,6 +242,18 @@ the same way `handleKeyDown` does it: the browser reports the *modified*
 character, so Caps Lock turns Cmd+C into `'C'` and a raw comparison would
 fall through to the catch-all and reintroduce #870.
 
+A modifier's *own* keydown is consumed and changes nothing. `Meta` /
+`Control` / `Shift` / `Alt` (and friends) fire **before** the character they
+modify, so pressing Cmd+C delivers `key === 'Meta'` first; letting that
+reach the catch-all would clear the image selection a beat before the copy
+branch could read it, and #870 would survive on a real keyboard while a
+test synthesizing only the combined keydown passed.
+
+The catch-all — every other key clears the selection and falls through —
+re-renders before returning. A key `TextEditor` ignores would otherwise
+leave the selection overlay painted around an image that is no longer
+selected.
+
 The modifier is `e.metaKey || e.ctrlKey` — **either**, not the platform's.
 `navigator.platform` is an empty string in some browsers, so a
 platform-keyed choice picks the wrong modifier there and the shortcut lands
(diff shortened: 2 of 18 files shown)


hackerwins:
## Design-fit finding: reusing `model/range-slices.ts` — not taken, with reasoning

The blast of this PR is `getSelectedBlocks()` growing its own cell-resolution
traversal, and the review asks for it to route through `visitRangeSlices`
instead, "the module that exists specifically to be the single traversal of
which blocks a `DocRange` covers".

I checked the module before deciding, and it rules itself out for this fix.
`packages/docs/src/model/range-slices.ts` says so in its own header:

> The table case is **not recursive**: a nested table's blocks are never
> visited, so nested content is neither read nor written.

and, on the endpoint guard:

> an endpoint inside a **nested** table (…) has no index to walk from;
> nothing is visited.

So routing the copy path through it would return `[]` for a selection inside a
nested table — which is exactly the empty payload issue #872 is about, just
moved one nesting level deeper. This branch now has a regression test for that
case (`copy-integrity.test.ts`, the nested-table cell copy), and mutation-checking
confirms it is load-bearing: replacing `resolveNestedTableLayout` with a flat
`layout.blocks.find` fails that test **and nothing else** — every earlier table
test passes under the mutation, because they all use a top-level table.

One factual correction to the finding: it contrasts the two as "layout-based"
vs "doc-based", but they read *different maps* — `visitRangeSlices` uses
`doc.blockParentMap`, the new code uses `layout.blockParentMap`. It is not a
substitution.

I agree with the underlying point that two traversals is one too many. The
module's own comment already names the fix and its consequence:

> That gap is recorded in `docs/design/docs/tables/docs-nested-tables.md`;
> because there is only one traversal, closing it would close it for the read
> and the write together.

That is the right change — make `visitRangeSlices` recurse, then collapse the
copy path onto it — but it alters read *and* write semantics for every caller
in one step, which does not belong in a copy-path bug fix. Filing it against
the nested-tables doc rather than doing it here.

Also not taken, both recorded in the PR body as known limitations rather than
silently dropped: context-menu Copy/Cut over a click-selected image (real gap,
separate change), and the inlined mod-key detection (it matches the existing
idiom in the same file, `text-editor.ts:858`).

The rest of the review round is applied — see the two commits since. The
test-adequacy findings were all correct: the cut path had no test at all, and
the precedence test was vacuous.

hackerwins:
@claude rerun

The five lens failures on this head are not findings — every one reported
`[POOL_EXHAUSTED] every credential in the pool was retired` and finished in
~53s, against 10-16 minutes for a run that actually analyses the diff. They
carry no information about the code. The timestamps show why: #985's real run
finished at 08:07:1x and drained the pool; this PR's lenses started at ~08:07
and hit an empty one.

Build and test lanes on this head are green (`verify-self`, `verify-browser`,
`verify-integration`, both `verify-backend-image` architectures).

hackerwins:
@claude rerun

The pool has recovered — #985's panel just completed a full run (5 of 6 lenses
green, ~10 min each). The six failures on this head were all
`[POOL_EXHAUSTED]` in under a minute and carry no information about the diff.

Build and test lanes on this head are green: `verify-self`, `verify-browser`,
`verify-integration`, both `verify-backend-image` architectures, and codecov.

This head has three rounds of fixes the panel has never seen, including the
round-3 fix for the finding that mattered most: `imageKeyHandler` cleared the
image selection on the bare modifier keydown that precedes Cmd/Ctrl+C, so the
issue #870 fix never fired in a real browser while a suite of single
synthesized `{key:'c', metaKey:true}` events stayed green.

hackerwins:
@claude rerun

New head `4e1b9be17`. Since the last panel run this branch gained the fixer's
two commits plus one from me:

- The clearing invariant now holds on `undoFn`/`redoFn`, all six table
  structure APIs and `resetAfterDocumentReplace`, and `abortImageResizeDrag()`
  extends the same reasoning to the resize drag's captured coordinate.
- The #872 fix was resolving cells against the **body** layout while its
  siblings use `getActiveLayout()`. That was wider than the finding said: a
  plain header paragraph copied nothing, and cut was worse — `deleteSelection`
  has its own header/footer branch that consults no layout, so it removed the
  text and wrote an empty clipboard. Four `getSelectedText` call sites plus
  `getSelectedBlocks`/`getSelectedTableCells` now use `getActiveLayout()`.
- Three claims that the context menu already offers Copy for a click-selected
  image were false (`getActiveSelection()` is null for one) and are corrected
  in code and in `docs-image-editing.md`.

On the read-only resize gates the earlier "known coverage gap" note was wrong
about the cause: the blocker was three mounted canvases, not jsdom geometry.
The arming gate and the three-gate conjunction are now pinned; the move and
commit gates provably cannot be — arming is the only assignment to
`imageResizeDrag` and `readOnly` is fixed at `initialize()`, so deleting both
leaves the suite green. That is recorded as an architecture fact rather than
claimed as coverage.

hackerwins:
@claude rerun

The `agent:blocked` latch is clear now but the panel has still not run on
`4e1b9be17` — CI completed green on this head (`verify-self`,
`verify-browser`, `verify-integration`, both `verify-backend-image`
architectures, codecov) while the latch was still set, so the
`workflow_run`-triggered panel never fired for it.

This head carries the fixer's two commits plus one from me; the summary is in
the previous comment. The largest item is not in any earlier finding: the #872
cell resolution was using the **body** layout, and that was wider than the
review said — a plain header paragraph copied nothing, and cut removed the text
while writing an empty clipboard, because `deleteSelection` has its own
header/footer branch that consults no layout.

PR #985, opened 2026-08-29
Surface Docs page setup, text exports, and format painter

hackerwins:
## Summary

Four Docs capabilities were fully implemented in the engine but had no way to
reach them. This wires them up; the behaviour itself already shipped.

- **Page setup dialog.** `PageSetup` and `store.setPageSetup` were complete with
  zero UI — `docs-pagination.md` deferred it "to frontend integration". New
  `docs-page-setup-dialog.tsx` (paper size, orientation, four margins) reachable
  from the toolbar and the mobile overflow menu. Margins are stored as CSS px
  @96dpi and shown in inches.
- **Markdown and plain-text export.** `serializeMarkdown` and `serialize/text.ts`
  ship in the engine and the CLI already consumes them; the export dropdown
  offered only DOCX and PDF. Two entries added through the existing
  `runExport` → `downloadBlob` path.
- **Format painter toolbar button.** Implemented but keyboard-only
  (`Mod+Shift+C` / `Mod+Alt+V`), with no button in the toolbar at all. **No
  behaviour is reimplemented**: the two `handleKeyDown` case bodies moved into
  `TextEditor.copyFormat()` / `pasteFormat()` and the shortcuts now call them,
  so the button and the keyboard are literally the same code.

Some engine surface was unavoidable. A page-setup write is snapshot + store +
`doc.refresh()` + layout invalidation + repaint, and all five already lived
inline in the ruler's `onMarginChange`; they are extracted to one
`writePageSetup` that both callers share, so this removes a duplicate rather
than adding one.

Docs updated: `docs-pagination.md`, `docs-wordprocessor-roadmap.md`, plus the
user guide — which also had a pre-existing error, claiming pages default to A4
when `DEFAULT_PAGE_SETUP` is Letter.

## Review findings applied

Second commit, from a multi-agent review of the branch diff:

- **The new mutating API bypassed the read-only gate.** `editor.ts` carries a
  `MUTATING_METHODS` allowlist that neutralizes every mutating `EditorAPI`
  method in read-only mode, with an explicit "keep this in sync" instruction;
  `setPageSetup` and `pasteFormat` were added without it. No UI reaches them
  today (the toolbar is gated on `!readOnly`), but the allowlist's threat model
  is any holder of `api`, and `editor-read-only.test.ts` pins it per method.
  `pasteFormat` returns a boolean, so it gets an explicit `() => false` rather
  than the shared void no-op, which would report a write it never made.
- **`step={0.05}` made the browser silently veto Apply.** The margin inputs sit
  in a `<form>` with no `noValidate`, so native constraint validation runs
  first. The ruler writes arbitrary pixel margins, so a seeded value like
  `1.07"` is routine — the dialog's own validator was happy and Apply stayed
  enabled, but submission was blocked. jsdom does not run interactive
  validation, so the test suite structurally could not catch this. Now
  `step="any"`.
- **`parseInches` accepted anything `Number()` would take.** `""` is `0`, so an
  emptied field applied a 0" margin with no error; `"0x60"` applied 96 inches.
  It now requires a plain decimal.

## Test plan

- `pnpm verify:fast` green (11 lanes, `EXIT=0`); `pnpm --filter @wafflebase/frontend build` clean.
- 28 new tests across 6 files: page-setup engine round-trip and undo,
  dialog seeding / write-back / validation / cancel / re-seed-on-open,
  export entries driven through the real Radix menu, and format painter
  including a real keydown test proving the shortcuts drive the same buffer.
- Read-only cases added for both new mutating methods.

## Not in this PR

**Word count** was on the original list and is *not* here: `wordCount` appears
nowhere in any package, so it is a counting engine to build, not a surface to
wire. Flagged in the roadmap doc so the next reader does not re-file it as
"pure wiring".

## Known limitations

- The toolbar button is a two-press flow (arm, then apply) while the slides
  format painter with the same icon is one-press; matching it needs a
  pointer-driven paint mode, which is new engine behaviour. Pressing the lit
  button with no selection discards the held format silently.
- `editor.focus()` after Apply is a no-op — Radix's focus-restore-on-close wins.
  `more-fonts-dialog.tsx` documents the `onCloseAutoFocus` workaround for this
  exact case; not adopted here.
- The two new controls add ~34 px to the mobile row, which by calculation (not
  measurement) tips 360 px into horizontal scroll. Worth a look in devtools
  before merge.

## Summary by CodeRabbit

* **New Features**
  * Added Markdown and plain-text export options for Docs.
  * Added a Format Painter toolbar control with keyboard shortcuts.
  * Added Page Setup for paper size, orientation, and margins.
* **Bug Fixes**
  * Improved validation and recovery for invalid page layouts during import and export.
  * Blocked unsafe links and image sources across Markdown and content exports.
  * Preserved readable text when unsafe links are removed.
  * Improved Format Painter behavior for tables and boundary selections.
* **Documentation**
  * Updated export, pagination, roadmap, and Markdown behavior guidance.
  * Added Format Painter and Page Setup usage instructions.


hackerwins:
@claude rerun

The panel has not run on this head (`d164f350a`) at all — `agent:blocked` is
latched from an earlier round, so the gate skipped it. Four rounds of fixes on
this branch are unreviewed, including two CRITICAL security findings closed in
the last one (the CommonMark angle-bracket destination and entity-reference
bypasses of the export URL gate).

Build and test lanes on this head are green: `verify-self`, `verify-browser`,
`verify-integration`, both `verify-backend-image` architectures, and codecov.

Note for whoever reads a failure here: the pool has been returning
`[POOL_EXHAUSTED]` since ~09:11Z. Only 2 of the 9 supported slots are
registered (`CLAUDE_CODE_OAUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN_1`), so a
single busy day drains it. A lens that fails in under a minute is that, not a
finding.

hackerwins:
@claude rerun

New head `a9b459b55`; the panel has not run on it. Build and test lanes are
green (`verify-self`, `verify-browser`, `verify-integration`, both
`verify-backend-image` architectures, codecov).

This head closes the three findings from the last round that were actually this
PR's to close. The substantive one was not a doc fix: making
`EditorAPI.setPageSetup` throw left the Page Setup dialog's own validation
looser than the engine's floor, so a form the dialog accepted could reach the
setter and take down the React tree. The band is empty for integer page boxes —
which is why it was not obvious — but `resolvePageSetup` preserves fractional
dimensions, so a collaborator's CRDT write or a CLI-set geometry opens it.
`MIN_CONTENT_PX` is now shared so both sides assert the same expression.

The two lens failures on the previous head were `getDoc()` (issue #989) and the
PDF link-annotation injection (issue #990) — both predate this branch and are
filed, per the maintainer's scope decision.

Issue #998, opened 2026-08-31
Agentic office workflow: measure what an agent can do through the CLI


hackerwins:
## Context

`packages/cli/skills/` ships 17 skill files plus a `SKILL.md` that describes them
as "self-contained instructions for AI agents", and the CLI has `--format json`
and a `schema` subcommand. **There is no measurement record of how far that
surface actually gets an agent.** Every claim about it today is a guess.

This tracks a proposal from @ggyuchive (v1, 2026-08-29) to answer that by
measurement rather than assertion, plus a capability audit done while preparing
it.

Precedent: `scripts/agent/eval/` already measures the code-review panel against
CodeRabbit — freeze past PRs, replay K times, score on fixed criteria. This is
the same method pointed at office work instead of review.

## What gets built

Two new pieces; everything else is reused as-is.

| Piece | Owns | Deliberately does not own |
| --- | --- | --- |
| `wafflebase-office-agent` | System prompt, the command scope exposed to the agent, host exec flags | The model loop — borrowed from the host CLI via `--bare --restricted --system-prompt` |
| `wafflebase-office-bench` | 27 tasks + fixtures, expected end-state, scoring, run metrics | Any design that favors `office-agent` — it must score vanilla Claude Code on the same basis |

Nothing in this project builds a model loop. The host CLI (Claude Code / Codex)
already is one.

### Why office work is auto-gradable here

Coding agents work because the compiler is a free grader. Office agents usually
have no equivalent. Wafflebase does:

- **Formula engine** — the ANTLR4 recalculation is the compiler analog; a stored
  cell value is ground truth.
- **Yorkie CRDT** — checkpoint/undo already exists, so a wrecked fixture rolls
  back without new machinery.
- **`/api/v1`** — fixtures, isolation and read-back are all API-driven, which is
  what makes hundreds of repeat runs possible.
- **Document-type breadth** — sheet · doc · slides · note · board · pdf · image
  in one workspace, so cross-type tasks are constructible and gradable.

The cross-type point is also the structural difference from Claude for Microsoft
365: there, each app is bolted on separately; here one agent spans the set, and
the 9 `hard` tasks are exactly that shape.

### Bench shape

27 tasks — easy 9 / medium 9 / hard 9 — across 5 personas (Sales Ops,
Accountant, PM, Marketing, Developer) and 7 document types. Persona × complexity
is borrowed from RAG test-set generation (Ragas' `Persona(name,
[...]

PR #999, opened 2026-08-31
Agentic office workflow design, and the CLI commands it found missing

hackerwins:
Two commits: the design that asks the question, and the commands the audit
inside it found missing.

Tracking issue: #998. Original proposal by @ggyuchive (v1, 2026-08-29).

## 1. `130a91b5b` — the design

`packages/cli/skills/` describes itself as instructions for AI agents, but
nothing measures how far that surface actually gets one. Every claim about it
today is an assertion with no run behind it.

`docs/design/agentic-office-workflow.md` writes down the proposal that answers
it by measurement: a `wafflebase-office-agent` (system prompt + command scope;
the model loop is borrowed from the host, not built) and a
`wafflebase-office-bench` (27 hand-written tasks, 5 personas, 7 document types,
graded by document-state inspection with **no LLM in the loop**). Two score
branches keep the agent's skill and the CLI's coverage separable. The claim
under test is that office specialization shifts the curve left — lowering the
minimum viable model class — rather than raising the ceiling.

The doc records that the bench has **zero runs** and a known isolation hole
(the subject's working root auto-loads `CLAUDE.md` and held 45 expected
answers), both as top rows of the risk table. No number from it is quotable
yet, and the doc says so rather than implying otherwise.

### Why the audit is re-measured rather than restated

The original A/B/C split does not survive contact with the routes actually
declared in `packages/backend/src/api/v1/*`:

| | Original | Measured | Evidence |
| --- | --- | --- | --- |
| Tab rearrange (3) | A | **B** | `tabs.controller.ts` has `@Get` · `@Post` · `@Patch(':tabId')` only — no DELETE |
| Comments (6) | A | **B** | No comment controller under `/api/v1`; comments live in the Yorkie CRDT |
| Sheet floating images (2) | A | **B** | Workspace image upload/read only; no per-worksheet endpoint |
| Slide editing (5) | A | **A′** | No granular slide ops — only whole-document `PUT content` |
| Folders + copy (7) | A | **A′** | `folder.controller.ts:22-23` is JWT-only and outside `/api/v1`; unreachable with an API key, reachable with a `wafflebase login` session (`jwt.strategy.ts:21`) |
| Board (2) | B | B (confirmed) | `docs-content.controller.ts:94` rejects any non-`doc`/`slides`/`note` type |
| Workspace images (3) | — | **A** (new) | `POST`/`DELETE .../images/:imageId`, `GET .../images/:imageId` — absent from the original list |

Net: "fill in the missing commands" covers roughly **37** items, not 57. That
changes what gets built first, which is the whole point of the audit.

## 2. `06d732285` — closing class A

38 backend capabilities the web editor exercises daily, with no way to call
them from a terminal. Nothing here is a new feature; every endpoint already
existed, which is why this closes in the CLI layer alone.

```
sheets styles / sheet-style / column-styles / row-styles
sheets column-widths / row-heights
sheets conditional-formats / data-validations
sheets filter / pivot
sheets freeze / hidden / merges
sheets charts
sheets clear / insert / delete / move
docs|slides|notes set-content
images upload / get / delete
```

All read their JSON body from stdin or `--data`, the way `sheets cells batch`
already does, and every one supports `--dry-run` and the global `--format`.

The four structure verbs mirror the backend's own `POST clear|insert|delete|move`
one-to-one. `delete` removes rows or columns; `clear` empties a cell range while
keeping the grid. `sheets delete` deliberately does **not** claim the bare
`delete` alias its siblings claim (`clear`, `insert`, `move` all do), because
that name reads as document deletion.

Every command carries a `schema/registry.ts` entry — a command in the commander
tree but not the registry is invisible to the agents this surface exists for.
Cross-checked by walking the built program against the registry:

```
commander leaves: 86
registry entries: 85
LEAVES MISSING FROM REGISTRY: [ 'schema' ]   # pre-existing on main
REGISTRY ORPHANS: (none)
ALIAS PROBLEMS: (none)
```

Classes A′, B and C are untouched — folders and copy need an auth decision,
comments and board need backend endpoints first. The class-A table in the
design doc is left **as measured**, with a status note under it: it is the
*before* half of the comparison the bench exists to make.

## How this was built, and what needed correcting

The nine capability groups were implemented by parallel agents behind a
sequential seam for the three shared files (`http-client.ts`,
`schema/registry.ts`, `sheets.ts`), since concurrent edits there would have
overwritten each other. Two things came out of review rather than out of the
agents:

- **Test files were in the wrong place.** The orchestration prompt mandated
  `src/commands/*.spec.ts`; the package's actual convention is
  `test/*.test.ts`, with zero `src` specs on `main`. That mismatch had been
  papered over by widening `vitest.config.ts`. The nine suites were moved to
  the house location and the config reverted. Relocating surfaced a real
  defect: the `vi.mock('../client/http-client.js')` paths had not been
  rewritten, and an unresolved mock path silently exercises the real module —
  88 tests were asserting against a client that was never faked.
- **`delete-range` was renamed to `delete`.** The original name inverted what
  the command does (`clear` is what empties a range) and broke the
  insert/delete/move trio's symmetry with the backend.

## Test plan

- [x] `pnpm verify:fast` — exit 0, run independently after each change, not
      taken on an agent's word
- [x] `pnpm --filter @wafflebase/cli test` — **51 files / 958 tests** passing
      (up from 42 / 771 on `main`; +187 tests)
- [x] `pnpm --filter @wafflebase/cli typecheck` — exit 0
- [x] Commander tree ↔ schema registry cross-check (output above)
- [ ] Not exercised against a live backend. Every command is covered by
      request-shape tests with a faked HTTP layer; none has been run against a
      real server.
- [ ] Reviewer check: the Step 3 ordering in the todo file is judgement, not
      bench evidence — the bench has never run.

### Known, not addressed

`pnpm --filter @wafflebase/cli lint:check` exits 2 (`packages/cli/src` matches
no ESLint config) and `prettier --check` flags three files. Both reproduce on
an untouched tree and are pre-existing.

## Summary by CodeRabbit

- **New Features**
  - Added CLI commands to replace document, presentation, and note content.
  - Added workspace image upload, download, and deletion commands.
  - Expanded spreadsheet tools for styles, dimensions, views, rules, charts, filters, pivots, and row/column operations.
  - Added dry-run previews, JSON input via stdin or `--data`, validation, and formatted output across new commands.

- **Documentation**
  - Added office-agent workflow benchmark, evaluation guidance, capability audits, and task tracking.
  - Updated CLI command reference and examples.


hackerwins:
### Code review

Found 1 issue:

1. Five `set` commands do not unwrap the envelope their own `get` prints, so the documented `get | set` round-trip sends a doubly-wrapped body and the server rejects it. `getMerges` returns `{ merges: … }` and `output(res.data, fmt)` prints that envelope, but `merges set` passes the parsed JSON straight to `setMerges`, which wraps it again — `{"merges":{"merges":{…}}}`. Same shape for `filter`, `pivot`, `column-widths` and `row-heights`. `docs/design/cli.md` states the rule for this whole get/set group: "Each `set` also accepts the envelope its own `get` prints, so `sheets styles get D | sheets styles set D` round-trips."

`merges set` takes the value as-is:

https://github.com/wafflebase/wafflebase/blob/06d732285910f0abddfa754cf8b70a4a02180fbc/packages/cli/src/commands/sheets-view.ts#L274-L279

`filter set` and `pivot set`, same:

https://github.com/wafflebase/wafflebase/blob/06d732285910f0abddfa754cf8b70a4a02180fbc/packages/cli/src/commands/sheets-analysis.ts#L133-L138

https://github.com/wafflebase/wafflebase/blob/06d732285910f0abddfa754cf8b70a4a02180fbc/packages/cli/src/commands/sheets-analysis.ts#L209-L214

`column-widths` / `row-heights` go through `readSizeMap`, a bare `JSON.parse` with no unwrap:

https://github.com/wafflebase/wafflebase/blob/06d732285910f0abddfa754cf8b70a4a02180fbc/packages/cli/src/commands/sheets-dimensions.ts#L42-L46

The other seven resources in the same PR do handle it — `sheets-styles.ts` has a shared `unwrap()` and `sheets-rules.ts` an `extractRules()`, both with tests named for the round-trip:

https://github.com/wafflebase/wafflebase/blob/06d732285910f0abddfa754cf8b70a4a02180fbc/packages/cli/src/commands/sheets-styles.ts#L73-L78

The gap shipped untested because no suite pipes a real `get` stdout into `set` — the `set` tests all feed the already-bare shape via `--data`. Fix: reuse `unwrap()` in the five handlers and add the round-trip regression test the styles suite already has.

Checked and rejected as false positives: the `get` handlers' `--dry-run` block sitting outside `try` (matches `cells.ts` on `main`, and `runCli` envelopes it anyway at cli.ts#L101), `--data ''` read as absent (same `if (dataStr)` convention as `cells.ts` on `main`), and missing `packages/cli/skills/` entries (no guidance file requires them, and every new command is in `schema/registry.ts`).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1000, opened 2026-08-31
Add the template gallery: publish a document, start one from it

hackerwins:
## Summary

Adds the template gallery — publish a document as a template, and start a new
document from one. Design: [`docs/design/template-gallery.md`](https://github.com/wafflebase/wafflebase/blob/feat/template-gallery/docs/design/template-gallery.md).

The design starts from how Canva and CapCut actually share documents. Neither
invents a file format: a template is an ordinary artifact plus a listing, using
one copies it into your own account, and the original is never touched. They
differ only in parameterization — CapCut marks replaceable clips because its
template must be usable in thirty seconds on a phone. An office document is
opened in its editor either way, so we follow Canva and copy the whole thing;
`{{placeholder}}` slots are recorded as **declined**, not deferred.

That reading is why the diff is small. `DocumentCopyService` already duplicates
all eight document types with rollback; its only template-relevant limit was a
destination hardcoded to the source's workspace. So "use this template" is that
service with a `dest`, and preview is a `viewer` share link minted at publish,
which the `/shared/:token` viewers and the Yorkie auth webhook already
understand.

A template is deliberately **not** a `Document.type`: that column routes which
editor opens a document, and a template must open in the editor of whatever it
is a template of. The listing is a sidecar row with a unique `documentId`, so
publishing is an upsert.

### Authorization

Two rules carry it:

- **Publishing is manager-gated at every tier**, workspace included. It hands
  the content to an audience workspace membership no longer bounds — the same
  escalation an editor share link is, so it takes the same bar.
- **Using inverts.** Read authority comes from the listing's visibility; write
  authority from membership of the *destination* workspace. This is the only
  path where a document crosses a workspace boundary.

Two gaps found in self-review and fixed here, both with regression tests:

- Read authority follows `document.workspaceId`, not the listing's denormalized
  copy — otherwise moving a document to another workspace leaves its old
  workspace reading it through the listing.
- `GET /documents/:id/template` is membership-gated rather than
  visibility-gated — otherwise anyone holding an *expiring* viewer share link
  (which reveals the document id) could trade up to the listing's
  *non-expiring* preview token.

### Scope

Ships **Phase 1** — unlisted template links. `public` is refused with a `400`
until Phase 3 exists rather than silently downgraded, because a publisher told
"ok" would believe their template was in a gallery that nothing reviews.

There is **no discovery surface yet**: a template is reachable only by someone
sending you its URL. Phase 2 (the collection endpoint, thumbnails, taxonomy,
the workspace Templates tab) and Phase 3 (review, frozen-copy promotion, image
re-hosting, public browse, licensing, takedown) are planned in full in the
design doc's scope map and in
[`docs/tasks/active/20260901-template-gallery-todo.md`](https://github.com/wafflebase/wafflebase/blob/feat/template-gallery/docs/tasks/active/20260901-template-gallery-todo.md).

Three Phase 1 shortcuts, each recorded in the design doc with its reason:
preview opens `/shared/<token>` in a new tab rather than embedding
(`SharedDocument` is 1088 lines keyed on `useParams`, and a preview pane is not
worth risking the live share route); a logged-out visitor goes to `/login`
without returning to `/t/:id` (needs the OAuth `state` to carry a
server-validated return path); and thumbnail capture is not wired (the column,
serving path and card rendering are).

## Test plan

- `pnpm verify:fast` green; `verify:self` green on push.
- 30 new `TemplateService` cases: the publish gate (author / owner / plain
  member / non-member), every visibility tier including anonymous and
  cross-workspace, `404`-not-`403` on tiers the caller may not see, the
  membership gate on the document-scoped lookup, the document-moved case,
  best-effort `useCount`, and unpublish revoking the preview link.
- 4 new `DocumentCopyService` cases for the destination parameter, including
  that a template use lands in the destination *root* rather than inheriting
  the source's folder id from another workspace, and that the existing
  "Make a copy" behavior is unchanged when no destination is passed.
- `uniqueTitle` unit tests.
- Manual smoke pending in `pnpm dev`: publish → copy link → open `/t/:id`
  signed out → preview → sign in → use → land in the new document.

## Summary by CodeRabbit

- **New Features**
  - Added template publishing, copyable links, previews, metadata, usage counts, and unpublishing from the Share dialog.
  - Added a public template landing page and support for creating documents in a selected workspace or folder.
  - Added unlisted, workspace, and public visibility options; public publishing remains subject to review.
- **Bug Fixes**
  - Prevented duplicate titles when creating documents from templates.
  - Improved handling of expired sessions and clipboard-copy failures.
- **Documentation**
  - Added template gallery design, task tracking, and API documentation.


hackerwins:
### Code review

Reviewed across five lenses (CLAUDE.md compliance, shallow bug scan, git history of the modified files, prior PR comments, and code-comment adherence). Four findings survived verification. All four are addressed in 6c84da6d6; two were real defects in code this PR adds.

**1. `publish()` blanked unmentioned fields on a partial re-publish, which could widen visibility** (fixed)

`publish()` is an upsert on a route documented as "publish **or re-publish**", but every optional field fell back to a first-publish default rather than the existing row. A body of `{ thumbnailId }` reset `visibility` to `unlisted` — the *more* permissive tier, since holding the id is the whole capability — silently widening a workspace-scoped listing. `update()` had the correct pattern on the adjacent lines.

https://github.com/wafflebase/wafflebase/blob/515ca4161d0032b37e5efcffd8d4a48600296824/packages/backend/src/template/template.service.ts#L113-L140

**2. `unpublish()` could strand a permanent anonymous read capability** (fixed)

The listing was deleted first and a failed share-link revoke was caught and swallowed, still returning `{ deleted: true }`. The preview link is non-expiring and nothing sweeps it, so a transient failure left the document anonymously readable with no listing remaining to surface or revoke it — the exact outcome the method's own docstring says it prevents. Revoking first and letting failure throw destroys nothing.

https://github.com/wafflebase/wafflebase/blob/515ca4161d0032b37e5efcffd8d4a48600296824/packages/backend/src/template/template.service.ts#L190-L213

**3. Risks table stated Phase 3 mitigations in the present tense** (fixed)

"A publisher's own use does not increment it, and `use` is throttled per user" describes 3d, not shipped behavior. Only the global per-IP throttle applies today.

https://github.com/wafflebase/wafflebase/blob/515ca4161d0032b37e5efcffd8d4a48600296824/docs/design/template-gallery.md#L406-L410

**4. Task status header contradicted the branch** (fixed)

Said "implementation not started, not approved" while the same branch ships Phase 1.

---

**One limitation recorded rather than fixed.** A `datasource` / `lakehouse` tab carries a workspace-scoped `TabMeta.datasourceId`; cross-workspace copy is new in this PR, so such a tab now lands inert in the destination. Verified this is **not** an access bypass — every datasource route re-derives authorization from the row's own `workspaceId`. Refusing to publish a document holding such a tab is filed as a Phase 2 item, since a template depending on a private connection is not shareable and the check costs a Yorkie read at publish time.

**One finding rejected.** `docs/tasks/README.md` landed two commits after the todo file rather than alongside it (CLAUDE.md "Commit Together"). The branch is consistent as merged and rewriting pushed history to satisfy the letter of it is not worth it.

hackerwins:
Thanks — went through all 7. Six addressed in 106cd69c8, one rejected.

**The most valuable finding was the lint one, and not for the reason it was filed.** `verify:fast` runs `pnpm backend test` but never `pnpm backend lint` — the only backend ESLint in the chain is `arch:backend`, which uses the separate `eslint.arch.config.mjs`. So all 39 errors passed both `verify:fast` and `verify:self`. Compounding it, `packages/backend`'s `lint` script runs with `--fix`, so invoking it rewrites the tree rather than reporting on it.

The new `src/template/` files are now clean under a check-mode run. The 20 errors in `document-copy.service.spec.ts` and `document.module.ts` are **left alone** — confirmed present on `main` by stashing and re-linting, so they are pre-existing debt from this same gap, and fixing them here would bury the diff. Recorded in the task's lessons file.

Addressed:

- **Task status** — fixed in 6c84da6d6 (an earlier reviewer caught it too).
- **`prettier` / `require-await` / `no-unsafe-assignment`** — fixed. The `no-unsafe-assignment` cases came from nesting `expect.objectContaining` (typed `any`) inside an object literal; rewrote those three assertions to read the captured call argument with `toMatchObject`, which is lint-clean and reads better.
- **Tests for `TemplateService.update`** — added 7 cases: manager authorization, 404, that an omitted field means "leave alone" while `''` means "clear" (the distinction `publish` got wrong and that 6c84da6d6 fixes), and that `public` is refused on an edit as well as on a publish.
- **`isAuthExpiredError` in `template-landing.tsx`** — correct, and it was already handled in `template-share-section.tsx`, so this was an inconsistency in my own code. Fixed.
- **Clipboard errors conflated with the mutation** — correct and the worse of the two: a denied clipboard permission after a successful publish reported "Failed to publish template" while the listing was already live. Clipboard writes now report separately, and `handleCopy` no longer rejects without feedback.
- **Security/privacy finding on `template.service.ts`** — this is the `unpublish()` ordering issue, already fixed in 6c84da6d6: the share link is now revoked *before* the listing is deleted and a failure throws, because the link is non-expiring and nothing sweeps it.

Rejected:

- **`tags TEXT[]` in the migration** — the comment body contains the analysis scripts but no stated claim, so there is nothing to act on. The column is written on every path (`publish` always supplies `tags`) and Prisma reads a scalar list without a default as `[]`, so a missing `DEFAULT '{}'` would only matter to a raw INSERT that bypasses Prisma. Happy to reconsider with a concrete failure case.

PR #1001, opened 2026-09-01
Make templates browsable: the collection endpoint and its surfaces

hackerwins:
## Summary

Phase 2 of the template gallery ([design](https://github.com/wafflebase/wafflebase/blob/feat/template-gallery-browse/docs/design/template-gallery.md)). #1000 shipped one template, one link, one copy; nothing could **find** a listing you had not been sent. This builds the store.

### `GET /templates` — the collection endpoint

One endpoint behind the workspace Templates tab, the New-from-template picker, and later the public gallery. Three properties are load-bearing and are enforced in the service, not in any caller:

- **Visibility is a query constraint, never a post-filter** — a row the caller may not see is never fetched.
- **No scope selects `unlisted` listings.** Holding the id is that tier's entire access story; a collection would hand it out wholesale.
- **A card carries no `previewToken`.** A page of 24 would otherwise be 24 non-expiring read capabilities handed over to render a grid. It is a distinct type (`TemplateCardView`), not an optional field, so a response cannot grow one back by accident.

Paging is keyset on `(sort, id)`. A gallery ranked by a counter that changes under the reader would skip and repeat rows under offset paging; the `id` tiebreak is what makes a boundary inside a run of equal `useCount` unambiguous.

### Taxonomy

`category` becomes a closed list shared with the frontend — a facet is only useful if the values are closed, and a per-workspace table would let each workspace invent its own spelling of "Finance" for the public gallery to merge later. Tags stay freeform but are normalized on write, so `Budget`, `budget ` and `budget` stop being three facets.

### Surfaces

Workspace **Templates** tab at `/w/:workspaceId/templates`, and **New from template** in the documents create menu (workspace-scoped only — a use needs a destination workspace, and the all-workspaces list has no single one to offer). Cards render a document-type icon; thumbnail capture is Phase 2b.

### Both Phase 1 shortcuts cleared

A gallery makes both far more visible than a hand-sent link did.

**Preview now embeds the real read-only viewer** instead of opening a second tab. The design doc justified the shortcut by calling `SharedDocument` a 1088-line component keyed on `useParams`; **that was wrong**, and the doc now says so. Every component below the route already took `token` as a prop, so the split into `SharedDocumentByToken` was a function signature, not a refactor.

**A logged-out visitor returns to `/t/:id` after signing in.** The path rides in its own short-lived cookie rather than inside the OAuth `state`, which must stay an opaque token compared by equality. The cookie is **not** the boundary — the endpoint that sets it is unauthenticated, so its value is attacker-chosen by construction. `safeReturnPath()` is, and it runs **again on read** rather than trusting what was stored. It rejects rather than sanitizes: absolute URLs, schemes smuggled into the first segment (`/javascript:…`), protocol-relative values in both spellings the URL parser accepts (`//host` and `/\host`), and control characters — which is what closes `/\n/evil.example`, where the parser strips the newline and is then looking at `//evil.example`.

### Incidental

`TYPE_META` moves out of `document-list.tsx` into its own module, as its own comment always intended, so the gallery cards and the documents list define a type's appearance once. Typing the request in `GitHubAuthGuard` removed 4 pre-existing `no-unsafe-member-access` errors rather than adding a 5th.

## Test plan

- `pnpm verify:fast` green; `verify:self` green on push.
- **Backend lint checked explicitly in check mode** (`npx eslint` without `--fix`) — it is in no verify lane, which is how 39 errors slipped through on #1000. New files clean; pre-existing debt in `src/auth/` left alone and confirmed pre-existing by stash-comparing against `main`.
- 13 new `browse()` cases covering every property above: token omission, cross-tier leakage, the anonymous and non-member refusals, facet composition (both `type` and `workspaceId` constrain the same document relation), tag normalization on the facet, both sort orders, and cursor behavior including the lookahead row.
- 7 `normalizeTags` / taxonomy cases.
- 15 `safeReturnPath` / `loginRedirectUrl` cases, most of them open-redirect attempts.
- Manual smoke pending: publish with Workspace visibility → Templates tab → card → embedded preview → sign out → `/t/:id` → sign in → land back on the template → Use.

## Summary by CodeRabbit

* **New Features**
  * Added searchable, filterable template galleries with categories, tags, sorting, pagination, and public or workspace scopes.
  * Added a Templates tab for workspaces and “New from template” document creation.
  * Added in-page template previews and return-to-page handling after GitHub sign-in.
  * Added controls for editing template visibility, category, and tags.
* **Improvements**
  * Standardized document-type labels, icons, and colors throughout document lists.
  * Added validation and normalization for template categories and tags.
  * Improved template browsing with clearer loading, error, and empty states.


hackerwins:
### Code review

Self review (two independent passes: adversarial bug hunt, and doc/comment-vs-code plus repo-convention consistency) followed by the CodeRabbit pass. Eight findings total; all real ones are fixed in 0f0dfad68 and 4b9bed003.

**Found by actually opening the page — the three that mattered most:**

1. **The Templates page had no way in.** Route registered, but no sidebar entry and no `ROUTE_TITLES` entry, so it existed only for someone who typed the URL.
2. **Browsing failed outright** with "Failed to load templates". Every workspace-scoped URL carries the *slug* (`/w/hackerwins-s-workspace/templates`) and the pages read it off `useParams`, but `BrowseTemplatesDto` required a UUID — so the only value the gallery ever sends was the one it refused. `UseTemplateDto` had the same bug on the destination, which the picker would have hit next. `assertMember` resolves either form, so the DTOs were simply wrong to narrow it. CodeRabbit found this independently and proposed resolving on the frontend; fixed at the DTO instead, since the resolution already exists server-side and every other workspace-scoped route relies on it.
3. **Publishing into the gallery was a two-step nobody would find** — "Publish as template" made an unlisted listing and the visibility control only appeared afterwards, so the gallery stayed empty unless the publisher noticed a second dropdown and pressed Save again. Visibility is now chosen before publishing, defaulting to workspace.

**From the review passes:**

4. `typeMeta()` looked up a wire string in an object literal, so `documentType: "toString"` returned an inherited function rather than the fallback — `Icon` undefined, React throws on render. Own-property check now, the same trap [`login-form.tsx` already documents](https://github.com/wafflebase/wafflebase/blob/4b9bed003/packages/frontend/src/components/login-form.tsx#L15-L24) for its `?error=` lookup.
5. A whitespace-only tag filter passed `@Length`, normalized to nothing, and became `has: undefined`, which Prisma drops — `?tag=%20%20` silently returned the whole unfiltered gallery. Refused at the DTO, and the service no longer emits an undefined filter either.
6. **`login-return-path.spec.ts` was a binary blob.** It contained *raw* NUL and SOH bytes where it meant `\x00` and `\x01`, so Git classified the file as binary and it rendered as `Bin 0 -> 3369 bytes` — unreviewable in this PR, with no line-level history afterwards. Jest never noticed; those bytes are legal in a string literal. Now 91 lines of readable text.
7. The Share dialog gated its unpublish button on `listing.canManage` but rendered the visibility/category/tags editor unconditionally, so a plain member got controls whose Save could only ever 403. Same gate now.
8. `login-return-path.ts` called `isSecureCookie()` rather than `oauth-state.ts`'s documented external entry point `useSecureCookies()`, which every other consumer uses. Fixed, and the export it had forced is reverted — `oauth-state.ts` is byte-identical to `main` again.

**Held up under adversarial review, for the record.** `safeReturnPath` was attacked with protocol-relative forms in both spellings, scheme smuggling, control characters (including `\n`-disguised `//`), unicode whitespace, percent-encoded separators, `@`-in-authority, and dot-segment confusion. No bypass found — and the redirect is built by concatenation onto a fixed origin rather than `new URL(path, base)`, so even inputs that produce a double slash stay on the correct host. `browse()`'s visibility scoping, the shared `documentWhere` clause, and the keyset pagination were verified against the generated Prisma types.

PR #1033, opened 2026-09-06
Keep a page number from absorbing the text typed next to it

### LINKED ISSUE #871: Typing after a page number in a header produces nothing visible
### What happened:

In a header or footer, once **Page number** has been inserted, typing directly
after it does nothing on screen. The caret advances as if the characters went
in, but nothing appears. Typing in a header *before* inserting a page number
works normally.

Typing next to a page number inserts ordinary text next to it.

1. Open a docs document and **double-click the top page margin** to enter header
   editing.
2. Click **# Page number** in the header toolbar.
3. Type any characters without moving the caret.

→ The caret advances; nothing is drawn. 

https://github.com/user-attachments/assets/52ad5f95-ef2e-4a55-8510-3c7af2c380b5

### Environment:

- Operating system: macOS
- Browser and version: Whale 4.39.410.14 (arm64)

shjohw12:
## Summary

Typing right after a page number in a header did nothing on screen — the caret
advanced as if the characters went in, but nothing appeared.

A page number is a `#` placeholder carrying `pageNumber`, and the renderer
replaces such a run **whole** with the page's number. Text merged into that run
can therefore never be drawn, while the offsets keep counting it. The rule that
structural inlines do not absorb typed text already existed; both insertion
paths spelled it as `style.image` alone. They now use `isStructuralInline`, the
predicate the model already defines for that category.

## Why

**Both paths, not one.** An edit is carried twice: `applyInsertText` updates the
cache the screen renders from, and `YorkieDocStore.insertText` edits the Yorkie
tree that is stored, synced to other editors and read back on reload. Fixing the
cache alone would look correct until the next reload rebuilt it from the tree —
a harder bug than the one being fixed. A test asserts the two describe the same
runs, so reverting either half fails it.

`isStructuralInline` is exported from both package entries: the Node entry is
what the frontend's `.integration.ts` lane resolves, and `entry-parity` guards
that pairing.

## Linked Issues

Fixes #871

## Author checklist

## Verification

Skip reason (if applicable): none — left unticked until CI reports. `pnpm verify:fast`
is green locally on `upstream/main` @ `e9dc0a359`.

7 new tests: 4 on `applyInsertText` (cache), 3 driving `YorkieDocStore` against a
real Yorkie tree. Each half was reverted in turn to confirm they fail:

```
(a) block-helpers reverted   → 3 docs tests + the cache/tree divergence test
(b) yorkie-doc-store reverted → 2 frontend tests
    AssertionError: page number and typed text are separate runs: expected 1 to be 2
    AssertionError: cache and tree must describe the same runs: expected [ '#abc' ] to deeply equal [ '#', 'abc' ]
```

## Risk Assessment

- **User-facing risk:** low. Only the decision to split a run changes. Documents
  that already absorbed text are not repaired — new input is correct from here.
- **Rollback plan:** single commit, revert it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Text inserted next to page numbers or images now remains in a separate, regular text segment.
  - Structural inline elements are no longer split during text insertion.
  - Inserted text no longer inherits page-number or image formatting.
  - Behavior is consistent across document and header editing paths.

- **Documentation**
  - Added guidance explaining page-number inline behavior and text insertion rules.

- **Developer Experience**
  - Exposed the structural-inline helper through standard and Node-compatible package exports.

### DIFF
--- a/docs/design/docs/docs-header-footer.md
+++ b/docs/design/docs/docs-header-footer.md
@@ -75,12 +75,18 @@ alignment) apply normally.
 Because the *whole run* is replaced, a page-number inline must describe exactly
 one page number and nothing else. That makes it a **structural inline**, the
 same category as an image (`isStructuralInline` in `model/types.ts`), and it
-carries that category's insertion rule: **typing next to one splits the run**,
-so the typed text lands in its own inline without the structural style rather
-than being absorbed into it. Absorbed text is text the renderer can never draw
-— the run is replaced by the page's number regardless of how many characters
-it holds — while the offsets keep counting it, so the caret advances over
-characters that are absent from the screen.
+carries that category's insertion rule: **typed text lands beside the run,
+never inside it** — in its own inline, without the structural style. Absorbed
+text is text the renderer can never draw — the run is replaced by the page's
+number regardless of how many characters it holds — while the offsets keep
+counting it, so the caret advances over characters that are absent from the
+screen.
+
+Beside, and the run itself is never cut in two: two runs carrying `pageNumber`
+paint the page number twice. A well-formed one holds a single `"#"` and there
+is nothing to cut, but a document edited before this rule existed holds runs
+that absorbed text, and those are exactly the documents the rule has to be
+right about. The typed text goes wholly before or after such a run.
 
 The rule is enforced twice, because an edit is carried through two separate
 code paths: `applyInsertText` (`store/block-helpers.ts`) updates the cache the
--- a/docs/design/docs/docs-header-footer.md
+++ b/docs/design/docs/docs-header-footer.md
@@ -72,6 +72,22 @@ During rendering, the text is replaced with the current page number string
 (`"1"`, `"2"`, ...). All other style properties (bold, italic, fontSize, color,
 alignment) apply normally.
 
+Because the *whole run* is replaced, a page-number inline must describe exactly
+one page number and nothing else. That makes it a **structural inline**, the
+same category as an image (`isStructuralInline` in `model/types.ts`), and it
+carries that category's insertion rule: **typing next to one splits the run**,
+so the typed text lands in its own inline without the structural style rather
+than being absorbed into it. Absorbed text is text the renderer can never draw
+— the run is replaced by the page's number regardless of how many characters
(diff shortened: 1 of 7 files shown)


hackerwins:
Reviewed the full diff. The two-path argument holds — fixing only `applyInsertText` would have looked right until the next reload — and `isStructuralInline` is the correct predicate to spell it with. I pushed one follow-up commit rather than blocking on it.

**The cache cut the structural run in two.** `applyInsertText` split at the caret and gave `{...inline.style}` to *both* halves, so a run that had already absorbed text — which is every header edited before this fix, the population the fix exists for — came back as two runs carrying `pageNumber`. The renderer replaces each of them whole with the page number, so one page number becomes two. It also disagreed with the tree path, which inserts the new node before or after the whole run and never through it:

```
applyInsertText({ text: "#abc", pageNumber }, 2, "X")
  cache: ["#a"(pn), "X", "bc"(pn)]     tree: ["#abc"(pn), "X"]
```

The branch now splices the typed text in beside the run instead of splitting it. For a well-formed one-character run the two spellings are identical, so nothing about the shipped behaviour changes; the difference is only on the damaged runs. Test added, and it fails on the previous commit with exactly the divergence above.

**On the two CodeRabbit findings:**

- *Header/footer cases* — taken. `insertPageNumber` refuses to run outside a header or footer, so the one region a page number can exist in was the one region no test covered. Added a header case asserting the tree and the cache agree.
- *Empty structural inlines* — real, but not this branch. `normalizeInlines` drops **every** empty inline (`block-helpers.ts:92`), so the cache cannot hold an empty page-number run after any edit path, not just this one, while `treeNodeToBlock` filters only empty *image* inlines. That asymmetry predates the PR and is invisible (an empty run paints no glyphs). Closing it means changing one of those two filters, which is its own change with its own blast radius.

Not fixed, noted for later: `getSplitPointStyle` (`block-helpers.ts:227`) drops `image` but not `pageNumber`, so pressing Enter at a page number gives the new block an empty inline still carrying the flag. Same family, different verb, pre-existing.

hackerwins (approved):
Approving. The two-path fix is right and the added commit closes the one case where the cache and the tree still disagreed. Merging once the heavy lanes report.

PR #1061, opened 2026-09-13
Let any workspace member mint their own API keys

hackerwins:
## Summary

Minting an API key was gated on `assertOwner`, and the frontend hid the whole API Keys section behind `isOwner`. A member could neither create a key nor see that any existed, which shut every non-owner out of the CLI and the v1 REST API.

The gate had no security work to do. `ApiKeyStrategy` builds the request identity as `id: apiKey.createdBy`, so a key carries its creator's authority rather than the workspace's; `WorkspaceScopeGuard` re-checks that membership on every request; and `/api/v1` exposes no workspace administration for a key to reach. A member's key is that member's own hands, reached over HTTP — the ownership check withheld a credential for authority they already had.

Visibility and revocation are scoped by creator instead:

| Caller | `POST` | `GET` returns | `DELETE` accepts |
| --- | --- | --- | --- |
| member | ✅ | keys they minted | keys they minted |
| owner | ✅ | every key in the workspace | any key in the workspace |

A member naming somebody else's key gets **404, not 403** — whether the workspace holds another member's integration is itself workspace information, and a refusal has no need to confirm it.

### Notes on the implementation

- The controller reads the role off the member row `assertMember` already returns, so the scope costs no extra query.
- `revoke` moves to `updateMany` because narrowing by `createdBy` does not fit `update`'s unique-only `where`; the row count is then the whole ownership answer. The revoked row is read back through an **explicit selection** — `update`'s default return had been handing back `hashedKey`.
- No client-side gating on the revoke button: the server only ever returns keys the caller may revoke, so a second copy of the rule would only drift.
- The `Created by` column renders for an owner only; a member's rows are all their own.
- No CLI change — it calls the same routes, so it opens to members for free.

### Doc consistency

Five sites still said minting is owner-only and inferred from it that a key's holder is an owner: `docs/design/backend.md`, `docs/design/cli.md`, `docs/design/agentic-office-workflow.md`, and comments in `api/v1/documents.controller.ts` and `api/v1/folders.controller.ts`. The gating logic never relied on that (it resolves the role per request), but left alone that text is what would make the next reader think this contradicts the commits that hardened the API-key membership invariant. All updated.

## Test plan

- `pnpm verify:fast` — green.
- `packages/backend/src/api-key/api-key.controller.spec.ts` (new): member create allowed, non-member refused, member scoped to own keys on list and revoke, owner unscoped on both.
- `api-key.service.spec.ts`: `createdBy` filtering on list and revoke, 404 for a key outside the scope, revoked row returned without its hash.
- `packages/backend/test/api-key-http.e2e-spec.ts` rewritten and **run against a live Postgres** (5/5 pass): member mints (201); member's listing is their own; owner's is everyone's; member revoking the owner's key → 404; both revoke their own; non-member refused on create and list.
- Local `pnpm dev` smoke of the settings page still to be done by the author.

Authorization was reviewed as a privilege boundary: every `/api/v1` controller mounts `WorkspaceScopeGuard`, manager-gated routes re-derive `member.role` live, `scopes` is `@IsIn(['read','write'])` under a `forbidNonWhitelisted` pipe, and key management itself is JWT-only, so a `Bearer wfb_` token cannot mint or revoke keys. There is no role-demotion endpoint — a role is fixed at invite acceptance, and `removeMember` revokes the departing user's keys in the same transaction.

## Authorship

**This PR was written autonomously by Claude Code** in the session linked below; a human has not reviewed it line by line. What is *not* verified: the local `pnpm dev` smoke of the settings page noted in the test plan above. Everything else claimed here is covered by the tests in this diff, by `pnpm verify:fast`, or by the live-Postgres e2e run.

https://claude.ai/code/session_01Y3afwXgsdgyJCt9ezLdA5R

## Summary by CodeRabbit

- **New Features**
  - Workspace members can create, view, and revoke their own API keys.
  - Workspace owners can manage all workspace API keys.
  - API key management is now available to all members in workspace settings.
  - Owners can see who created each key.
  - API keys follow the creator’s current membership permissions for document and folder actions.

- **Bug Fixes**
  - Unauthorized access to another member’s keys is prevented without revealing whether they exist.

- **Documentation**
  - Updated API key permissions and management guidance across product documentation.


hackerwins:
@claude review this.

hackerwins:
@claude rerun

hackerwins:
@claude loop

PR #1073, opened 2026-09-17
Sheets: link every URL inside a cell, not just whole-cell URLs

hackerwins:
## Summary

Sheets rendered a cell as a hyperlink only when the cell's **entire** value was
a bare URL. The detector was anchored and rejected interior whitespace, so the
shapes people actually type matched nothing.

Measured on a live sprint-planning sheet, one screen of a column held **9 URLs
and 2 of them were links**:

| Cell | URLs | Linked | Why not |
| ---- | ---- | ------ | ------- |
| H4   | 2    | 0      | a sentence precedes the URLs |
| H10  | 2    | 0      | `- 릴리즈노트:` / `- PR:` labels |
| H26  | 3    | 0      | `시제품 :` / `기획 필요사항 :` labels |
| H15  | 1    | 1      | value is the URL and nothing else |
| H17  | 1    | 1      | same |

The linked cells were the exception, not the rule. Discoverability compounded
it: the only way to open one was Ctrl/Cmd+click, the cursor never changed, and
the shortcut was absent from the help sheet — on a document shared **view-only**,
where a reader has no reason to guess a modifier.

This widens detection from *the whole value* to **every URL span inside it**,
and stays render-time: no `Cell` field, no CRDT change, no Yorkie schema change,
no xlsx path touched.

Design: [`docs/design/sheets/sheet-hyperlink.md`](docs/design/sheets/sheet-hyperlink.md) ·
Task: [`docs/tasks/active/20260917-sheets-cell-link-spans-todo.md`](docs/tasks/active/20260917-sheets-cell-link-spans-todo.md)

### Why not a real link model

| | Two links in one cell | Display text ≠ URL | xlsx round-trip | Cost |
| --- | --- | --- | --- | --- |
| `Cell.link?: string` (Excel) | ✗ | ✓ | ✓ | low |
| rich-text runs (Google `textFormatRuns`) | ✓ | ✓ | ✗ lossy | high |
| **render-time spans (this PR)** | ✓ | ✗ | n/a | low |

A single field cannot express H10/H26, which *is* the complaint. Google's
per-run model can, but Sheets has no inline-run model to hang it on and the
10 MB Yorkie ceiling counts CRDT nodes — and it would not buy xlsx fidelity
anyway, since SpreadsheetML runs carry no link property at all. The two
approaches compose rather than compete: when a persisted link lands, span
detection stays as the fallback, which is exactly how Google layers
auto-linkification under explicit links.

### Geometry comes from the paint

`GridCanvas.render()` repaints the whole canvas every frame, so the painter
records each span's screen box as it draws and hover/click read that back. Hit
geometry is therefore *the arithmetic that produced the pixels* rather than a
reconstruction of it — "the underline is here but the click lands there" is not
a reachable state. It also removed the async `getCell` from the click path.

### What also changed

- Pointer cursor over a span; hover card listing the cell's links (Open / Copy).
- Ctrl/Cmd+click narrowed from the cell to the **span** under the pointer.
- A read-only (share-link) viewer opens a link with a **plain click**.
- `mailto:`, `www.` and bare addresses; formula results are linkified too.
- `toLineStartX` extracted to `layout.ts`, deduping alignment math that was
  copy-pasted three times in `gridcanvas.ts`.
- `url-detect.ts` deleted, superseded by `cell-links.ts`.

### Two blockers caught in review, and why they are worth reading

**The scan has to be linear, and twice it wasn't.** Detection runs per line,
per visible cell, per **frame**, so its cost is a correctness property of the
render loop, not a micro-optimisation.

- The first implementation used one regex alternation whose address branch
  began `[A-Za-z0-9._%+-]+@`. That backtracks quadratically: a 32 k cell took
  **~2.1 s per pass**. Replaced with an index scanner.
- The scanner then had the *same* defect by its own route — a scheme match
  consumes URL characters forward, and when the result does not parse the scan
  resumes one character later and re-reads the run. `'https://['.repeat(4000)`
  measured **429 ms**. It survived because all four adversarial fixtures had
  been written against the regex that was replaced.

Both walks are now bounded, and both **reject rather than truncate** at the
bound — truncating is not the conservative choice it looks like, because a
64-character suffix of a local part still parses as an address and a
2048-character prefix of a URL still parses as a URL. That was live:
`'x'.repeat(70) + '@b.com'` underlined from the middle of a word and opened an
address nobody typed.

**Hit boxes were bounded by the cell, which is not what bounds the pixels.**
With a freeze the grid is drawn as four separately clipped quadrants; without
one the headers are painted *over* the cells with no clip at all. So a link
scrolled behind a frozen pane stayed hoverable and clickable where nothing was
drawn — and the click guard does not cover the frozen pane, so in read-only a
plain click on a frozen cell could open a URL from a cell the reader cannot
see. The painter now records the region it clips to and every hit box is
intersected with it.

Also from review: the email branch had reintroduced the hazard the design
explicitly refuses for hostnames — `isHostname` accepts any two-letter alpha
TLD, and `.png` / `.sh` are both TLDs *and* file extensions, so `image@2x.png`,
`logo@3x.jpg` and `build@2.sh` became mail links. And the hover card never
appeared on merged or overflowing cells, because the pointer's cell and the
painter's cell disagree there — precisely the wide cells that hold several
links.

## Test plan

- `pnpm verify:fast` — green.
- `packages/sheets/src/view/__tests__/cell-links.test.ts` — 51 cases. Fixtures
  are the real H4/H10/H26 strings above, the named false positives
  (`build.sh`, `v0.2.3-rc.5`, `creators/26.09.1700`, `image@2x.png`), Korean
  particle adjacency, bracket balance, and the adversarial inputs under a
  **wall-clock assertion**, so a future pattern-based rewrite fails the suite
  rather than the user.
- `packages/sheets/test/view/gridcanvas-links.test.ts` — the clip threading
  (frozen quadrant, header strip, straddling the frozen edge) and `linkAt`.
  **Mutation-checked**: removing the `paintRegion` intersection fails 4 of its
  10 cases.
- **Browser, in the real engine** (standalone sheets playground, which mounts
  the real `GridCanvas`): labelled and multi-line URLs paint per span;
  `build.sh v0.2.3-rc.5 creators/26.09.1700` stays entirely plain; with two
  links in one cell each opens its own URL and the plain text between them
  opens nothing; `mailto:foo@bar.com` links including the prefix; the pointer
  turns only over a span, not over the gap between two spans in the same cell.
  Glyphs are painted once — no double-drawn bolding from the segment walk.

### Not verified in a browser

The **hover card** and the **read-only plain-click** were not exercised in the
full app: 5173 was held by another local checkout and CORS (pinned to
`FRONTEND_URL`) blocked the port mine fell back to. They are covered by types,
lint and unit tests only. The **frozen-pane clipping** likewise has no browser
run — the playground has no freeze UI — but it is covered by the mutation-
checked test above.

## Known limitations

- **Ctrl/Cmd+click no longer multi-selects a cell containing a URL.** The
  conflict predates this change, but its reach grew from "cells whose entire
  value is a URL" to "any cell with a URL substring" — including the
  `- PR: https://…` columns this targets. Resolving it means moving the editor
  gesture (Google uses `Alt+Enter`), which is a separate decision. **Happy to
  change it in this PR if you'd rather.**
- **Userinfo still reads as the host.** `https://good.com@evil.com` paints as
  written and opens `evil.com`. The hover card shows the real host, so the card
  is honest; the underline and the modifier-click are not. Pre-existing in kind.
- **In read-only, a cell whose span is clicked cannot be selected** — the click
  opens instead. Probably the right trade for a viewer, but it is a behaviour
  change.
- **The card is mouse-only**; there is no keyboard route to a link.
- **Nothing is persisted**, so a span stops being a link the moment the
  surrounding text changes, and nothing survives xlsx or CSV. That is the
  accepted cost of not touching the model; roadmap step 3 in the design doc
  pays it off (`HYPERLINK()`'s dropped URL first, then `Cell.link` + Ctrl+K).

## Summary by CodeRabbit

- **New Features**
  - Spreadsheet cells now detect multiple URLs, email addresses, multiline content, and formula results.
  - Links are visibly styled with precise hover and click targeting.
  - Hover cards list available links with options to open or copy them.
  - Read-only viewers can open links with a plain click; Ctrl/Cmd+click remains available elsewhere.
  - Added a shortcut reference for opening links.

- **Documentation**
  - Added design documentation and implementation lessons for spreadsheet link handling.

- **Tests**
  - Expanded coverage for detection, safety, layout, clipping, hit testing, and hover behavior.


hackerwins:
@coderabbitai retry this.

hackerwins:
@claude loop.

PR #1084, opened 2026-09-19
Fix stale find matches before replacing

### LINKED ISSUE #1083: Docs: Find & Replace replaces the wrong text after the document is edited
### What happened:

In Docs, if the document content is edited while the Find & Replace dialog remains open, Replace may modify unrelated text instead of the current search target.

For example, starting with: `cat dog`
I searched for dog, set the replacement text to fox, and then inserted big  at the beginning of the document before clicking Replace.

The expected result was: `big cat fox`
However, the actual result was: `big fox dog`

The searched text, dog, remains unchanged, while the unrelated word cat is replaced instead.

Even if the document is edited after performing a search, Find & Replace should replace the matching text in the current document state.

```
Before replacement: big cat dog
Expected result:    big cat fox
Actual result:      big fox dog
```

1. Open a Docs document and enter: `cat dog`
2. Open Find & Replace.
3. Enter dog in Find and fox in Replace with. Do not click Replace yet.
4. While keeping the Find & Replace dialog open, insert big  at the beginning of the document.
    The document should now contain: `big cat dog`
5. Without re-entering the search term or closing the Find & Replace dialog, click Replace.
6. Observe that the document becomes: `big fox dog`
     instead of: `big cat fox`

I reproduced this behavior locally on the main branch at commit `ef5aa3e96` using Google Chrome.

Based on code investigation, the issue appears to be related to `FindReplaceState` retaining the character positions calculated when the search was performed. If the document content changes afterward, replacement still appears to use those stale positions.

Related files:
- `packages/docs/src/view/find-replace.ts`
- `packages/frontend/src/app/docs/docs-find-bar.tsx`

In automated tests connecting the actual Find & Replace component with the document model, the same issue was reproduced for both Replace and Replace All.

A control test where the document was not edited after searching behaved correctly. I also searched existing issues and pull requests but did not find one describing the same behavior.

The browser reproduction described above was manually verified for Replace. The Replace All case was confirmed through automated testing.

### Environment:

- Operating system: macOS
- Browser and version: Google Chrome 153.0.8010.53
- Environment: Local development server (`http://localhost:5173`)
- Commit: `ef5aa3e96`

yexnsxo:
## Summary

- Refresh Find & Replace matches against the current document before `Replace` and `Replace All`.
- Add regression tests for document edits that shift previously stored search offsets.

## Why

`FindReplaceState` stores the offsets of search matches when a search is run.
If the document is edited while the Find & Replace panel remains open, those
stored offsets can become stale.

As a result, `Replace` or `Replace All` could modify text at the old offsets
instead of the text that currently matches the query.

This change refreshes the matches immediately before replacing so the
replacement uses offsets from the current document.

## Linked Issues

Fixes #1083

## Author checklist

## Verification

forced the whole suite.

Skip reason (if applicable):

- `verify-self` was run locally, but `docs:check` failed on a timing-sensitive
  test in `large-paste-indicator.test.ts`. The same test file passed 10/10
  when rerun independently.
- The Find & Replace regression tests passed 14/14.
- Manually verified the reported reproduction:
  `cat dog` → search for `dog` → insert `big ` → Replace with `fox`
  → `big cat fox`.
- `verify-integration` was not run locally because this change only touches the
  Docs Find & Replace state and its unit tests.

## Risk Assessment

- User-facing risk: Low. The change only refreshes existing search results before replacement.
- Data/security risk: None identified.
- Rollback plan: Revert this commit to restore the previous replacement behavior.

## Notes for Reviewers

  - No visual UI changes.
  - After the document is edited, existing search highlights and the match counter
    can still remain visually stale until another search or replace action refreshes
    them. Keeping these synchronized immediately could be handled separately.
- Follow-up work (if any):
  - Consider synchronizing search highlights and the match counter immediately when
    the document changes while Find & Replace is open.
  - A dedicated regression test could also be added for the case where the previously
    matched text is removed before Replace. The current refresh safely makes this a
    no-op when no matches remain.

## Summary by CodeRabbit

- **Bug Fixes**
  - Find & Replace now refreshes matches against the current document before replacing text.
  - Replace continues targeting the selected occurrence when edits shift surrounding text or add matches.
  - Replace All correctly handles shifted, newly added, and removed matches after document edits.
  - Replacements no longer apply to stale offsets or unintended text.


hackerwins:
@claude loop

hackerwins:
@claude review

hackerwins (dismissed):
Thanks for your contribution.

hackerwins:
Thanks for the fix — the diagnosis in #1083 is right and the re-search is the
right shape. I've pushed a follow-up commit (64a2b6a) to this branch addressing
the review panel, so we can land it together.

### Addressed

**The blocking finding.** Re-searching removes the stale-offset write, but
`search()` re-binds the active match by *ordinal* index, and that index is only
stable under the edit `search()` was written for — a replace, which shrinks the
match set from a known position. Under an arbitrary intervening edit it isn't:
in `dog cat dog` with the **second** `dog` active, typing `dog ` at the head
makes index 1 name the occurrence that was just inserted, so Replace rewrites
text you never had highlighted. The failure moved rather than closed.

A private `refresh()` now re-runs the search and re-anchors the active match by
position — the nearest match in the same block, ties going to the later one (an
insertion at the active match's own start offset pushes it forward), with
`search()`'s clamped ordinal left standing only when that block holds no match
at all. `should keep the active match when one is inserted before it` pins it,
and fails on `afd0700a`.

**The find bar's stale guards.** `handleReplace` / `handleReplaceAll` re-checked
`activeIndex` and `matches.length` against whatever the last search left behind,
so they could only ever refuse a replace the refreshed state would allow — and
when they fired they also skipped `syncHighlights()`, leaving the stale counter
on screen to justify the refusal. Both are gone; `readOnly` stays, being the one
condition the component alone knows.

**The test gaps.** The `replaceAll` case now carries two matches plus one the
search never saw, so it exercises the reverse sweep instead of duplicating the
`replaceActive` case; the newly reachable zero-matches-after-edit early return
is covered; and both refresh cases now assert `matches` / `activeIndex`, not
just the final text.

### Not changed, with reasons

- **"the new `search()` can throw uncaught out of the click handler"** — it
  can't. `Doc.searchText` wraps `new RegExp(...)` in its own `try`/`catch` and
  returns `[]` on a bad pattern (`packages/docs/src/model/document.ts:906-912`),
  which is exactly what the find bar's `catch` was setting by hand.
- **"Replace All now rewrites matches the user was never shown"** — that is what
  Replace All means; its scope is the document, not the last render. Writing a
  subset and leaving occurrences behind is the more surprising outcome. The real
  complaint underneath is the stale counter, which is the relocated finding
  below.
- **"the extra regex scan amplifies the ReDoS surface"** — real surface,
  pre-existing, but one extra scan per *click* against a pattern already
  re-scanned on every keystroke isn't meaningful amplification. A pathological
  pattern hangs the tab at search time, long before the Replace button. Bounding
  it belongs with `searchText`.

### Known limitation (your follow-up note, and the panel's relocated finding)

Highlights, the match counter and next/previous still read whatever the last
search left behind — only the two write paths refresh. Closing that means
re-running the search on every document change, and there's no single hook for
it: remote edits reach `store.onRemoteChange`, which `docs-view.tsx` already
owns, and local typing reaches none of it. Worth its own issue rather than
growing this one.

yexnsxo:
Thanks for the follow-up commit! I’ll take a closer look at the re-anchoring logic as well!!😃

